PR #2, opened 2020-03-26
update atari.py

Mehooz:
update atari.py, setup.py
reformat some other files in Pycharm
add some frameworks for mujoco, but haven't completed


Mehooz:
fix a bug in setup.py

PR #52, opened 2020-05-21
ensure dqn reproducibility

magicly:
if dqn in infer model, eps should be set to 0, and no need to random, to ensure reproducibility


Trinkle23897:
Thanks!
By the way, according to my previous experience, if there is no randomness in the decision-making process, DQN may fail in some cases and its performance will drop down comparing with the "little-randomness" version.
The most important thing is to ensure the distribution of trajectories in test mode is the same as which in the eval mode.

PR #61, opened 2020-05-29
Minor refactor for Batch class.

### LINKED ISSUE #60: Refactoring of Batch class
    + [ ] exception-raising bug
    + [ ] RL algorithm bug
    + [x] new feature request
- [x] I have mentioned version numbers, operating system and environment, where applicable.

___________

I just reviewed your implementation of [Batch](https://github.com/thu-ml/tianshou/blob/master/tianshou/data/batch.py) class. It is nice, but you are always accessing the value of the stored objects using key instead of iterators. I think it would be both clearer and more efficient to rely on iterators everywhere, since it would no longer access the hashing table every time you want to get an item.

If you want, I can do it by myself and open a PR.

duburcqa:
Related to issue https://github.com/thu-ml/tianshou/issues/60

### DIFF
--- a/tianshou/data/batch.py
+++ b/tianshou/data/batch.py
@@ -4,7 +4,7 @@ import numpy as np
 from typing import Any, List, Union, Iterator, Optional
 
 
-class Batch(object):
+class Batch:
     """Tianshou provides :class:`~tianshou.data.Batch` as the internal data
     structure to pass any kind of data to other methods, for example, a
     collector gives a :class:`~tianshou.data.Batch` to policy for learning.
@@ -80,9 +80,9 @@ class Batch(object):
                     ])
             elif isinstance(v, dict):
                 self._meta[k] = list(v.keys())
-                for k_ in v.keys():
+                for k_, v_ in v.items():
                     k__ = '_' + k + '@' + k_
-                    self.__dict__[k__] = v[k_]
+                    self.__dict__[k__] = v_
             else:
                 self.__dict__[k] = kwargs[k]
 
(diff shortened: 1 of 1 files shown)


tianshou/data/batch.py
```diff
@@ -128,12 +128,13 @@ def __repr__(self) -> str:
...
         """Return self.keys()."""
-        return sorted([
-            i for i in self.__dict__ if i[0] != '_'] + list(self._meta))
+        return list(self._meta.keys()) + \
+            [k for k in self.__dict__.keys() if i[0] != '_']
 
     def values(self) -> List[Any]:
         """Return self.values()."""
-        return [self[k] for k in self.keys()]
+        return list(self._meta.values()) +
```
Trinkle23897: I think the values in self._meta should not be returned. Because the data in self._meta is the inner name of batch storage, just some strings.

tianshou/data/batch.py
```diff
@@ -128,12 +128,13 @@ def __repr__(self) -> str:
 
     def keys(self) -> List[str]:
         """Return self.keys()."""
-        return sorted([
```
duburcqa, line 131: IMHO, sorting the keys was not a good idea if it was not strictly necessary.

Trinkle23897:
It seems to fail to pass the code-style check. Could you please follow the rules in https://tianshou.readthedocs.io/en/latest/contributing.html?

tianshou/data/batch.py
duburcqa: Yes, I fixed that.

duburcqa:
Yes of course :)

PR #63, opened 2020-05-29
Robust conversion from/to numpy/pytorch

### LINKED ISSUE #62: Fully support from/to numpy/pytorch for Batch
The current implementation of PPO and other policy algorithm do not support action dict because of [this line](https://github.com/thu-ml/tianshou/blob/be9ce44290826962889787558cb4984384a30fac/tianshou/policy/modelfree/ppo.py#L132).

It could be solved by adding to new method to `Batch` class to convert back the relevant fields to torch.Tensor.

I'm opening a PR to fix that.

duburcqa:
Related to issue https://github.com/thu-ml/tianshou/issues/62

### DIFF
--- a/test/base/test_batch.py
+++ b/test/base/test_batch.py
@@ -1,4 +1,5 @@
 import pytest
+import torch
 import numpy as np
 
 from tianshou.data import Batch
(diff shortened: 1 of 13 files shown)


duburcqa:
I still need to replace `torch.tensor` by `to_torch` everywhere it makes sense.

PR #74, opened 2020-06-04
Enable getattr for SubprocVecEnv.

duburcqa:
    + [ ] exception-raising fix
    + [x] new feature

So far it is not possible to access the attributes of the environments of a SubprocVecEnv. This would be useful for debug to get the value of some internal buffers, but also to simply get the action and observation spaces for the environments.

### DIFF
--- a/test/base/test_env.py
+++ b/test/base/test_env.py
@@ -12,7 +12,7 @@ else:  # pytest
 def test_vecenv(size=10, num=8, sleep=0.001):
     verbose = __name__ == '__main__'
     env_fns = [
-        lambda: MyTestEnv(size=i, sleep=sleep)
+        lambda i=i: MyTestEnv(size=i, sleep=sleep)
         for i in range(size, size + num)
     ]
     venv = [
--- a/test/base/test_env.py
+++ b/test/base/test_env.py
@@ -1,5 +1,6 @@
 import time
 import numpy as np
+from gym.spaces.discrete import Discrete
 from tianshou.env import VectorEnv, SubprocVectorEnv, RayVectorEnv
 
 if __name__ == '__main__':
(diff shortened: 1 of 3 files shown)


Trinkle23897:
What about other vecenv? (Dummy)

duburcqa:
It is less elegant, but simple list comprehension would do the trick without having to modify the current implementation:
```
envs_a = [env.a for env in vec_envs.envs]
```
Perhaps similar `getattr` methods should be added to `VectorEnv` for consistency of the API. What do you think ?

Trinkle23897:
I agree with you!

duburcqa:
Done ! Note that for SubprocVectorEnv, only serializable attribute can be retrieved from the environements. This is by design of multiprocessing. In practice it should not be a big deal (now that Batch are serializable).

duburcqa:
> And test...?

Tests are coming x)

duburcqa:
It should be ok now :)

test/base/test_env.py
```diff
@@ -11,14 +11,8 @@
...
-        lambda: MyTestEnv(size=size, sleep=sleep),
-        lambda: MyTestEnv(size=size + 1, sleep=sleep),
-        lambda: MyTestEnv(size=size + 2, sleep=sleep),
-        lambda: MyTestEnv(size=size + 3, sleep=sleep),
-        lambda: MyTestEnv(size=size + 4, sleep=sleep),
-        lambda: MyTestEnv(size=size + 5, sleep=sleep),
-        lambda: MyTestEnv(size=size + 6, sleep=sleep),
-        lambda: MyTestEnv(size=size + 7, sleep=sleep),
+        lambda: MyTestEnv(size=i, sleep=sleep)
+        for i in range(size, size + num)
```
Trinkle23897, line 16: No, you can have a test and will find out that all env's size are identical.

Trinkle23897:
```python
In [17]: env_fns = [lambda: MyTestEnv(size=i, sleep=sleep) 
    ...:         for i in range(size, size + num)]                                                                                                                                                                  

In [18]: env_fn = [i() for i in env_fns]                                                                                                                                                                            

In [19]: print(env_fn[0].size, env_fn[1].size)                                                                                                                                                                      
17 17
```

duburcqa: Yeh I found the issue. It was coming from the lambdas using current `i` value instead of the one existing at definition. Sorry about that.

PR #79, opened 2020-06-16
Fixed TSAIL Link in README.md

huangyz0918:


Trinkle23897:
Nope, Institute for Artificial Intelligence is not equal to TSAIL.

PR #87, opened 2020-06-23
Batch refactoring

duburcqa:
- Get rid of metadata property `_meta`, which simplifies and clarifies the implementation
- Enable to initialize a `Batch` instance using a `dict` instead of individual keys only, simular to Python `dict` itself or `gym.spaces.Dict`
- Add full support of 0-dim `np.array`
- Do not sort `Batch` keys anymore for efficiency. Add `items` method. Now it mimics completely the behavior of native `dict`
- Rename `cat` in `cat_` and add `cat` method for consistency with in-place / out-of-place torch API
- Add `stack` method, which is equivalent to torch/numpy `stack` method (It adds an extra dimension)
- Preserve original data type when stacking multiple `Batch`/`dict`. 0-dim `np.array` must be used to get a 1-dim `np.array` after Batch stacking.

### DIFF
--- a/tianshou/data/batch.py
+++ b/tianshou/data/batch.py
@@ -1,4 +1,5 @@
 import torch
+import copy
 import pprint
 import warnings
 import numpy as np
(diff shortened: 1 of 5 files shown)


duburcqa:
@Trinkle23897 Ready to be merged if you are ok with the changes

tianshou/data/batch.py
```diff
@@ -220,35 +224,55 @@ def to_torch(self,
     def append(self, batch: 'Batch') -> None:
         warnings.warn('Method append will be removed soon, please use '
                       ':meth:`~tianshou.data.Batch.cat`')
```
Trinkle23897, line 226: here, change cat to cat_

PR #93, opened 2020-06-24
Buffer refactoring to support batch over batch reliably

### LINKED ISSUE #90: Batch to Numpy compatibility

______

Input:
```
```
Output:
```
Batch(
    a: array([2., 3.]),
    b: 5.5,
)
```

duburcqa:
- Add axis option to Batch `stack` method
- Add support of item assignment for Batch (partial item assignment also supported)
- Generalized slicing using any type that is usually accepted by `np.ndarray` for both `Batch` and `Buffer` classes
- Improve robustness of Batch implement by avoiding using direct assignment
- Remove metadata attribute of Buffer class and inherit directly from Batch. It greatly simplify and clarify the implementation.

This PR incidently fixes the support of deeply nested Batch over Batch for Buffers.

### DIFF
--- a/test/base/test_batch.py
+++ b/test/base/test_batch.py
@@ -65,10 +65,15 @@ def test_batch():
     assert batch2_sum.a.b == (batch2.a.b + 1.0) * 2
     assert batch2_sum.a.c == (batch2.a.c + 1.0) * 2
     assert batch2_sum.a.d.e == (batch2.a.d.e + 1.0) * 2
-    batch2.a.d[0] = {'e': 4.0}
-    assert batch2.a.d.e[0] == 4.0
-    batch2.a.d[0] = Batch(e=5.0)
-    assert batch2.a.d.e[0] == 5.0
+    batch3 = Batch(a={
+        'c': np.zeros(1),
+        'd': Batch(e=np.array([0.0]), f=np.array([3.0]))})
+    batch3.a.d[0] = {'e': 4.0}
+    assert batch3.a.d.e[0] == 4.0
+    batch3.a.d[0] = Batch(f=5.0)
+    assert batch3.a.d.f[0] == 5.0
+    with pytest.raises(ValueError):
+        batch3.a.d[0] = Batch(f=5.0, g=0.0)
 
 
 def test_batch_over_batch():
--- a/test/base/test_batch.py
+++ b/test/base/test_batch.py
@@ -65,6 +65,10 @@ def test_batch():
     assert batch2_sum.a.b == (batch2.a.b + 1.0) * 2
     assert batch2_sum.a.c == (batch2.a.c + 1.0) * 2
     assert batch2_sum.a.d.e == (batch2.a.d.e + 1.0) * 2
+    batch2.a.d[0] = {'e': 4.0}
+    assert batch2.a.d.e[0] == 4.0
+    batch2.a.d[0] = Batch(e=5.0)
+    assert batch2.a.d.e[0] == 5.0
 
 
 def test_batch_over_batch():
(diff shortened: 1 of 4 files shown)


Trinkle23897:
Do you think the `_meta` should be removed as Batch?

duburcqa:
Since you are asking, yes I think so. It makes everything pretty difficult to understand and maintain without adding any feature at the end. However, I'm afraid I don't really understand how this mechanism is working right now, so I'm not able to refactor this submodule by myself. It would be awesome to take advantage of the newly added flexibility of Batch to simplify the implement of this module.

Trinkle23897:
It is almost the same as the previous `Batch` implementation. The difference includes automatically create the np.array with given data. Nothing special.

duburcqa:
Ok I'll have a look at it.

duburcqa:
I think I should be able to handle it finally. I let you know.

duburcqa:
@Trinkle23897 Ready to be merged ! I must admit I'm pretty happy with the result :)

Trinkle23897:
The failure of unit tests is because the env info may not be consistent throughout the whole trajectory, e.g., 
```
step 0: info = {}
step 1: info = {}
...
step T - 1: info = {'TimeLimit.truncated': True}
```

```python3
In [6]: envname = 'Pendulum-v0'

In [7]: e = gym.make(envname) 
   ...: e.reset() 
   ...: cnt = 0 
   ...: last = {} 
   ...: while True: 
   ...:     _, _, d, info = e.step(e.action_space.sample()) 
   ...:     if info != last: 
   ...:         print(cnt, last, info) 
   ...:     cnt += 1 
   ...:     last = info 
   ...:     if d: 
   ...:         break 
   ...:
199 {} {'TimeLimit.truncated': True}
```

Some other gym environments have the same behavior, including `FetchReach-v1`. I'm not sure whether gym has documented the rules in `info`, so I write those tests to simulate these behaviors.

duburcqa:
I see. Supporting this behavior is not straightforward. Are we supposed to add None everywhere but in only index for which this key is actually available?  Conversely, if some key are missing in the new item, we could try to set the missing ones to none or nan, depending on the type.

duburcqa:
So here is the new behaviour I implemented:

```
batch = Batch(c=[Batch(d=1.0), Batch(d=9.0)])                                                                                                                                                       
batch[0] = Batch(a=2, b=3)                                                                                                                                                                          
>>> Batch(
    c: Batch(
           d: array([nan,  9.]),
       ),
    b: [3, None],
    a: [2, None],
)
```

Trinkle23897:
One possible solution is to only broadcast the reserved 7 keys.
Another solution is that this new feature can be only appeared in Buffer instead of Batch.

duburcqa:
It should not be an issue, since only item assignment is broadcasted, not direct assignement, as in your example.

Trinkle23897:
> It should not be an issue, since only item assignment is broadcasted, not direct assignment, as in your example.

So can I interpret this to mean that only the Batch assignment will trigger this broadcast operation?

duburcqa:
Doing `batch.obs[indices] = ...` (item assignment) triggers this broadcasting, while `batch.obs = ...` (direct assignment) does not. It is important, since item assignment must not break item access, which means doing afterward `batch.obs[indices]`must be valid under any condition, which is not the case if you do not broadcast, while direct assignment has not such guarantee.

duburcqa:
> Another solution is that this new feature can be only appeared in Buffer instead of Batch.

Finally I will do this, since it is impossible to get a reliable behaviour using broadcasting (lower level nested Batch instances have no idea of the size of the top Batch, making it impossible to perform broadcast), and it was pretty hard to maintain.

duburcqa:
@Trinkle23897 PR ready to be merged ! All tests passing successfully finally !

Trinkle23897:
By the way, I found that in #92, the overhead of `Batch.__getitem__` grows 20% because of calling `_valid_bounds`. Do you have any idea?

duburcqa:
YEs, I'm on it !

duburcqa:
@Trinkle23897 Can you check this branch ? https://github.com/thu-ml/tianshou/pull/95

PR #95, opened 2020-06-25
Do not check bounds since it is always valid when everything is fine.

### LINKED ISSUE #90: Batch to Numpy compatibility

### LINKED ISSUE #96: Batch & Buffer profiling
Related to #95

duburcqa:
Avoid the overhead of checking bounds for subfields since it should be fine useless the user is messing up terribly somewhere else.

### DIFF
--- a/tianshou/data/batch.py
+++ b/tianshou/data/batch.py
@@ -171,18 +171,12 @@ class Batch:
                     b[k] = Batch()
                 elif hasattr(v, '__len__') and (not isinstance(
                         v, (np.ndarray, torch.Tensor)) or v.ndim > 0):
-                    if _valid_bounds(len(v), index):
-                        if isinstance(index, (int, np.integer)) or \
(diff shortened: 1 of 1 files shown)


Trinkle23897:
Simple solution, I like it!

Trinkle23897:
Hmm... I didn't figure out what's the bottleneck. Still 8 mins per unittest.
The seed is fixed so that the training process would be exactly the same as the previous version.

duburcqa:
Do you have any idea how we could profile the code ? Perhaps Batch is supporting too many types, requiring many `isinstance` check for each of them. 8 minutes on your computer or github ? Because sometimes github slow down for some reason.

Trinkle23897:
On GitHub. But in my computer, it is almost the same as previous. I'll open an issue about profiling Batch.

PR #97, opened 2020-06-25
Use lower-level API to reduce overhead.

### LINKED ISSUE #96: Batch & Buffer profiling

duburcqa:
Related to https://github.com/thu-ml/tianshou/issues/96

### DIFF
--- a/test/base/test_batch.py
+++ b/test/base/test_batch.py
@@ -72,7 +72,7 @@ def test_batch():
     assert batch3.a.d.e[0] == 4.0
     batch3.a.d[0] = Batch(f=5.0)
     assert batch3.a.d.f[0] == 5.0
-    with pytest.raises(ValueError):
+    with pytest.raises(KeyError):
         batch3.a.d[0] = Batch(f=5.0, g=0.0)
 
 
--- a/test/base/test_batch.py
+++ b/test/base/test_batch.py
@@ -39,12 +39,17 @@ def test_batch():
         'c': np.zeros(1),
         'd': Batch(e=np.array(3.0))}])
     assert len(batch2) == 1
+    assert Batch().size == 0
+    assert batch2.size == 1
     with pytest.raises(IndexError):
         batch2[-2]
     with pytest.raises(IndexError):
         batch2[1]
+    assert batch2[0].size == 1
     with pytest.raises(TypeError):
         batch2[0][0]
+    with pytest.raises(TypeError):
+        len(batch2[0])
     assert isinstance(batch2[0].a.c, np.ndarray)
     assert isinstance(batch2[0].a.b, np.float64)
     assert isinstance(batch2[0].a.d.e, np.float64)
--- a/test/base/test_batch.py
+++ b/test/base/test_batch.py
@@ -117,10 +117,15 @@ def test_batch_cat_and_stack():
     b12_stack = Batch.stack((b1, b2))
     assert isinstance(b12_stack.a.d.e, np.ndarray)
     assert b12_stack.a.d.e.ndim == 2
-    b3 = Batch(a=np.zeros((3, 4)))
-    b4 = Batch(a=np.ones((3, 4)))
+    b3 = Batch(a=np.zeros((3, 4)),
+               b=torch.ones((2, 5)),
+               c=Batch(d=[[1], [2]]))
+    b4 = Batch(a=np.ones((3, 4)),
+               b=torch.ones((2, 5)),
+               c=Batch(d=[[0], [3]]))
     b34_stack = Batch.stack((b3, b4), axis=1)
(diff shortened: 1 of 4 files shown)


duburcqa:
@Trinkle23897 Ready to be merged ! Welcome performance comeback :smiley:

Trinkle23897:
cool!

Trinkle23897:
I'll merge one hour later (if you want to add more).

duburcqa:
It should be OK now. I just noticed that some unit test were missing and added them. Feel free to merge when you want :+1:

Trinkle23897:
A bug:
```python
In [1]: from tianshou.data import ReplayBuffer, Batch                                                                                                                                                               

In [2]: b = ReplayBuffer(size=10)                                                                                                                                                                                   

In [3]: b.add(1,1,1,1,1, {'a':3, 'b': {'c': 5}})                                                                                                                                                                    

In [4]: b                                                                                                                                                                                                           
Out[4]: 
ReplayBuffer(
    obs: array([1, 0, 0, 0, 0, 0, 0, 0, 0, 0]),
    act: array([1, 0, 0, 0, 0, 0, 0, 0, 0, 0]),
    rew: array([1, 0, 0, 0, 0, 0, 0, 0, 0, 0]),
    done: array([1, 0, 0, 0, 0, 0, 0, 0, 0, 0]),
    obs_next: array([1, 0, 0, 0, 0, 0, 0, 0, 0, 0]),
    info: Batch(
              a: array([3, 3, 3, 3, 3, 3, 3, 3, 3, 3]),
              b: Batch(
                     c: array([5, 5, 5, 5, 5, 5, 5, 5, 5, 5]),
                 ),
          ),
    policy: Batch(),
)
```
The `b.info.a` and `b.info.b.c` should be zero-initialized.

duburcqa:
Why ? There is no unit test checking it and you are not supposed to access those data since you are ahead of the buffer index. I can easily fix it but for me it is not a bug if it is not part of the specifications (and hence checked by unit tests).

Trinkle23897:
Yes, it's my fault not to add this script to unittest. 
Btw,
```python
In [6]: b[1]                                                                                                                                                                                                        
Out[6]: 
Batch(
    obs: 0,
    act: 0,
    rew: 0,
    done: 0,
    obs_next: 0,
    info: Batch(
              a: 3,
              b: Batch(
                     c: 5,
                 ),
          ),
    policy: Batch(),
)

In [7]: b[1].info.b.c                                                                                                                                                                                               
Out[7]: 5
```
I just want to keep all the syntax to be consistent: the obs is zero-init, so that the info batch should also be zero-init.

You can image a scenario: a user add info0, info2, info4, info6. At some time he wants to get info1, which should be `{}`. However, in this implementation, it will return info0.

duburcqa:
Yes I understand your point, yet I'm not sure if zeroed yes better than None/nan (as I'm doing for partial item assignment of Batch). For now I did as you asked and added unit tests.

Trinkle23897:
Hmm, I think None is better than zero.
But if we save `None` instead of zero, the numpy array will fall back to np.objects, which will possibly cause some performance issue (e.g. the alignment and indexing)...? I'm not sure.

duburcqa:
> Hmm, I think None is better than zero.
But if we save None instead of zero, the numpy array will fall back to np.objects, which will possibly cause some performance issue (e.g. the alignment and indexing)...? I'm not sure.

Yes, that's why I'm using nan for `np.array` to avoid such issue.

duburcqa:
So, I'm on it. I'll let you know when it is ready.

duburcqa:
Done ! Here is the new behavior. Unit test included. (Note that only '0' is available for integer arrays since there is no such thing as nan/None for them).

```                                                                       
In [1]:    b = ReplayBuffer(size=10)                                                                                                                                                                       
In [2]:    b.add(1, 1, 1, 'str', 1, {'a': 3, 'b': {'c': 5.0}})                                                                                                                                             

In [2]:   b  
ReplayBuffer(
    obs: array([1, 0, 0, 0, 0, 0, 0, 0, 0, 0]),
    act: array([1, 0, 0, 0, 0, 0, 0, 0, 0, 0]),
    rew: array([1, 0, 0, 0, 0, 0, 0, 0, 0, 0]),
    done: array(['str', None, None, None, None, None, None, None, None, None],
                dtype=object),
    obs_next: array([1, 0, 0, 0, 0, 0, 0, 0, 0, 0]),
    info: Batch(
              a: array([3, 0, 0, 0, 0, 0, 0, 0, 0, 0]),
              b: Batch(
                     c: array([ 5., nan, nan, nan, nan, nan, nan, nan, nan, nan]),
                 ),
          ),
    policy: Batch(),
)
```

Trinkle23897:
And one more thing: it would be of great help if you add the new feature of `Batch` and `Buffer` into the documentation accordingly!

Issue #19, opened 2020-04-04
RNN support


miriaford:
- [ ] I have marked all applicable categories:
    + [ ] exception-raising bug
    + [ ] RL algorithm bug
    + [ ] documentation request (i.e. "X is missing from the documentation.")
    + [x] new feature request
- [x] I have visited the [source website], and in particular read the [known issues]
- [x] I have searched through the [issue tracker] for duplicates
- [ ] I have mentioned version numbers, operating system and environment, where applicable:
  ```python
  import tianshou, sys
  print(tianshou.__version__, sys.version, sys.platform)
  ```

  [source website]: https://github.com/thu-ml/tianshou/
  [known issues]: https://github.com/thu-ml/tianshou/#faq-and-known-issues
  [issue tracker]: https://github.com/thu-ml/tianshou/issues?q=

I see on README that RNN support is on your TODO list. However, the module API seems to support RNN ( `forward(obs, state)` method). Could you please provide some examples on how to train RNN policy? Thanks!

Trinkle23897:
Oh yes, in fact it has already support the rnn policy but I will write a small demonstration script after finishing the documentation. Thanks for pushing me.

miriaford:
Thanks! Do you mind quickly posting a snippet here and explain how to use the RNN policy? I need to run something ASAP, thanks a lot!

fengredrum:
Actually, my experiments show that a RNN policy is nearly always inferior to a MLP policy in robotic control domains. 
Besides, RNNs suffer from slow training speed :)

miriaford:
It varies from case to case, but I agree that RNN might not be the best option. I do need it as a baseline though, so will still appreciate it if someone can explain how to run RNN policy with Tianshou. Thanks!

Trinkle23897:
@miriaford I upload a script in [test/discrete/test_drqn.py](https://github.com/thu-ml/tianshou/blob/master/test/discrete/test_drqn.py).
What I have changed in the codebase:
1. modify the replay buffer to support frame_stack sampling;
2. fix collector for clearing a state when the corresponding episode is finished;

How to use:
1. add an argument "stack_num" in training collector's replay buffer [code snippet](https://github.com/thu-ml/tianshou/blob/master/tianshou/data/buffer.py#L97);
2. change your actor-network to recurrent style [code snippet](https://github.com/thu-ml/tianshou/blob/master/test/discrete/net.py#L57);
[...]

PR #100, opened 2020-06-26
Enable partial stacking at Batch level

### LINKED ISSUE #90: Batch to Numpy compatibility

### LINKED ISSUE #96: Batch & Buffer profiling

duburcqa:
Allow stacking of inconsistent batches:
```
In [1]: from tianshou.data import Batch, to_numpy ; import numpy as np ; import torch                                                                                                                       
In [2]: Batch(np.array([{'is_success': True, 'reward': {'done': 0.0001, 'direction': -0.11480981}}, 
   ...:                 {'is_success': False, 'reward': {'done': 0.0001}}], dtype=object))                                                                                                                  
Out[2]: 
Batch(
    is_success: array([ True, False]),
    reward: Batch(
                done: array([0.0001, 0.0001]),
                direction: array([-0.11480981,  0.        ]),
            ),
)
```

Before that, the result was unexpected (no failure and different result depending of Batch order).

As you may notice, I also replaced 'nan' in float array by 0.0, since at the end it is usually meaningful and convenient.

It comes with a slight performance improvement: https://github.com/thu-ml/tianshou/issues/96

output_V2

### DIFF
--- a/test/base/test_batch.py
+++ b/test/base/test_batch.py
@@ -126,6 +126,13 @@ def test_batch_cat_and_stack():
     b34_stack = Batch.stack((b3, b4), axis=1)
     assert np.all(b34_stack.a == np.stack((b3.a, b4.a), axis=1))
     assert np.all(b34_stack.c.d == list(map(list, zip(b3.c.d, b4.c.d))))
+    b5_dict = np.array([{'a': False, 'b': {'c': 2.0, 'd': 1.0}},
+                        {'a': True, 'b': {'c': 3.0}}])
+    b5 = Batch(b5_dict)
+    assert b5.a[0] == np.array(False) and b5.a[1] == np.array(True)
+    assert np.all(b5.b.c == np.stack([e['b']['c'] for e in b5_dict], axis=0))
+    assert b5.b.d[0] == b5_dict[0]['b']['d']
+    assert b5.b.d[1] == 0.0
 
 
 def test_batch_over_batch_to_torch():
--- a/test/base/test_buffer.py
+++ b/test/base/test_buffer.py
@@ -36,7 +36,7 @@ def test_replaybuffer(size=10, bufsize=20):
     assert b.info.a[0] == 3 and b.info.a.dtype == np.integer
     assert np.all(b.info.a[1:] == 0)
     assert b.info.b.c[0] == 5.0 and b.info.b.c.dtype == np.inexact
-    assert np.all(np.isnan(b.info.b.c[1:]))
+    assert np.all(b.info.b.c[1:] == 0.0)
 
 
 def test_ignore_obs_next(size=10):
--- a/tianshou/data/batch.py
+++ b/tianshou/data/batch.py
@@ -171,11 +171,15 @@ class Batch:
                 f"Index {index} out of bounds for Batch of len {len(self)}.")
         else:
             b = Batch()
+            is_index_scalar = isinstance(index, (int, np.integer)) or \
+                (isinstance(index, np.ndarray) and index.ndim == 0)
             for k, v in self.items():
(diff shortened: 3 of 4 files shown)


duburcqa:
@Trinkle23897 Just for you to know, I'm an engineer and for the first time, and thanks to this PR (and all the previous ones) a real case application is finally running without raising an exception :laughing: Perhaps it is time to release 0.2.4 :stuck_out_tongue:

Trinkle23897:
Sure and much appreciate~
How about the documentation update?

duburcqa:
Yes I'll do it now. Could you please tell me what/where to update them and what you have in mind exactly ?

Trinkle23897:
https://github.com/thu-ml/tianshou/blob/master/CONTRIBUTING.md

duburcqa:
Yes I know this, but you mean only updating this page https://tianshou.readthedocs.io/en/latest/api/tianshou.data.html ?

Trinkle23897:
Exactly. I mean that you can add the new feature in https://tianshou.readthedocs.io/en/latest/api/tianshou.data.html#tianshou.data.Batch
```
Batch has other methods, including __getitem__(), __len__(), append(), and split():
```

also with buffer.

Trinkle23897:
You can make the documentation show `__xxx__` by modifying this line:
https://github.com/thu-ml/tianshou/blob/70aa7bf93ebe25d61e4566f3880b07a40eb702aa/docs/conf.py#L61
For example, add `__idiv__` into 'special-members'.
And the readthedocs page will be updated only if the master version has been updated. So it is better to preview the documentation in your own local computer.

duburcqa:
Added missing documentation. It turns out that I didn't modified the behaviour of Buffer by any means, it was only internal modifications, so I have not updated it.

Trinkle23897:
Could it support slicing method such as `batch[:, i]`?
Because I find that the advance RNN hidden state (#19) needs this and otherwise I should recursively get the value.

duburcqa:
In practice it shouldn't be very hard, but conceptually Batch is missing the notion of multidimensional samples, for instance only `len` is available, and it corresponds to the dimension to the first axis of the stored data. The first step would be to add a `shape` method, similar to numpy. After that It shouldn't be much trouble. Maybe you could open an issue.

The issue is that `list` cannot be supported since computing the shape of a `list` is very expensive and grows exponentially with depth.

duburcqa:
Adding this property to `Batch` class would do the trick (note that `isinstance` `if ... else` should be used to early return `[]` in case of native Python scalar) :

```
    @property
    def shape(self) -> List[int]:
        """Return self.size."""
        if len(self.__dict__.keys()) == 0:
            return []
        else:
            data_shape = []
            for v in self.__dict__.values():
                try:
                    data_shape.append(v.shape)
                except AttributeError:
                    raise TypeError("No support for 'shape' method with "\
                                    f"type {type(v)} in class Batch.")
            return min(*data_shape) if len(data_shape) > 1 else data_shape[0]
```

Then, `_valid_bounds(len(self), index)` must be replaced by `np.all(map(_valid_bounds, zip(self.shape, index))`, ONLY in the case where index is a actually a list of list of indices. No need to change anything else, since the rest of the method is already compatible with multidimensional indexing (expect for list, but list is not supported by `shape`, so an exception will be raised before reaching this point).

NB: You can ping me on the issue/PR if you want a review.

duburcqa:
By the way, it may be good to overload direct assignment to make sure that every list are converted to numpy. Indeed, list are a recurrent issue for slicing, stacking and concatenation, and it introduces many `isinstance` `if ... else ...` for no real advantage.

Issue #105, opened 2020-06-29
Refactoring of Buffer & Collector


Trinkle23897:
In the current implementation, in order to store data chronologically, I use some cache-buffer. However, it is hard to extend and maintain.

A better way is to use a 2d buffer. It still uses Batch to store data, but the first dimension is env_num. For example, if we collect data from 4 envs by vector-env, the replay buffer will store them as `[env_num=4, timestep, ...]`.

This refactor would greatly simplify the code of collector, but should change some other parts of code (e.g., `buffer.sample()` and `collector.sample()`).

Trinkle23897:
I think it is a prerequisite of #103

youkaichao:
I do not agree with the idea of 2d buffer. The formulation of reinforcement learning only involves a single environment. We use multiple environments for the pure purpose of accelerating simulation. So, conceptually, there is one buffer, one environment. I see no significance in distinguishing each environment.

I think we can maintain a main buffer, with multiple cache buffers, which may be enough for almost all use-cases.

youkaichao:
If one really wants to distinguish between multiple environments, then trajectories from each enviroment should be sampled independently, which can be achieved by instantiating multiple ``collector``.

That said, it seems one buffer for ``collector`` is enough. No need to support 2d buffer.

![image](https://user-images.githubusercontent.com/23236638/86992442-0f2c2400-c1d4-11ea-8d4d-1fb61a7e12be.png)

duburcqa:
I get your point, and honestly I don't really care if it works at the end, but I disagree with your reasoning. There is not requirement for the implementation to match the theory of the __current__ state of the art. The implementation must be designed to be easy to maintain and to understand, and computationally efficient. I don't see any reason why to stick that close to the theory: if it works, it works. To me these are extra constraints only for the sake of elegance, which is the curse of the researcher not the programmer. So maybe multiple-dimensional buffer is a bad idea, because perhaps it is neither "easy to maintain and to understand, and computationally efficient", I don't know, but if it is, to me it is a very bad idea not to take advantage of this.

youkaichao:
[...]

PR #106, opened 2020-06-29
Advanced Batch slicing & minor fix of RNN support

### LINKED ISSUE #104: Advance slicing method of Batch
In recurrent-style training, it stores data in the format of `[batch_size, stack_num, ...]`. And if we want to get the first hidden state, `batch[:, 0]` is the best way instead of manually and recursively get the value in the batch. (#19)

Also with the multi-agent training scenario (someone is working on it), the batch-data will store in the format of `[env_num, agent_num, ...]`. If we want to get data from a specific agent's perspective, the best way is `batch[:, agent_id]`.

@duburcqa

Trinkle23897:
Related to #104

### DIFF
--- a/test/base/test_batch.py
+++ b/test/base/test_batch.py
@@ -150,14 +150,16 @@ def test_batch_cat_and_stack_and_empty():
     assert np.allclose(b5.b.c, [2, 0])
     assert np.allclose(b5.b.d, [1, 0])
     data = Batch(a=[False, True],
-                 b={'c': [2., 'st'], 'd': [1, None], 'e': [2., 1.]},
-                 c=np.array([1, 3, 4], dtype=np.int))
+                 b={'c': [2., 'st'], 'd': [1, None], 'e': [2., float('nan')]},
+                 c=np.array([1, 3, 4], dtype=np.int),
+                 t=torch.tensor([4, 5, 6, 7.]))
     data[-1] = Batch.empty(data[1])
     assert np.allclose(data.c, [1, 3, 0])
     assert np.allclose(data.a, [False, False])
     assert list(data.b.c) == ['2.0', '']
     assert list(data.b.d) == [1, None]
     assert np.allclose(data.b.e, [2, 0])
+    assert torch.allclose(data.t, torch.tensor([4, 5, 6, 0.]))
     b0 = Batch()
     b0.empty_()
     assert b0.shape == []
--- a/test/base/test_batch.py
+++ b/test/base/test_batch.py
@@ -46,8 +46,6 @@ def test_batch():
     with pytest.raises(IndexError):
         batch2[1]
     assert batch2[0].shape[0] == 1
-    with pytest.raises(IndexError):
-        batch2[0][0]
     with pytest.raises(TypeError):
         len(batch2[0])
     assert isinstance(batch2[0].a.c, np.ndarray)
@@ -104,7 +102,7 @@ def test_batch_over_batch():
     assert batch4.a.b[0, 0] == 1.0
 
 
-def test_batch_cat_and_stack():
+def test_batch_cat_and_stack_and_empty():
     b1 = Batch(a=[{'b': np.float64(1.0), 'd': Batch(e=np.array(3.0))}])
     b2 = Batch(a=[{'b': np.float64(4.0), 'd': {'e': np.array(6.0)}}])
     b12_cat_out = Batch.cat((b1, b2))
@@ -133,6 +131,10 @@ def test_batch_cat_and_stack():
     assert np.all(b5.b.c == np.stack([e['b']['c'] for e in b5_dict], axis=0))
     assert b5.b.d[0] == b5_dict[0]['b']['d']
     assert b5.b.d[1] == 0.0
+    b5[1] = Batch.empty(b5[0])
+    assert np.allclose(b5.a, [False, False])
+    assert np.allclose(b5.b.c, [2, 0])
+    assert np.allclose(b5.b.d, [1, 0])
 
 
 def test_batch_over_batch_to_torch():
@@ -215,3 +217,5 @@ if __name__ == '__main__':
     test_utils_to_torch()
     test_batch_pickle()
     test_batch_from_to_numpy_without_copy()
+    test_batch_numpy_compatibility()
+    test_batch_cat_and_stack_and_empty()
--- a/test/base/test_batch.py
+++ b/test/base/test_batch.py
@@ -150,13 +150,14 @@ def test_batch_cat_and_stack_and_empty():
     assert np.allclose(b5.b.c, [2, 0])
     assert np.allclose(b5.b.d, [1, 0])
     data = Batch(a=[False, True],
-                 b={'c': [2., 'st'], 'd': [1, None]},
+                 b={'c': [2., 'st'], 'd': [1, None], 'e': [2., 1.]},
                  c=np.array([1, 3, 4], dtype=np.int))
     data[-1] = Batch.empty(data[1])
     assert np.allclose(data.c, [1, 3, 0])
     assert np.allclose(data.a, [False, False])
     assert list(data.b.c) == ['2.0', '']
     assert list(data.b.d) == [1, None]
+    assert np.allclose(data.b.e, [2, 0])
     b0 = Batch()
     b0.empty_()
     assert b0.shape == []
--- a/test/base/test_batch.py
+++ b/test/base/test_batch.py
@@ -13,9 +13,9 @@ def test_batch():
     batch.obs = [1]
     assert batch.obs == [1]
     batch.cat_(batch)
-    assert batch.obs == [1, 1]
+    assert np.allclose(batch.obs, [1, 1])
     assert batch.np.shape == (6, 4)
-    assert batch[0].obs == batch[1].obs
+    assert np.allclose(batch[0].obs, batch[1].obs)
     batch.obs = np.arange(5)
     for i, b in enumerate(batch.split(1, shuffle=False)):
         if i != 5:
@@ -39,14 +39,14 @@ def test_batch():
         'c': np.zeros(1),
         'd': Batch(e=np.array(3.0))}])
     assert len(batch2) == 1
-    assert Batch().size == 0
-    assert batch2.size == 1
+    assert Batch().shape == []
+    assert batch2.shape[0] == 1
     with pytest.raises(IndexError):
         batch2[-2]
     with pytest.raises(IndexError):
         batch2[1]
-    assert batch2[0].size == 1
-    with pytest.raises(TypeError):
+    assert batch2[0].shape[0] == 1
+    with pytest.raises(IndexError):
         batch2[0][0]
     with pytest.raises(TypeError):
         len(batch2[0])
@@ -87,18 +87,18 @@ def test_batch_over_batch():
     batch2.b.b[-1] = 0
     print(batch2)
     for k, v in batch2.items():
-        assert batch2[k] == v
+        assert np.all(batch2[k] == v)
     assert batch2[-1].b.b == 0
     batch2.cat_(Batch(c=[6, 7, 8], b=batch))
-    assert batch2.c == [6, 7, 8, 6, 7, 8]
-    assert batch2.b.a == [3, 4, 5, 3, 4, 5]
-    assert batch2.b.b == [4, 5, 0, 4, 5, 0]
+    assert np.allclose(batch2.c, [6, 7, 8, 6, 7, 8])
+    assert np.allclose(batch2.b.a, [3, 4, 5, 3, 4, 5])
+    assert np.allclose(batch2.b.b, [4, 5, 0, 4, 5, 0])
     d = {'a': [3, 4, 5], 'b': [4, 5, 6]}
     batch3 = Batch(c=[6, 7, 8], b=d)
     batch3.cat_(Batch(c=[6, 7, 8], b=d))
-    assert batch3.c == [6, 7, 8, 6, 7, 8]
-    assert batch3.b.a == [3, 4, 5, 3, 4, 5]
-    assert batch3.b.b == [4, 5, 6, 4, 5, 6]
+    assert np.allclose(batch3.c, [6, 7, 8, 6, 7, 8])
+    assert np.allclose(batch3.b.a, [3, 4, 5, 3, 4, 5])
+    assert np.allclose(batch3.b.b, [4, 5, 6, 4, 5, 6])
     batch4 = Batch(({'a': {'b': np.array([1.0])}},))
     assert batch4.a.b.ndim == 2
     assert batch4.a.b[0, 0] == 1.0
--- a/test/base/test_batch.py
+++ b/test/base/test_batch.py
@@ -45,7 +45,9 @@ def test_batch():
         batch2[-2]
     with pytest.raises(IndexError):
         batch2[1]
-    assert batch2[0].shape[0] == 1
+    assert batch2[0].shape == []
+    with pytest.raises(IndexError):
+        batch2[0][0]
     with pytest.raises(TypeError):
         len(batch2[0])
     assert isinstance(batch2[0].a.c, np.ndarray)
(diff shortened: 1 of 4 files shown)


Trinkle23897:
@duburcqa I add as what you said, but it turns out that it is not compatible with tests in `test/base`.

Some issues:
- min(*data_shape) will return the smallest data_shape, not the smallest size in every dimension. E.g., `min((1,2,3,4), (3,2,1))` will result in `(1,2,3,4)` instead of `(1,2,1)`;
- if we want to force use np.array instead of list, we should further overwrite the `__setattr__` method;
- some tests fail strangely, `TypeError: zip argument #2 must support iteration`, and in "test_collector_with_dict_state" it fails to deep-copy. Should we change all the list in test to np.ndarray or add extra except-handling method within Batch?

One more thing: ListReplayBuffer use list instead of np.ndarray. To entirely change list to np.ndarray, this class should be abandoned, which will related to #105 ?

Trinkle23897:
https://github.com/thu-ml/tianshou/blob/c639446c663487a3bc24d138b86c64bc43d33dad/tianshou/data/collector.py#L203-L204
I haven't figured it out the best practice to make it compatible with _create_value (None issue). Maybe we should create `batch.zero_()`?

tianshou/data/batch.py
```diff
@@ -240,6 +242,8 @@ def __init__(self,
                 if isinstance(v, dict) or _is_batch_set(v):
                     self.__dict__[k] = Batch(v)
                 else:
+                    if isinstance(v, list):
+                        v = np.array(v, copy=True)
```
duburcqa: Why to copy ? I don't think you should

tianshou/data/batch.py
```diff
@@ -534,21 +538,23 @@ def __len__(self) -> int:
...
-                    r.append(1)
-            return min(r) if len(r) > 0 else 0
+                try:
+                    if len(v.shape) > 0:
+                        data_shape.append(v.shape)
+                except AttributeError:
+                    if isinstance(v, list):
+                        data_shape.append(tuple([len(v)]))
+                    else:
+                        raise TypeError("No support for 'shape' method with "
```
duburcqa: Add would rather use `__setitem__` and `__setattr__` to make sure that `list` are converted to numpy, so that this check can be avoided.

duburcqa:
> Maybe we should create batch.zero_()

Yes it would be nice, or rather `empty_`, since it is doing something closer to `empty` than `zero`.

Trinkle23897:
> * some tests fail strangely, `TypeError: zip argument #2 must support iteration`, and in "test_collector_with_dict_state" it fails to deep-copy. Should we change all the list in test to np.ndarray or add extra except-handling method within Batch?
> 
> One more thing: ListReplayBuffer use list instead of np.ndarray. To entirely change list to np.ndarray, this class should be abandoned, which will related to #105 ?

How about these issue?

duburcqa:
> min(*data_shape) will return the smallest data_shape, not the smallest size in every dimension. E.g., min((1,2,3,4), (3,2,1)) will result in (1,2,3,4) instead of (1,2,1);

True. It should be replaced by `list(map(min, zip(*data_shape)))`

duburcqa:
> if we want to force use np.array instead of list, we should further overwrite the __setattr__ method;

Yes, and I think it is a good thing to better control the data added to Batch instances or modified.

duburcqa:
> some tests fail strangely, TypeError: zip argument #2 must support iteration, and in "test_collector_with_dict_state" it fails to deep-copy. Should we change all the list in test to np.ndarray or add extra except-handling method within Batch?

I think it is related to `__setattr__` issue, not converting list to array.

duburcqa:
> One more thing: ListReplayBuffer use list instead of np.ndarray. To entirely change list to np.ndarray, this class should be abandoned.

Good point. I don't know how to handle this, but it is only used by `Collector`, that you are refactoring, so maybe it would be better to refactor this part. I would only a raw list of Batch/dict, replacing:
```
self._cached_buf[i].add(**data)
```
by simply
```
self._cached_buf[i].append(Batch(data))
```

Trinkle23897:
> > One more thing: ListReplayBuffer use list instead of np.ndarray. To entirely change list to np.ndarray, this class should be abandoned.
> 
> Good point. I don't know how to handle this, but it is only used by `Collector`, that you are refactoring, so maybe it would be better to refactor this part. I would only a raw list of Batch/dict, replacing:
> 
> ```
> self._cached_buf[i].add(**data)
> ```
> 
> by simply
> 
> ```
> self._cached_buf[i].append(Batch(data))
> ```

I found an interesting fact:
```python
        if self._meta.__dict__.get(name, None) is None:
            self._meta.__dict__[name] = []
        self._meta.__dict__[name].append(inst)
```
this will bypass both of the `__setitem__` and `__setattr__` method. So the ListReplayBuffer would not be changed.

Trinkle23897:
Why did you use @staticmethod in Batch.stack, @classmethod in Batch.cat?

duburcqa:
> this will bypass both of the __setitem__ and __setattr__ method. So the ListReplayBuffer would not be changed.

Yes, but in fact it would be better if we could use `self._meta[name]` instead of `self._meta.__dict__[name]`, precisely to avoid such bypass behavior, but I was afraid of the impact on the performances (yet I haven't tried).

> Why did you use `@staticmethod` in Batch.stack, `@classmethod` in Batch.cat?

No reason, `@staticmethod` is enough for both.

Trinkle23897:
@duburcqa It should be ready now. Please have a code review at your convenience.
By the way, I have not removed the redundant code. You can comment and point them out (which will be added in this pr).

tianshou/data/batch.py
```diff
@@ -511,13 +531,35 @@ def stack_(self,
...
-        """Stack a :class:`~tianshou.data.Batch` object into a
-        single new batch.
+        """Stack a :class:`~tianshou.data.Batch` object into a single new
+        batch.
         """
         batch = Batch()
         batch.stack_(batches, axis)
         return batch
 
+    def empty_(self) -> None:
```
duburcqa: It should return the updated Batch

tianshou/data/batch.py
```diff
@@ -534,21 +576,24 @@ def __len__(self) -> int:
...
+                    if len(v.shape) > 0:
+                        data_shape.append(v.shape)
+                except AttributeError:
+                    if isinstance(v, list):
+                        data_shape.append(tuple([len(v)]))
+                    else:
+                        raise TypeError("No support for 'shape' method with "
+                                        f"type {type(v)} in class Batch.")
+            return list(map(min, zip(*data_shape))) if len(data_shape) > 1 \
+                else data_shape[0] if len(data_shape) > 0 else []
```
duburcqa: No need for `data_shape[0] if len(data_shape) > 0 else []` since `len(data_shape)` cannot be 0.

tianshou/data/batch.py
```diff
@@ -534,21 +576,24 @@ def __len__(self) -> int:
...
-                        v, (np.ndarray, torch.Tensor)) or v.ndim > 0):
-                    r.append(len(v))
-                else:
-                    r.append(1)
-            return min(r) if len(r) > 0 else 0
+                try:
+                    if len(v.shape) > 0:
+                        data_shape.append(v.shape)
+                except AttributeError:
+                    if isinstance(v, list):
```
duburcqa: It cannot be `list`

tianshou/data/batch.py
```diff
@@ -534,21 +576,24 @@ def __len__(self) -> int:
...
-                if isinstance(v, Batch):
-                    r.append(v.size)
-                elif hasattr(v, '__len__') and (not isinstance(
-                        v, (np.ndarray, torch.Tensor)) or v.ndim > 0):
-                    r.append(len(v))
-                else:
-                    r.append(1)
-            return min(r) if len(r) > 0 else 0
+                try:
+                    if len(v.shape) > 0:
```
duburcqa: Why to do an exception for scalar ? I don't think it is a good idea since it does not support slicing, and it is not consistent with `len` (`len(v) != v.shape[0]`)

tianshou/data/batch.py
Trinkle23897: Nope.
```python
In [3]: batch2 = Batch(a=[{ 
   ...:         'b': np.float64(1.0), 
   ...:         'c': np.zeros(1), 
   ...:         'd': Batch(e=np.array(3.0))}])

In [6]: batch2[0]
Out[6]: 
Batch(
    a: Batch(
           d: Batch(
                  e: 3.0,
              ),
           c: array([0.]),
           b: 1.0,
       ),
)
```
where batch2[0].shape will cause data_shape == [].

tianshou/data/batch.py
Trinkle23897: I think Batch.empty is doing this thing instead of Batch.empty_

tianshou/data/batch.py
```diff
@@ -511,13 +531,35 @@ def stack_(self,
...
         batch.stack_(batches, axis)
         return batch
 
+    def empty_(self) -> None:
+        """Empty a :class:`~tianshou.data.Batch` object with 0 or ``None``
+        filled.
+        """
+        for k, v in self.items():
+            if isinstance(v, np.ndarray) and v.dtype == np.object:
+                self.__dict__[k].fill(None)
```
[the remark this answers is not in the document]
Trinkle23897: I've not got it. Could you give an example?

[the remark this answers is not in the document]
duburcqa: It fails if `isinstance(dtype, (np.integer, int))`, for example: `np.array([1]).fill(None)`

I think it is more practical to use None only for object array, since it is usually more appropriate to have 0 for unknown data. For instance, if you do averaging of info data, missing data must be ignored, and using 0.0 is handy instead of having explicitly skip `nan` values.
```
self.__dict__[k].fill(None if self.__dict__[k].dtype == np.object else 0)
```

tianshou/data/batch.py
duburcqa: No, both should. See for instance https://pytorch.org/docs/stable/tensors.html#torch.Tensor.zero_

duburcqa:
Very good job ! Still 2 comments to address :grin:

duburcqa:
I was thinking `Batch.__init__` could have an extra argument `copy` similar to numpy (default to False), to enforce `deepcopy` of input argument. It could be convenient and it is only 2 lines of code.

duburcqa:
By the way, did you do some profiling to check the performances ?

tianshou/data/batch.py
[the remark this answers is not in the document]
Trinkle23897: But np.array([1]) is `dtype==np.int64` instead of np.object, which will not get into line 547.

Trinkle23897:
> By the way, did you do some profiling to check the performances?

On my computer, it is the same as the last commit. However, GitHub shows that after the third commit of this PR, it shows down to some degree. I think maybe it is because of slicing: every time the `__getitem__` will recursively call the `_valid_bounds`.

duburcqa:
output

Apparently it is fine, at least for PPO.

test/base/test_batch.py
```diff
@@ -133,6 +145,21 @@ def test_batch_cat_and_stack():
     assert np.all(b5.b.c == np.stack([e['b']['c'] for e in b5_dict], axis=0))
     assert b5.b.d[0] == b5_dict[0]['b']['d']
     assert b5.b.d[1] == 0.0
+    b5[1] = Batch.empty(b5[0])
+    assert np.allclose(b5.a, [False, False])
+    assert np.allclose(b5.b.c, [2, 0])
+    assert np.allclose(b5.b.d, [1, 0])
+    data = Batch(a=[False, True],
+                 b={'c': [2., 'st'], 'd': [1, None]},
```
duburcqa: `np.array([2., 1.])` or `[2., 1.]` would be better to test, since `'c': [2., 'st']` is similar to `'d': [1, None]`

duburcqa (approved):

tianshou/data/batch.py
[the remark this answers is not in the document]
duburcqa: Yes you are right, I missed that point somehow x)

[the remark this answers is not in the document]
duburcqa: Yet, you could have used `fill` for any numpy array using `self.__dict__[k].fill(None if self.__dict__[k].dtype == np.object else 0)`, which is a bit clearer than `* 0` to me, but as you wish !

duburcqa (approved):

duburcqa (approved):

tianshou/data/batch.py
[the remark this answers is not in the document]
Trinkle23897: But *0 can cause another issue: if a==inf or a==nan, *0 will still be nan instead of 0.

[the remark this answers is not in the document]
duburcqa: So are you fixing line 557 in accordance then ?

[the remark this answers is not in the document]
Trinkle23897: Yes, but I found it is non-trivial: how can we determine whether a np.array has flexible type?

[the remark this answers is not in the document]
duburcqa: What do you mean by "flexible" type ? `np.object` is for objects, `np.inexact` is for floats, and `np.integer` is for integer. For me, we should use None only for `np.object`, 0 otherwise.

tianshou/data/batch.py
```diff
@@ -511,13 +530,42 @@ def stack_(self,
...
+            if v is None:
+                continue
+            if isinstance(v, Batch):
+                self.__dict__[k].empty_()
+            elif isinstance(v, np.ndarray) and v.dtype == np.object:
+                self.__dict__[k].fill(None)
+            elif isinstance(v, torch.Tensor):  # cannot apply fill_ directly
+                self.__dict__[k] = torch.zeros_like(self.__dict__[k])
+            else:  # np
+                self.__dict__[k] *= 0
```
duburcqa, line 554: Why not `self.__dict__[k].fill(0)` ?

tianshou/data/batch.py
```diff
@@ -511,13 +530,42 @@ def stack_(self,
...
+        """
+        for k, v in self.items():
+            if v is None:
+                continue
+            if isinstance(v, Batch):
+                self.__dict__[k].empty_()
+            elif isinstance(v, np.ndarray) and v.dtype == np.object:
+                self.__dict__[k].fill(None)
+            elif isinstance(v, torch.Tensor):  # cannot apply fill_ directly
+                self.__dict__[k] = torch.zeros_like(self.__dict__[k])
```
duburcqa, line 552: You could use `fill_` for torch tensors, it is more robust (reserve dtype and device) and avoid memory allocation.

test/base/test_batch.py
```diff
@@ -133,6 +145,24 @@ def test_batch_cat_and_stack():
...
+                 b={'c': [2., 'st'], 'd': [1, None], 'e': [2., float('nan')]},
+                 c=np.array([1, 3, 4], dtype=np.int),
+                 t=torch.tensor([4, 5, 6, 7.]))
+    data[-1] = Batch.empty(data[1])
+    assert np.allclose(data.c, [1, 3, 0])
+    assert np.allclose(data.a, [False, False])
+    assert list(data.b.c) == ['2.0', '']
+    assert list(data.b.d) == [1, None]
+    assert np.allclose(data.b.e, [2, 0])
+    assert torch.allclose(data.t, torch.tensor([4, 5, 6, 0.]))
```
Trinkle23897, line 162: Here, if we use fill_, the result will be [4, 0, 6, 0]. @duburcqa

tianshou/data/batch.py
Trinkle23897: because I want the np.array(['str', 'str']) to be np.array(['', 'str']) after empty

Trinkle23897:
I'd like to merge because I have other things to do (two projects and one deadline tomorrow). You can create a new PR to adapt your thought, and I can make a code review for you :)

Something to do:
- change Batch.empty to in-place fill
- Batch(copy=True)

test/base/test_batch.py
duburcqa: What the heck ? `torch.tensor([4, 5, 6, 7.]).fill_(0) == torch.tensor([0., 0., 0., 0.])` on by computer

Trinkle23897: I mean the test in test/base/test_batch.py line 162, where the operation is `data[-1] = Batch.empty(data[1])`. It simulates an undefined behavior.

PR #110, opened 2020-07-06
change Batch.empty to in-place fill; add copy option for Batch construction

youkaichao:
    + [ ] exception-raising fix
    + [x] new feature

  ```python
  import tianshou, torch, sys
  ```

### DIFF
--- a/test/base/test_batch.py
+++ b/test/base/test_batch.py
@@ -116,7 +116,7 @@ def test_batch_over_batch():
     assert np.allclose(batch5.b.c.squeeze(), [[0, 1]] * 3)
 
 
-def test_batch_cat_and_stack_and_empty():
+def test_batch_cat_and_stack():
     b1 = Batch(a=[{'b': np.float64(1.0), 'd': Batch(e=np.array(3.0))}])
     b2 = Batch(a=[{'b': np.float64(4.0), 'd': {'e': np.array(6.0)}}])
     b12_cat_out = Batch.cat((b1, b2))
@@ -145,26 +145,6 @@ def test_batch_cat_and_stack_and_empty():
     assert np.all(b5.b.c == np.stack([e['b']['c'] for e in b5_dict], axis=0))
     assert b5.b.d[0] == b5_dict[0]['b']['d']
     assert b5.b.d[1] == 0.0
-    b5[1] = Batch.empty(b5[0])
-    assert np.allclose(b5.a, [False, False])
-    assert np.allclose(b5.b.c, [2, 0])
-    assert np.allclose(b5.b.d, [1, 0])
-    data = Batch(a=[False, True],
-                 b={'c': np.array([2., 'st'], dtype=np.object),
-                    'd': [1, None],
-                    'e': [2., float('nan')]},
-                 c=np.array([1, 3, 4], dtype=np.int),
-                 t=torch.tensor([4, 5, 6, 7.]))
-    data[-1] = Batch.empty(data[1])
-    assert np.allclose(data.c, [1, 3, 0])
-    assert np.allclose(data.a, [False, False])
-    assert list(data.b.c) == [2.0, None]
-    assert list(data.b.d) == [1, None]
-    assert np.allclose(data.b.e, [2, 0])
-    assert torch.allclose(data.t, torch.tensor([4, 5, 6, 0.]))
-    b0 = Batch()
-    b0.empty_()
-    assert b0.shape == []
 
 
 def test_batch_over_batch_to_torch():
@@ -228,8 +208,8 @@ def test_batch_from_to_numpy_without_copy():
 
 
 def test_batch_copy():
-    batch = Batch(a=[3, 4, 5], b=[4, 5, 6])
-    batch2 = Batch({'c': [6, 7, 8], 'b': batch})
+    batch = Batch(a=np.array([3, 4, 5]), b=np.array([4, 5, 6]))
+    batch2 = Batch({'c': np.array([6, 7, 8]), 'b': batch})
     batch3 = Batch(batch2, copy=True)
     assert batch2.c is not batch3.c
     assert batch2.b is not batch3.b
@@ -243,6 +223,29 @@ def test_batch_copy():
 
 
 def test_batch_empty():
+    b5_dict = np.array([{'a': False, 'b': {'c': 2.0, 'd': 1.0}},
+                        {'a': True, 'b': {'c': 3.0}}])
+    b5 = Batch(b5_dict)
+    b5[1] = Batch.empty(b5[0])
+    assert np.allclose(b5.a, [False, False])
+    assert np.allclose(b5.b.c, [2, 0])
+    assert np.allclose(b5.b.d, [1, 0])
+    data = Batch(a=[False, True],
+                 b={'c': np.array([2., 'st'], dtype=np.object),
+                    'd': [1, None],
+                    'e': [2., float('nan')]},
+                 c=np.array([1, 3, 4], dtype=np.int),
+                 t=torch.tensor([4, 5, 6, 7.]))
+    data[-1] = Batch.empty(data[1])
+    assert np.allclose(data.c, [1, 3, 0])
+    assert np.allclose(data.a, [False, False])
+    assert list(data.b.c) == [2.0, None]
+    assert list(data.b.d) == [1, None]
+    assert np.allclose(data.b.e, [2, 0])
+    assert torch.allclose(data.t, torch.tensor([4, 5, 6, 0.]))
+    b0 = Batch()
+    b0.empty_()
+    assert b0.shape == []
     batch = Batch(a=[object(), object()])
     batch_e = Batch.empty(batch)
     assert batch_e.a[0] is None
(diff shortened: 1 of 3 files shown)


duburcqa:
Give me a few minutes to review your PR please.

test/base/test_batch.py
```diff
@@ -150,13 +150,15 @@ def test_batch_cat_and_stack_and_empty():
...
+                 b={'c': np.array([2., 'st'], dtype=np.object),
+                    'd': [1, None],
+                    'e': [2., float('nan')]},
                  c=np.array([1, 3, 4], dtype=np.int),
                  t=torch.tensor([4, 5, 6, 7.]))
     data[-1] = Batch.empty(data[1])
     assert np.allclose(data.c, [1, 3, 0])
     assert np.allclose(data.a, [False, False])
-    assert list(data.b.c) == ['2.0', '']
+    assert list(data.b.c) == [2.0, None]
```
Trinkle23897: Add a testcase using `Batch.empty_`

tianshou/data/batch.py
```diff
@@ -243,6 +243,10 @@ class Batch:
     Convenience helpers are available to convert in-place the stored data into
     Numpy arrays or Torch tensors.
 
+    For Numpy arrays, only data types with ``np.object`` and numbers are
+    supported. For strings or other data types, however, they can be held
+    in ``np.object`` arrays.
+
```
Trinkle23897: Move to line 71 or 88

test/base/test_batch.py
```diff
@@ -225,6 +227,21 @@ def test_batch_from_to_numpy_without_copy():
...
+def test_batch_copy():
+    batch = Batch(a=[3, 4, 5], b=[4, 5, 6])
+    batch2 = Batch({'c': [6, 7, 8], 'b': batch})
+    batch3 = Batch(batch2, copy=True)
+    assert batch2.c is not batch3.c
+    assert batch2.b is not batch3.b
+    assert batch2.b.a is not batch3.b.a
+    assert batch2.b.b is not batch3.b.b
+
+    batch = Batch(a=np.array([Batch(), Batch()], dtype=np.object))
```
duburcqa: rather use `[Batch(), Batch()]`. Nobody is going to use `np.array([Batch(), Batch()], dtype=np.object)` explicitly.

tianshou/data/batch.py
```diff
@@ -546,14 +553,19 @@ def empty_(self) -> 'Batch':
...
-                    self.__dict__[k] = np.nan_to_num(self.__dict__[k])
+                self.__dict__[k].zero_()
+            elif isinstance(v, np.ndarray):
+                if v.dtype == np.object:
+                    self.__dict__[k].fill(None)
+                else:  # np
+                    self.__dict__[k].fill(0)
+            else:
+                # scalar value
+                if isinstance(v, np.number) or isinstance(v, np.bool_):
```
duburcqa: rather use `isinstance(v, (np.generic, Number))` (if it is working as expected of course).

tianshou/data/batch.py
```diff
@@ -546,14 +553,19 @@ def empty_(self) -> 'Batch':
...
+                    self.__dict__[k] = self.__dict__[k].__class__(0)
```
duburcqa: `self.__dict__[k] *= 0` is better in that particular case, since memory allocation must be avoided.

test/base/test_batch.py
```diff
@@ -225,6 +227,21 @@ def test_batch_from_to_numpy_without_copy():
     assert c_mem_addr_new == c_mem_addr_orig
 
 
+def test_batch_copy():
+    batch = Batch(a=[3, 4, 5], b=[4, 5, 6])
```
duburcqa: Your test does not make sense. You are dealing with immutable data, so it will be copy when converted in `np.ndarray` by `Batch` constructor in any case. And you should check that `copy=False` is actually not copying the data. To this end, you have to use `batch = Batch(a=np.array([3, 4, 5]), b=np.array([4, 5, 6]))` instead. 

And be very careful when checking copy of `np.ndarray`, see the other test `test_batch_from_to_numpy_without_copy` for a robust way to do it. Using `is` is not enough.

test/base/test_batch.py
```diff
@@ -225,6 +227,21 @@ def test_batch_from_to_numpy_without_copy():
...
+    batch2 = Batch({'c': [6, 7, 8], 'b': batch})
```
duburcqa: same for `'c'=[6, 7, 8]`. Rather use `np.array([6, 7, 8])`

test/base/test_batch.py
youkaichao: Added in 1a35bea31cb6273056fcf17857fbd5ccf42d8434 .

tianshou/data/batch.py
youkaichao: Fixed in 0b5320e64aaf8ab36482cb50294fd4ed75bf0996 .

test/base/test_batch.py
youkaichao: Fixed in 050f5691e2ea8ce5c7ff0270c0b191334a4eab34 .

tianshou/data/batch.py
youkaichao: Scalar values in numpy are immutable. Here memory allocation cannot be avoided.
```
tmp = np.float64(1.0)
tmp
Out[160]: 1.0
tmp2 = tmp
tmp2 *= 0
tmp
Out[163]: 1.0
```

Trinkle23897: It is an issue of numpy scalar value:
```python
In [2]: a=np.array([1,1,3])

In [3]: a[0].fill(0)

In [4]: a
Out[4]: array([1, 1, 3])
```

duburcqa: > Scalar values in numpy are immutable. 

Good to know. Indeed it looks like a bug to me (fortunately, Torch scalar are working as expected) Anyway, I still find it clearer to do `self.__dict__[k] *= 0`, and it works with @Trinkle23897 example:
```
In [2]: a=np.array([1,1,3])

In [3]: a[0] *= 0

In [4]: a
Out[4]: array([0, 1, 3])
```

Trinkle23897: Yes but it will fail in this case:
```python
In [82]: d=Batch(a=[1,2,3])

In [83]: d[0].empty_()
Out[83]: 
Batch(
    a: 0,
)

In [84]: d
Out[84]: 
Batch(
    a: array([1, 2, 3]),
)
```
I'm implementing empty_(index=...) and this method can avoid the above issue:
```python
d.empty_(index=0)
```

Trinkle23897:
@duburcqa I update the code and you can have a review.

tianshou/data/batch.py
```diff
@@ -537,33 +545,42 @@ def stack(batches: List['Batch'], axis: int = 0) -> 'Batch':
...
+    def empty(batch: 'Batch', index: Union[
+        str, slice, int, np.integer, np.ndarray, List[int]] = None
+    ) -> 'Batch':
         """Return an empty :class:`~tianshou.data.Batch` object with 0 or
         ``None`` filled, the shape is the same as the given
         :class:`~tianshou.data.Batch`.
         """
-        batch = Batch(**batch)
-        batch.empty_()
+        batch = Batch(**batch, copy=True)
```
duburcqa: Straight `batch = deepcopy(batch)` should be faster, and consistent with `__mul__` ...etc

tianshou/data/batch.py
duburcqa: Hum... it's nice that `d.empty_(index=0)` is working fine, but it is very unfortunate that `d[0].empty_()` is not. I think it is a major source of error for the user (as it is for numpy).

tianshou/data/batch.py
```diff
@@ -537,33 +545,42 @@ def stack(batches: List['Batch'], axis: int = 0) -> 'Batch':
...
+            elif isinstance(v, torch.Tensor):
+                self.__dict__[k][index] = 0
+            elif isinstance(v, np.ndarray):
+                if v.dtype == np.object:
+                    self.__dict__[k][index] = None
+                else:
+                    self.__dict__[k][index] = 0
+            else:  # scalar value
+                if isinstance(v, (np.generic, Number)):
+                    self.__dict__[k] = 0
```
duburcqa: It does not preserve `dtype`, but `self.__dict__[k] *= 0` does.

tianshou/data/batch.py
Trinkle23897: I add a warning.

tianshou/data/batch.py
Trinkle23897: But *=0 will not resolve the nan issue:
```python
b = nan
b *= 0
# b is still nan
```

Trinkle23897:
@duburcqa any more comments?

duburcqa (approved):

tianshou/data/batch.py
duburcqa: Yes but it is an issue ? But I got your point !

duburcqa:
No it's fine ! Good job to both of you :+1:

Trinkle23897 (approved):

PR #111, opened 2020-07-06
Fix to_torch converters

duburcqa:
Fix `np.generic` variables not being properly converted to torch Tensor, and even raising an exception for `Batch.to_torch`.

### DIFF
--- a/tianshou/data/utils.py
+++ b/tianshou/data/utils.py
@@ -24,7 +24,9 @@ def to_torch(x: Union[torch.Tensor, dict, Batch, np.ndarray],
              device: Union[str, int, torch.device] = 'cpu'
              ) -> Union[dict, Batch, torch.Tensor]:
     """Return an object without np.ndarray."""
-    if isinstance(x, np.ndarray):
+    if isinstance(x, (np.generic, np.ndarray)):
+        if isinstance(x, np.generic):
+            x = np.asanyarray(x)
         x = torch.from_numpy(x).to(device)
         if dtype is not None:
             x = x.type(dtype)
--- a/tianshou/data/utils.py
+++ b/tianshou/data/utils.py
@@ -1,5 +1,6 @@
 import torch
 import numpy as np
+from numbers import Number
 from typing import Union, Optional
 
 from tianshou.data import Batch
@@ -24,12 +25,6 @@ def to_torch(x: Union[torch.Tensor, dict, Batch, np.ndarray],
              device: Union[str, int, torch.device] = 'cpu'
              ) -> Union[dict, Batch, torch.Tensor]:
     """Return an object without np.ndarray."""
-    if isinstance(x, (np.generic, np.ndarray)):
-        if isinstance(x, np.generic):
-            x = np.asanyarray(x)
-        x = torch.from_numpy(x).to(device)
-        if dtype is not None:
-            x = x.type(dtype)
     if isinstance(x, torch.Tensor):
         if dtype is not None:
             x = x.type(dtype)
@@ -39,6 +34,16 @@ def to_torch(x: Union[torch.Tensor, dict, Batch, np.ndarray],
             x[k] = to_torch(v, dtype, device)
     elif isinstance(x, Batch):
         x.to_torch(dtype, device)
+    elif isinstance(x, (np.number, np.bool_, Number)):
+        x = to_torch(np.asanyarray(x), dtype, device)
+    elif isinstance(x, list) and len(x) > 0 and \
+            isinstance(x[0], (np.number, np.bool_, Number)):
+        x = to_torch(np.asanyarray(x), dtype, device)
+    elif isinstance(x, np.ndarray) and \
+            isinstance(x.item(0), (np.number, np.bool_, Number)):
+        x = torch.from_numpy(x).to(device)
+        if dtype is not None:
+            x = x.type(dtype)
     return x
 
 
--- a/test/base/test_batch.py
+++ b/test/base/test_batch.py
@@ -149,9 +149,9 @@ def test_batch_cat_and_stack():
 
 def test_batch_over_batch_to_torch():
     batch = Batch(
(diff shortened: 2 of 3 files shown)


tianshou/data/utils.py
```diff
@@ -37,6 +37,8 @@ def to_torch(x: Union[torch.Tensor, dict, Batch, np.ndarray],
             x[k] = to_torch(v, dtype, device)
     elif isinstance(x, Batch):
         x.to_torch(dtype, device)
+    elif isinstance(x, np.generic):
```
duburcqa: Maybe it would be even better to use `else` instead of `isinstance(x, np.generic)`, to act as fallback. What do you think @Trinkle23897 ?

tianshou/data/utils.py
```diff
@@ -24,7 +24,9 @@ def to_torch(x: Union[torch.Tensor, dict, Batch, np.ndarray],
              device: Union[str, int, torch.device] = 'cpu'
              ) -> Union[dict, Batch, torch.Tensor]:
     """Return an object without np.ndarray."""
-    if isinstance(x, np.ndarray):
+    if isinstance(x, (np.generic, np.ndarray)):
+        if isinstance(x, np.generic):
+            x = np.asanyarray(x)
```
Trinkle23897: still raise an exception:
```python
In [17]: a=np.array([float('nan'), np.float32])[0]

In [18]: a
Out[18]: nan

In [19]: to_torch(a).dtype
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-19-156efb624c56> in <module>
----> 1 to_torch(a).dtype

AttributeError: 'float' object has no attribute 'dtype'
```

duburcqa: Do you think it should not in this case?

Trinkle23897: Emm...I think it may be a case in Batch.to_torch, but it works well:
```python
In [46]: a=Batch(a=np.array([0., float('nan')], np.float32))

In [47]: to_torch(a[1])
Out[47]: 
Batch(
    a: tensor(nan),
)
```

Trinkle23897 (approved):

tianshou/data/utils.py
duburcqa: No it's because you are confusing 2 different things: you are comparing `np.array([float('nan'), np.float32])`, which is equal to `array([nan, <class 'numpy.float32'>], dtype=object)`, since `np.float32` is a class, not an instance. And when you access an item from a numpy array of `object`, it completely remove numpy wrapping, so `a[0]` is not longer an numpy scalar in the first place, but a native python scalar.

duburcqa: The point is: does `to_torch([0., float('nan')])` should return a tensor ? For `Batch.to_torch` it is not an issue, since the list is automatically converted in `np.ndarray` and therefore converted in `Tensor`.

Trinkle23897: > No it's because you are confusing 2 different things: you are comparing np.array([float('nan'), np.float32]), which is equal to array([nan, <class 'numpy.float32'>], dtype=object), since np.float32 is a class, not an instance. And when you access an item from a numpy array of object, it completely remove numpy wrapping, so a[0] is not longer an numpy scalar in the first place, but a native python scalar.

No, I'm not packing nan with np.float32, instead `np.array([nan], np.float32)`. The second argument in np.array is dtype.
By default, the nan's dtype is `np.float64` and it will not raise such an error. However, the nan with np.foat32 will be converted to float('nan') instead of a numpy scalar when in indexing. The first example shows this issue.

Trinkle23897: > The point is: does to_torch([0., float('nan')]) should return a tensor ? For Batch.to_torch it is not an issue, since the list is automatically converted in np.ndarray and therefore converted in Tensor.

I think it should.

Trinkle23897: Oh sorry, I just notice that my first case is wrong :(

Trinkle23897:
By the way, from my point of view, the force-push is not a good way since it will send another email every time. Plus it will cover up the previous commit so that I cannot find out that log.

duburcqa:
Indeed, but in my opinion it is not a good practice to push many commits that are not functional. The aim of commits is to go back in the history, if you cannot because it is buggy it is not good either. I guess the best approach is to push only when it is ready, but I cannot do that since I'm working on several machine and not testing on the same one that I'm programming. 
Probably I should test on a separated branch before pushing.

duburcqa:
@Trinkle23897 It should be ready now

tianshou/data/utils.py
```diff
@@ -37,6 +34,16 @@ def to_torch(x: Union[torch.Tensor, dict, Batch, np.ndarray],
             x[k] = to_torch(v, dtype, device)
     elif isinstance(x, Batch):
         x.to_torch(dtype, device)
+    elif isinstance(x, (np.number, np.bool_, Number)):
+        x = to_torch(np.asanyarray(x), dtype, device)
+    elif isinstance(x, list) and len(x) > 0 and \
+            isinstance(x[0], (np.number, np.bool_, Number)):
```
Trinkle23897, line 40: Why do you use `isinstance(x, (np.number, np.bool_, Number))`?
Can we use the np array conversion and let numpy automatically determine the type? For example,
```python
if not isinstance(x, (torch.Tensor, np.ndarray)):
    x = np.asanyarray(x)
# then check the type
...
```

duburcqa: No, since `torch.Tensor` only supports the types `(np.number, np.bool_, Number)` :smile: You can check by yourself:
```
torch.tensor(np.asanyarray(['nan', 'er']))
>>> TypeError: can't convert np.ndarray of type numpy.str_. The only supported types are: float64, float32, float16, int64, int32, int16, int8, uint8, and bool.
```

Trinkle23897 (approved):

Trinkle23897:
I've sent you an invitation link.

PR #112, opened 2020-07-07
add type check for each element rather than the first element

### LINKED ISSUE #96: Batch & Buffer profiling

youkaichao:
    + [ ] exception-raising fix
    + [ ] new feature

  ```python
  import tianshou, torch, sys
  ```

### DIFF
--- a/test/base/test_batch.py
+++ b/test/base/test_batch.py
@@ -10,6 +10,7 @@ from tianshou.data import Batch, to_torch
 def test_batch():
     assert list(Batch()) == []
     batch = Batch(a=[torch.ones(3), torch.ones(3)])
+    assert torch.allclose(batch.a, torch.ones(2, 3))
     batch = Batch(obs=[0], np=np.zeros([3, 4]))
     assert batch.obs == batch["obs"]
     batch.obs = [1]
--- a/test/base/test_batch.py
+++ b/test/base/test_batch.py
@@ -8,6 +8,7 @@ from tianshou.data import Batch, to_torch
 
 
 def test_batch():
+    assert list(Batch()) == []
     batch = Batch(obs=[0], np=np.zeros([3, 4]))
     assert batch.obs == batch["obs"]
     batch.obs = [1]
--- a/test/base/test_batch.py
+++ b/test/base/test_batch.py
@@ -9,6 +9,7 @@ from tianshou.data import Batch, to_torch
 
 def test_batch():
     assert list(Batch()) == []
+    batch = Batch(a=[torch.ones(3), torch.ones(3)])
     batch = Batch(obs=[0], np=np.zeros([3, 4]))
     assert batch.obs == batch["obs"]
     batch.obs = [1]
--- a/tianshou/data/batch.py
+++ b/tianshou/data/batch.py
@@ -16,10 +16,10 @@ warnings.filterwarnings(
 
 def _is_batch_set(data: Any) -> bool:
     if isinstance(data, (list, tuple)):
-        if len(data) > 0 and isinstance(data[0], (dict, Batch)):
+        if len(data) > 0 and all(isinstance(e, (dict, Batch)) for e in data):
             return True
-    elif isinstance(data, np.ndarray):
-        if isinstance(data.item(0), (dict, Batch)):
+    elif isinstance(data, np.ndarray) and data.dtype == np.object:
+        if all(isinstance(e, (dict, Batch)) for e in data.tolist()):
             return True
     return False
 
@@ -520,9 +520,9 @@ class Batch:
         values_shared = [
             [e[k] for e in batches] for k in keys_shared]
         for k, v in zip(keys_shared, values_shared):
-            if isinstance(v[0], (dict, Batch)):
+            if all(isinstance(e, (dict, Batch)) for e in v):
                 self.__dict__[k] = Batch.stack(v, axis)
-            elif isinstance(v[0], torch.Tensor):
+            elif all(isinstance(e, torch.Tensor) for e in v):
                 self.__dict__[k] = torch.stack(v, axis)
             else:
                 self.__dict__[k] = np.stack(v, axis)
--- a/tianshou/data/batch.py
+++ b/tianshou/data/batch.py
@@ -43,14 +43,15 @@ def _valid_bounds(length: int, index: Union[
         return start_valid and stop_valid
 
 
-def _create_value(inst: Any, size: int) -> Union['Batch', np.ndarray]:
+def _create_value(inst: Any, size: int) -> Union[
+        'Batch', np.ndarray, torch.Tensor]:
     if isinstance(inst, np.ndarray):
         return np.full((size, *inst.shape),
                        fill_value=None if inst.dtype == np.object else 0,
                        dtype=inst.dtype)
     elif isinstance(inst, torch.Tensor):
         return torch.full((size, *inst.shape),
-                          fill_value=None if inst.dtype == np.object else 0,
+                          fill_value=0,
                           device=inst.device,
                           dtype=inst.dtype)
     elif isinstance(inst, (dict, Batch)):
--- a/tianshou/data/batch.py
+++ b/tianshou/data/batch.py
@@ -267,8 +267,11 @@ class Batch:
         if batch_dict is not None:
             if isinstance(batch_dict, (dict, Batch)):
                 for k, v in batch_dict.items():
-                    if isinstance(v, (list, tuple)):
-                        v_ = np.array(v)
+                    if isinstance(v, (list, tuple, np.ndarray)):
+                        if not isinstance(v, np.ndarray):
+                            v_ = np.array(v)
+                        else:
+                            v_ = v
                         if v_.dtype != np.object:
                             v = v_  # normal data list, this is the main case
                             if not issubclass(v.dtype.type,
--- a/tianshou/data/batch.py
+++ b/tianshou/data/batch.py
@@ -275,7 +275,11 @@ class Batch:
                                               (np.bool_, np.number)):
                                 v = v.astype(np.object)
                         else:
-                            v = Batch(v)  # list of dict / Batch
+                            if _is_batch_set(v):
+                                v = Batch(v)  # list of dict / Batch
+                            else:
+                                # this is actually a data list with objects
+                                v = v_
                         self.__dict__[k] = v
                     elif isinstance(v, dict):
                         self.__dict__[k] = Batch(v)
--- a/tianshou/data/batch.py
+++ b/tianshou/data/batch.py
@@ -271,7 +271,8 @@ class Batch:
                         v_ = np.array(v)
                         if v_.dtype != np.object:
                             v = v_  # normal data list, this is the main case
-                            if not issubclass(v.dtype.type, (np.bool_, np.number)):
+                            if not issubclass(v.dtype.type,
+                                              (np.bool_, np.number)):
                                 v = v.astype(np.object)
                         else:
                             v = Batch(v)  # list of dict / Batch
--- a/tianshou/data/batch.py
+++ b/tianshou/data/batch.py
@@ -264,18 +264,24 @@ class Batch:
                  **kwargs) -> None:
         if copy:
             batch_dict = deepcopy(batch_dict)
-        if _is_batch_set(batch_dict):
-            self.stack_(batch_dict)
-        elif isinstance(batch_dict, (dict, Batch)):
-            for k, v in batch_dict.items():
-                if isinstance(v, dict) or _is_batch_set(v):
-                    self.__dict__[k] = Batch(v)
-                else:
-                    if isinstance(v, list):
-                        v = np.array(v)
-                        if not issubclass(v.dtype.type, (np.bool_, np.number)):
-                            v = v.astype(np.object)
-                    self.__dict__[k] = v
+        if batch_dict is not None:
+            if isinstance(batch_dict, (dict, Batch)):
+                for k, v in batch_dict.items():
+                    if isinstance(v, (list, tuple)):
+                        v_ = np.array(v)
+                        if v_.dtype != np.object:
+                            v = v_  # normal data list, this is the main case
+                            if not issubclass(v.dtype.type, (np.bool_, np.number)):
+                                v = v.astype(np.object)
+                        else:
+                            v = Batch(v)  # list of dict / Batch
+                        self.__dict__[k] = v
+                    elif isinstance(v, dict):
+                        self.__dict__[k] = Batch(v)
+                    else:
+                        self.__dict__[k] = v
+            elif _is_batch_set(batch_dict):
+                self.stack_(batch_dict)
         if len(kwargs) > 0:
             self.__init__(kwargs, copy=copy)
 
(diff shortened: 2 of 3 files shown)


tianshou/data/batch.py
```diff
@@ -16,10 +16,10 @@
 
 def _is_batch_set(data: Any) -> bool:
     if isinstance(data, (list, tuple)):
-        if len(data) > 0 and isinstance(data[0], (dict, Batch)):
+        if len(data) > 0 and all(isinstance(e, (dict, Batch)) for e in data):
```
duburcqa, line 19: It seems heavily time consuming. Please post some profiling to check the performance.

tianshou/data/batch.py
```diff
@@ -16,10 +16,10 @@
...
             return True
-    elif isinstance(data, np.ndarray):
-        if isinstance(data.item(0), (dict, Batch)):
+    elif isinstance(data, np.ndarray) and data.dtype == np.object:
+        if all(isinstance(e, (dict, Batch)) for e in data.tolist()):
```
duburcqa, line 22: Same, `tolist` is doing a copy, which is time consuming.

I think you misunderstand the point of checking types: type checking is used to determine the appropriate behavior, NOT to check that the user is doing something crazy. You should be consider the user as an adversary. Checking everything is taking a lot of time, for little gain.

tianshou/data/batch.py
```diff
@@ -520,9 +521,9 @@ def stack_(self,
         values_shared = [
             [e[k] for e in batches] for k in keys_shared]
         for k, v in zip(keys_shared, values_shared):
-            if isinstance(v[0], (dict, Batch)):
+            if all(isinstance(e, (dict, Batch)) for e in v):
```
duburcqa, line 558: Same.

tianshou/data/batch.py
```diff
@@ -520,9 +521,9 @@ def stack_(self,
...
                 self.__dict__[k] = Batch.stack(v, axis)
-            elif isinstance(v[0], torch.Tensor):
+            elif all(isinstance(e, torch.Tensor) for e in v):
```
duburcqa, line 560: Same.

tianshou/data/utils.py
```diff
@@ -37,7 +37,7 @@ def to_torch(x: Union[torch.Tensor, dict, Batch, np.ndarray],
     elif isinstance(x, (np.number, np.bool_, Number)):
         x = to_torch(np.asanyarray(x), dtype, device)
     elif isinstance(x, list) and len(x) > 0 and \
-            isinstance(x[0], (np.number, np.bool_, Number)):
+            all(isinstance(e, (np.number, np.bool_, Number)) for e in x):
```
duburcqa, line 40: Same.

tianshou/data/batch.py
youkaichao: ``isinstance `` is a quick check. The most time is spent on data copy, environment simulation, and learning.

duburcqa: Not really. If you have a look to the profiling graph, we are already spending about 7% of the whole execution time checking types.

tianshou/data/batch.py
youkaichao: I think you misunderstand the spent time. Deepcopy is time-consuming, but here we assure it is an array with ``np.object`` type, so ``tolist()`` only copies a reference.

There is indeed a trade-off between flexibility and debug-ability. But these operators are cheap in terms of time-cost, and can make sure users will not be confused by strange exceptions.

I'm not considering the user as an adversary, I'm avoiding users from unexpected behaviors with extremely long traceback. When they code something wrong, they should get the hint about what is going wrong.

tianshou/data/batch.py
youkaichao: Well, we have to recognize that current tests are some toy examples. In reality, the reinforcement learning task can run for days. In that practical usage, these type checks are almost free. But without these type check, a program written to run for days can break during training, resulting in a waste of several days.

duburcqa: I don't fully agree. Even though this analysis is based on toy models, I don't think it is that far from the reality. And if it is, we should implement a proper benchmark for profiling Tianshou. Next, it is extremely likely to fail way before several days. Finally, reinforcement learning tasks running for days are not real. It does not exist outside the dream world outside universities and GAFAM alike companies. Usually, companies using reinforcement learning are consuming much less resources in average that you seem imagine. Personally, I'm working in a private company developing a self-balanced bipedal exoskeleton and clearly we are able to achieve remarkable results without this much computational resources.

tianshou/data/batch.py
duburcqa: > I think you misunderstand the spent time. Deepcopy is time-consuming, but here we assure it is an array with np.object type, so tolist() only copies a reference.

True

> I'm not considering the user as an adversary, I'm avoiding users from unexpected behaviors with extremely long traceback. When they code something wrong, they should get the hint about what is going wrong.

It is straightforward from the documentation that it does not make sense to use something else than `np.ndarray` of `Batch`/`dict` without keyword. I don't think you should infantilize people at the point of not reading the first line of the documentation. Even though it is cheap, we are creating `Batch` millions of times during the whole training process, so at the end it matters.

tianshou/data/batch.py
Trinkle23897: I think maybe we can change the code to avoid too many calls on _is_batch_set.

tianshou/data/batch.py
Trinkle23897: I profile the code and find that this pr will not affect `stack_` and `to_torch`'s behavior in the test script.

duburcqa:
@Trinkle23897 Do as you think it is the best approach. I gave a warning regarding compromise between performance and debug-ability and It seems you understood my point, so I will line up behind you.

tianshou/data/utils.py
Trinkle23897: change to `isinstance(x[0], (np.number, np.bool_, Number)) and all(isinstance(e, (np.number, np.bool_, Number)) for e in x)` because `all` will calculate all items in the list even if the first one cannot pass the check

youkaichao: According to [here](https://stackoverflow.com/questions/2580136/does-python-support-short-circuiting), python ``all`` supports short circuiting.

tianshou/data/batch.py
```diff
@@ -263,18 +264,25 @@ def __init__(self,
...
-                    if isinstance(v, list):
-                        v = np.array(v)
-                        if not issubclass(v.dtype.type, (np.bool_, np.number)):
-                            v = v.astype(np.object)
-                    self.__dict__[k] = v
+        if batch_dict is not None:
+            if isinstance(batch_dict, (dict, Batch)):
+                for k, v in batch_dict.items():
+                    if isinstance(v, (list, tuple)):
+                        v_ = np.array(v)
```
duburcqa: Why not to replace line 273 by this one ? So that introducing `v_` is unnecessary

youkaichao: Hi, I run a test to check the call of ``_is_batch_set``, the code is [here](https://github.com/youkaichao/tianshou/commit/650b1e790a6185e471778444fd93932927ebb5e8). The results indicate that most calls of ``_is_batch_set`` returns ``False``.

youkaichao: I run the code ``test/discrete/test_pg.py`` and record the call of ``_is_batch_set`` in three places where I mark them as init1, init2, and setattr.

Originally:
init1 called 287733 times, all return False.
init2 called 405493 times, with 397151 False.

After adding the filter ``if batch_dict is not None``, the calls changed into:
init1 called 201907 times, all return False.
init2 called 405493 times, with 397151 False.

After switching init1 and init2, the calls changed into:
init2 called 414729 times, with 397151 False.

Then we realize the actually cost is the call of _is_batch_set in init2. The major case is a list array. By chaning the code like this, no ``_is_batch_set`` is called.

duburcqa: Indeed, it is a bit twisted by why not !

tianshou/data/batch.py
```diff
@@ -263,18 +264,37 @@ def __init__(self,
...
-                            v = v.astype(np.object)
-                    self.__dict__[k] = v
+        if batch_dict is not None:
+            if isinstance(batch_dict, (dict, Batch)):
+                for k, v in batch_dict.items():
+                    if isinstance(v, (list, tuple, np.ndarray)):
+                        v_ = None
+                        if isinstance(v, np.ndarray):
+                            v_ = v
+                        elif all(isinstance(e, torch.Tensor) for e in v):
```
youkaichao: Previous code (like db0e2e5cd21071fa6056a9f3e3c84b540bf80bf6 ) fails on ``Batch(a=[torch.ones(3, 3), torch.ones(3, 3)])``. This commit fixes it.

test/base/test_batch.py
```diff
@@ -8,6 +8,7 @@
 
 
 def test_batch():
+    assert list(Batch()) == []
```
Trinkle23897, line 11: @duburcqa in #119 you can change it as you want. I also agree with you that `__iter__` is not a good solution, but I approved it.

tianshou/data/batch.py
```diff
@@ -263,18 +264,37 @@ def __init__(self,
...
+                            else:
+                                # this is actually a data list with objects
+                                v = v_
+                        self.__dict__[k] = v
+                    elif isinstance(v, dict):
+                        self.__dict__[k] = Batch(v)
+                    else:
+                        self.__dict__[k] = v
+            elif _is_batch_set(batch_dict):
+                self.stack_(batch_dict)
```
Trinkle23897, line 296: This time in test_dqn.py, the calling counts of _is_batch_set reduce from 150k to 3.5k.

Trinkle23897 (approved):
So at this time, it contains a test-case of `list(Batch())` (which I found it missing in the last PR, my fault), optimization of _is_batch_set, and compatible of list of torch.Tensor.
What's your opinion @duburcqa ?

tianshou/data/batch.py
```diff
@@ -263,18 +264,37 @@ def __init__(self,
...
-                        v = np.array(v)
-                        if not issubclass(v.dtype.type, (np.bool_, np.number)):
-                            v = v.astype(np.object)
-                    self.__dict__[k] = v
+        if batch_dict is not None:
+            if isinstance(batch_dict, (dict, Batch)):
+                for k, v in batch_dict.items():
+                    if isinstance(v, (list, tuple, np.ndarray)):
+                        v_ = None
+                        if isinstance(v, np.ndarray):
```
duburcqa: No need of this `if isinstance(v, np.ndarray)` case since you can replace line 279 by `np.asanyarray` because it is free of cost when the argument is already an array.

tianshou/data/batch.py
duburcqa: Indeed, nice catch. You should add this check to the unit tests.

tianshou/data/batch.py
duburcqa: Yes, nice job! (if it improving the overall performance, but I guess so from your comment)

tianshou/data/batch.py
duburcqa: I guess we solved this issue by reducing the number of calls to this method.

test/base/test_batch.py
duburcqa: Will do! :smile:

duburcqa:
> What's your opinion @duburcqa ?

Just my two remaining comments to address and we are good :+1:

tianshou/data/batch.py
youkaichao: Good. I have changed the code.
```
a = np.asarray([1, 2, 3])
b = np.asanyarray(a)
b is a # True
c = np.array(a)
c is a # False
```

youkaichao:
In summary, this PR does the following:

improvement: dramatic reduce of the call to _is_batch_set
bugfix: ``list(Batch())`` fail; ``Batch(a=[torch.ones(3), torch.ones(3)])`` fail;
misc: add type check for each element rather than the first element; add test case; ``_create_value`` with ``torch.Tensor`` does not have ``np.object`` type;

test/base/test_batch.py
```diff
@@ -8,6 +8,8 @@
...
+    batch = Batch(a=[torch.ones(3), torch.ones(3)])
```
Trinkle23897, line 12: add an assert after this line: what's this batch like?

Trinkle23897 (changes requested):

tianshou/data/batch.py
```diff
@@ -269,14 +269,13 @@ def __init__(self,
                 for k, v in batch_dict.items():
                     if isinstance(v, (list, tuple, np.ndarray)):
                         v_ = None
-                        if isinstance(v, np.ndarray):
-                            v_ = v
-                        elif all(isinstance(e, torch.Tensor) for e in v):
+                        if not isinstance(v, np.ndarray) and \
```
duburcqa, line 272: `if not isinstance(v, np.ndarray)` part is unnecessary no ? Since `np.ndarray` of `torch.Tensor` is impossible.

tianshou/data/batch.py
duburcqa: Yes, that's the behavior I was referring to!

Trinkle23897 (approved):

tianshou/data/batch.py
youkaichao: In this branch, ``v`` can be ``list, tuple, np.ndarray``.  ``if not isinstance(v, np.ndarray)`` means ``isinstance(v, (list, tuple))``, then I can iterate them at a low cost.

duburcqa: I don't get it. `all(isinstance(e, torch.Tensor) for e in np.array([1.0, 2.0]))` is fine. It has a higher cost than for list or tuple even using short circuiting ?

tianshou/data/batch.py
Trinkle23897: By the way, I found that the unittest total time goes back to less than 6 mins in average.

tianshou/data/batch.py
youkaichao: ``for e in np.array([1.0, 2.0])`` should have a higher cost, because numpy has to construct each ``e``.

duburcqa:
```
In [16]: a = np.zeros((10000000, 10))                                                                                                                                                                       

In [17]: %timeit all(isinstance(e, torch.Tensor) for e in a)                                                                                                                                                
481 ns ± 1.74 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [18]: a2 = a.tolist()                                                                                                                                                                                    

In [19]: %timeit all(isinstance(e, torch.Tensor) for e in a2)                                                                                                                                               
386 ns ± 3.09 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [20]: %timeit isinstance(a, np.ndarray)                                                                                                                                                                  
89.1 ns ± 0.531 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

Are you talking about this ? Indeed it is "faster" to check for `np.ndarray` but I don't know if it is significant in practice.

duburcqa (approved):

tianshou/data/batch.py
youkaichao: Yes, this is what I mean. You are such a quick coder :)

duburcqa:
Ready to be merged.

PR #114, opened 2020-07-07
bugfix for empty_

youkaichao:
    + [ ] exception-raising fix
    + [ ] new feature

  ```python
  import tianshou, torch, sys
  ```

### DIFF
--- a/tianshou/data/batch.py
+++ b/tianshou/data/batch.py
@@ -569,10 +569,12 @@ class Batch:
             else:  # scalar value
                 warnings.warn('You are calling Batch.empty on a NumPy scalar, '
                               'which may cause undefined behaviors.')
-                if isinstance(v, (np.generic, Number)):
+                if isinstance(v, (np.number, Number)):
                     self.__dict__[k] *= 0
                     if np.isnan(self.__dict__[k]):
                         self.__dict__[k] = 0
+                elif isinstance(v, np.bool_):
+                    self.__dict__[k] &= False
                 else:
                     self.__dict__[k] = None
         return self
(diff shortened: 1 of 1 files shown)


youkaichao:
``np.isnan`` does not support string argument, so the if-branch only works for numbers.
For bool type, ``*= 0`` would change the data type, and ``&=False`` suits better.

tianshou/data/batch.py
```diff
@@ -569,10 +569,12 @@ def empty_(self, index: Union[
             else:  # scalar value
                 warnings.warn('You are calling Batch.empty on a NumPy scalar, '
                               'which may cause undefined behaviors.')
-                if isinstance(v, (np.generic, Number)):
+                if isinstance(v, (np.number, Number)):
```
duburcqa: Why not to add `np.bool_` here. `self.__dict__[k] *= 0` is doing the same as `self.__dict__[k] &= False`, so I don't see the point in creating a new if case.

duburcqa: OK I just saw your comment. There is the same problem with native Python `bool` type. So maybe you were right in the firstt place, and using `self.__dict__[k] = v.__class__(0)` is the most robust approach after all.

youkaichao: Agree. Although ``self.__dict__[k] = v.__class__(0)`` will lead to new memory allocation, the memory cost of scalar is negligible.

duburcqa (approved):

tianshou/data/batch.py
duburcqa: Indeed.

Trinkle23897 (approved):

PR #115, opened 2020-07-07
fix Batch init for types other than number and bool

youkaichao:
    + [ ] exception-raising fix
    + [ ] new feature

  ```python
  import tianshou, torch, sys
  ```

### DIFF
--- a/tianshou/data/batch.py
+++ b/tianshou/data/batch.py
@@ -114,7 +114,7 @@ class Batch:
         >>> print(data[0])
         Batch(
             a: Batch(
-                b: array(['0.0', 'info'], dtype='<U32'),
+                b: array([0.0, 'info'], dtype=object),
             ),
         )
 
@@ -224,7 +224,7 @@ class Batch:
         Batch(
             a: array([False,  True]),
             b: Batch(
-                   c: array([0., 3.]),
+                   c: array([None, 'st']),
                    d: array([0., 0.]),
                ),
         )
@@ -270,6 +270,9 @@ class Batch:
                 else:
                     if isinstance(v, list):
                         v = np.array(v)
+                        if not np.issubdtype(v.dtype, np.bool_) and \
+                                not np.issubdtype(v.dtype, np.number):
+                            v = v.astype(np.object)
                     self.__dict__[k] = v
         if len(kwargs) > 0:
             self.__init__(kwargs, copy=copy)
@@ -525,7 +528,11 @@ class Batch:
             elif isinstance(v[0], torch.Tensor):
                 self.__dict__[k] = torch.stack(v, axis)
             else:
-                self.__dict__[k] = np.stack(v, axis)
+                v = np.stack(v, axis)
+                if not np.issubdtype(v.dtype, np.bool_) and \
+                        not np.issubdtype(v.dtype, np.number):
+                    v = v.astype(np.object)
+                self.__dict__[k] = v
         keys_partial = reduce(set.symmetric_difference, keys_map)
         for k in keys_partial:
             for i, e in enumerate(batches):
--- a/tianshou/data/batch.py
+++ b/tianshou/data/batch.py
@@ -87,9 +87,9 @@ class Batch:
 
     In short, you can define a :class:`Batch` with any key-value pair.
 
-    For Numpy arrays, only data types with ``np.object`` and numbers are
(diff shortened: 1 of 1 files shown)


youkaichao:
Previously, for data types other than bool and number (notoriously, strings) , ``Batch`` leaves them as they are. Now that we only support number, bool, and object. Object can actually hold any types. Storing other data types such as strings in object type simplifies the code and brings no harm.

youkaichao:
With this pr, ``np.ndarray`` in Batch only has three possible generic types: ``np.bool``, ``np.number`` and ``np.object``.

When values are passed into Batch constructor, if the values have types rather than ``np.bool`` and ``np.number``, they will be held in ``np.object`` type.

tianshou/data/batch.py
```diff
@@ -45,9 +45,13 @@ def _valid_bounds(length: int, index: Union[
 
 def _create_value(inst: Any, size: int) -> Union['Batch', np.ndarray]:
     if isinstance(inst, np.ndarray):
+        if issubclass(inst.dtype.type, (np.bool_, np.number)):
```
duburcqa, line 48: Are you sure `isinstance(inst.dtype, (np.bool_, np.number))` is not doing the same ? I'm interested to see a case were it is not, since `isinstance` is also checking inheritance.

tianshou/data/batch.py
```diff
@@ -337,25 +348,24 @@ def __setitem__(
                 if isinstance(val, Batch):
                     self.__dict__[key][index] = Batch()
                 elif isinstance(val, np.ndarray) and \
-                        val.dtype == np.integer:
-                    # Fallback for np.array of integer,
-                    # since neither None or nan is supported.
+                        issubclass(val.dtype.type, (np.bool_, np.number)):
```
duburcqa, line 349: same.

tianshou/data/batch.py
```diff
@@ -321,6 +329,9 @@ def __setitem__(
             index: Union[str, slice, int, np.integer, np.ndarray, List[int]],
             value: Any) -> None:
         """Assign value to self[index]."""
+        if isinstance(value, np.ndarray):
+            if not issubclass(value.dtype.type, (np.bool_, np.number)):
```
duburcqa, line 331: same.

tianshou/data/batch.py
```diff
@@ -337,25 +348,24 @@ def __setitem__(
...
                     self.__dict__[key][index] = None
 
-    def __iadd__(self, other: Union['Batch', Number]):
+    def __iadd__(self, other: Union['Batch', Number, np.number]):
         """Algebraic addition with another :class:`~tianshou.data.Batch`
         instance in-place."""
         if isinstance(other, Batch):
             for (k, r), v in zip(self.__dict__.items(),
                                  other.__dict__.values()):
+                # TODO are keys consistent?
```
duburcqa, line 360: No time to check anyway, and it is likely to fail somewhere else otherwise.

duburcqa:
Do not forget to do some profiling when modifying any data class, since their efficiency is critical for the whole training process.

tianshou/data/batch.py
youkaichao:
```
a = np.asarray([False, False])
isinstance(a.dtype, np.bool_) # False
```

``inst.dtype`` is actually of type ``np.dtype``.

duburcqa: Ah yes I see, so it is the same as `isinstance(inst.item(0), (np.bool_, np.number))` but `inst.item(0)` is doing an extra copy so I guess it is slower.

maybe you could have a look to `np.can_cast` instead of `isinstance`:
```
np.can_cast(np.asarray([1.0, 1.0]), np.inexact)
>>> True
np.can_cast(np.asarray([1.0, 1.0]), np.bool_)
>>> False
np.can_cast(np.asarray([False, False]), np.bool_)
>>> True
```

youkaichao: ``np. can_cast `` and ``np.issubdtype`` both works, but they cannot test multiple types in one function call. I choose ``issubclass`` because it supports multiple tests at the same time by calling ``issubclass(a, (b, c, d))``.

duburcqa: Indeed.

duburcqa (approved):

Trinkle23897 (approved):

PR #119, opened 2020-07-08
Raise exception for Batch __getitem__.

### LINKED ISSUE #118: Batch code regression
Why bound checking of `getitem` method have been removed ?
https://github.com/thu-ml/tianshou/blob/e0f4862d01e39e3f1fc6d3c90e7c705b2eef9b7e/tianshou/data/batch.py#L271

It leads to several issues, including the need of custom `__iter__`, which shouldn't be necessary:
https://github.com/thu-ml/tianshou/blob/481015932c78ed1eee991d1d7b1ac57545a11f83/tianshou/data/batch.py#L436

It was not the original plan https://github.com/thu-ml/tianshou/issues/104#issuecomment-651033749 and I'm not OK with changes made in commit f58358e2e9654c4c8ca3542e240b7b6545cd239a .

Performance issue was discussed [here](https://github.com/thu-ml/tianshou/pull/106#issuecomment-651589598), and it appears not to be an issue [here](https://github.com/thu-ml/tianshou/pull/106#issuecomment-651596000) as far as I remember. 

By standard, `__getitem__` is supposed to fail when invalid item is requested, and it is no longer the case.

So, unless it is turns out to have a significant impact on the performances, I suggest rolling back whose changes.

Plus, 10 minutes to review a PR https://github.com/thu-ml/tianshou/pull/117 feels a bit demanding...

duburcqa:
Fix https://github.com/thu-ml/tianshou/issues/118

### DIFF
--- a/tianshou/data/batch.py
+++ b/tianshou/data/batch.py
@@ -24,25 +24,6 @@ def _is_batch_set(data: Any) -> bool:
     return False
 
 
-def _valid_bounds(length: int, index: Union[
-        slice, int, np.integer, np.ndarray, List[int]]) -> bool:
-    if isinstance(index, (int, np.integer)):
-        return -length <= index and index < length
-    elif isinstance(index, (list, np.ndarray)):
-        return _valid_bounds(length, np.min(index)) and \
-            _valid_bounds(length, np.max(index))
-    elif isinstance(index, slice):
-        if index.start is not None:
-            start_valid = _valid_bounds(length, index.start)
-        else:
-            start_valid = True
-        if index.stop is not None:
-            stop_valid = _valid_bounds(length, index.stop - 1)
-        else:
-            stop_valid = True
-        return start_valid and stop_valid
-
-
 def _create_value(inst: Any, size: int) -> Union[
         'Batch', np.ndarray, torch.Tensor]:
     if isinstance(inst, np.ndarray):
@@ -333,13 +314,17 @@ class Batch:
         """Return self[index]."""
         if isinstance(index, str):
             return self.__dict__[index]
-        b = Batch()
-        for k, v in self.items():
-            if isinstance(v, Batch) and len(v.__dict__) == 0:
-                b.__dict__[k] = Batch()
-            else:
-                b.__dict__[k] = v[index]
-        return b
+        batch_items = self.items()
+        if len(batch_items) > 0:
+            b = Batch()
+            for k, v in batch_items:
+                if isinstance(v, Batch) and len(v.__dict__) == 0:
+                    b.__dict__[k] = Batch()
+                else:
+                    b.__dict__[k] = v[index]
+            return b
+        else:
+            raise IndexError("Cannot access item from empty Batch object.")
 
     def __setitem__(
             self,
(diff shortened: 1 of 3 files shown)


duburcqa:
Ok I see the issue. We are relying on allowing accessing item of empty batches. It does not look very robust. May be we should find a way to avoid such trick rather than overloading `__iter__`. For example, replacing `Batch()` by a list of batches with appropriate length for `Policy` instead of being hacky.

Trinkle23897:
> Ok I see the issue. We are relying on allowing accessing item of empty batches. It does not look very robust. Maybe we should find a way to avoid such trick rather than overloading `__iter__`.

Indeed. I think there are two possible solutions:

1. avoid empty Batch() from buffer side (e.g., creating info/policy). But this may cause some error such as this case:
```
timestep 0 1 2 3 4
info     / / / / {something}
```
before timestep 4, batch.info is `Batch()`, but should change to `None` to avoid empty Batch().

It also needs to overwrite the getattr: when getting a non-exist key, it needs to return `None` rather than raising an AttributeError.

2. keep empty Batch() indexing (keep the original code). I still prefer this one.

duburcqa:
I think the first is better. And it is pretty easy to do in 1 or 2 lines by modifying `Batch.empty` or/and `Buffer._add_to_buffer`. It would not harm the performance and it is consistent with standard. It will ease the compatibility with other libraries that are assuming standard behavior.

duburcqa:
> It also needs to overwrite the getattr: when getting a non-exist key, it needs to return None rather than raising an AttributeError.

Why? This is not consistent with  Python`dict`.

Trinkle23897:
> > It also needs to overwrite the getattr: when getting a non-exist key, it needs to return None rather than raising an AttributeError.
> 
> Why? This is not consistent with Python `dict`.

Because the downstream code (e.g., policy.learn()) may directly call batch.info. I'm not sure whether it will go wrong in the current test script. We can modify it together.

I think raise such an exception is okay, since my proposal will treat `None` as a dict, it still goes wrong.

duburcqa:
It should be working now. I think the solution is simple and elegant. No need to do any complex refactoring.

duburcqa:
Added unit test to make sure it is part of the specification of Batch.

Trinkle23897:
I'm waiting for #112 to merge into master.

duburcqa:
> I'm waiting for #112 to merge into master.

Yes of course !

tianshou/data/batch.py
```diff
@@ -333,13 +314,17 @@ def __getitem__(self, index: Union[
...
+        if len(batch_items) > 0:
+            b = Batch()
+            for k, v in batch_items:
+                if isinstance(v, Batch) and len(v.__dict__) == 0:
+                    b.__dict__[k] = Batch()
+                else:
+                    b.__dict__[k] = v[index]
+            return b
+        else:
+            raise IndexError("Cannot access item from empty Batch object.")
```
youkaichao, line 327: They may lead to a strange behavior:
```
d=Batch(a=[1,2,3], b=dict(c=[3,4,5]), c={}) 
d[2] # no problem
d.c[2] # exception
```
I'm not sure this is ok.

duburcqa: In my opinion it is the desired behavior since it means different things. Doing:
```
d=Batch(a=[1,2,3], b=dict(c=[3,4,5]), c={}) 
```
you are just "reserving" the key `c`. But as soon as you do `d.c`, you get an empty `Batch` and that's all. And as such, accessing an item from an empty Batch raises an exception. In those two cases, `Batch()` as a different meaning and behave differently.

I can understand that it is a bit odd, but `Batch()[2] = Batch()` is even stranger to me, and does not follow the standard.

Trinkle23897 (approved):

PR #120, opened 2020-07-08
bugfix for update with empty buffer; remove duplicate variable _weight_sum in PrioritizedReplayBuffer

youkaichao:
    + [ ] exception-raising fix
    + [ ] new feature

  ```python
  import tianshou, torch, sys
  ```


tianshou/data/buffer.py
```diff
@@ -151,6 +151,8 @@ def _add_to_buffer(self, name: str, inst: Any) -> None:
 
     def update(self, buffer: 'ReplayBuffer') -> None:
         """Move the data from the given buffer to self."""
+        if len(buffer) == 0:
+            return
```
youkaichao, line 155: bugfix for update with empty buffer

youkaichao:
Currently, in the ``PrioritizedReplayBuffer.sample()``, there is ``self.weight / self.weight.sum()``, which means each time the code will calculate the sum of weight, so we do not need a ``self._weight_sum`` variable any more.

In addition, since it is a ``PrioritizedReplayBuffer``, it makes no sense to pass ``importance_sample`` argument to ``sample()`` (``importance_sample`` should always be ``True``), which also makes the interface ``sample()`` consistent with other replaybuffers.

tianshou/data/buffer.py
duburcqa: Why not raise an exception ? Is doing this allowed ?

youkaichao: Update a buffer with another empty buffer is allowed. Just like extending a list with another list ``a += b``. We just need to assure it behaves correctly.

duburcqa (approved):

Trinkle23897:
I’m wondering which one is better in prioritized buffer: segment tree or np.sum. Maybe I’ll do this tomorrow.

PR #125, opened 2020-07-10
Improve collector

### LINKED ISSUE #105: Refactoring of Buffer & Collector

### LINKED ISSUE #127: Inconsistent type annotation
    + [ ] exception-raising bug
    + [ ] RL algorithm bug
    + [ ] new feature request
  ```python
  import tianshou, torch, sys
  ```

Currently, there are some inconsistent type annotations. For example ,``reward`` can be ``float`` or ``np.ndarray``. This issue tries to track the type of some concepts, and one day we can fix them once in all.

youkaichao:
This pull request keeps improving the ``collector``.

1. make fileds with empty Batch rather than None after reset
2. add reward_metric argument to collector to deal with marl
3. rewrite collector code, with the internal data `self.data` be a `Batch`

### DIFF
--- a/test/base/test_collector.py
+++ b/test/base/test_collector.py
@@ -27,7 +27,7 @@ class MyPolicy(BasePolicy):
 
 def preprocess_fn(**kwargs):
     # modify info before adding into the buffer
-    # if info is not provided from env, it will be a `Batch()`.
+    # if info is not provided from env, it will be a ``Batch()``.
     if not kwargs.get('info', Batch()).is_empty():
         n = len(kwargs['obs'])
         info = kwargs['info']
--- a/test/base/test_collector.py
+++ b/test/base/test_collector.py
@@ -126,15 +126,19 @@ def test_collector_with_ma():
     policy = MyPolicy()
     c0 = Collector(policy, env, ReplayBuffer(size=100),
                    preprocess_fn, reward_metric=reward_metric)
-    c0.collect(n_step=3)
-    c0.collect(n_episode=3)
+    r = c0.collect(n_step=3)['rew']
+    assert np.asanyarray(r).size == 1 and r == 0.
+    r = c0.collect(n_episode=3)['rew']
+    assert np.asanyarray(r).size == 1 and r == 4.
     env_fns = [lambda x=i: MyTestEnv(size=x, sleep=0, ma_rew=4)
                for i in [2, 3, 4, 5]]
     envs = VectorEnv(env_fns)
     c1 = Collector(policy, envs, ReplayBuffer(size=100),
                    preprocess_fn, reward_metric=reward_metric)
-    c1.collect(n_step=10)
-    c1.collect(n_episode=[2, 1, 1, 2])
+    r = c1.collect(n_step=10)['rew']
+    assert np.asanyarray(r).size == 1 and r == 4.
+    r = c1.collect(n_episode=[2, 1, 1, 2])['rew']
+    assert np.asanyarray(r).size == 1 and r == 4.
     batch = c1.sample(10)
     print(batch)
     c0.buffer.update(c1.buffer)
@@ -150,7 +154,8 @@ def test_collector_with_ma():
                        [[x] * 4 for x in rew])
     c2 = Collector(policy, envs, ReplayBuffer(size=100, stack_num=4),
                    preprocess_fn, reward_metric=reward_metric)
-    c2.collect(n_episode=[0, 0, 0, 10])
+    r = c2.collect(n_episode=[0, 0, 0, 10])['rew']
+    assert np.asanyarray(r).size == 1 and r == 4.
     batch = c2.sample(10)
     print(batch['obs_next'])
 
--- a/test/base/test_collector.py
+++ b/test/base/test_collector.py
@@ -27,16 +27,16 @@ class MyPolicy(BasePolicy):
 
 def preprocess_fn(**kwargs):
     # modify info before adding into the buffer
-    if kwargs.get('info', None) is not None:
+    # if info is not provided from env, it will be a `Batch()`.
+    if not kwargs.get('info', Batch()).is_empty():
         n = len(kwargs['obs'])
         info = kwargs['info']
         for i in range(n):
             info[i].update(rew=kwargs['rew'][i])
         return {'info': info}
-        # or
-        # return Batch(info=info)
+        # or: return Batch(info=info)
     else:
-        return {}
+        return Batch()
 
 
 class Logger(object):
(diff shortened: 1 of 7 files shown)


duburcqa (dismissed):

Trinkle23897:
> 3. add get_final_reward_fn argument to collector to deal with marl

~~I don't agree with this point. I think the collector should return a np.ndarray of original reward data instead of processed data in this case.~~

youkaichao:
> > 1. add get_final_reward_fn argument to collector to deal with marl
> 
> I don't agree with this point. I think the collector should return a np.ndarray of original reward data instead of processed data in this case.

Disagree with your disagree. Even though the variable name is ``rew``,  it actually serves as a scalar metric to indicate the training process, just like accuray or loss. It is natural for people to provide a function to return a scalar metric here if ``rew`` is a vector.

duburcqa:
> It is natural for people to provide a function to return a scalar metric here if rew is a vector.

I agree that `rew` should always be a scalar, still I don't know if dealing with a vector must be done using an handle, for just doing something like simple averaging, and if the user wants to do something fancy, it is still possible for him to wrap the environment by himself to perform any other aggregation formula.

tianshou/data/collector.py
```diff
@@ -87,68 +92,56 @@ class Collector(object):
...
-            else:
-                raise TypeError('The buffer in data collector is invalid!')
+            self._cached_buf = [ListReplayBuffer()
+                                for _ in range(self.env_num)]
         self.stat_size = stat_size
         self._action_noise = action_noise
+
+        def _rew_metric(x):
+            flag = np.isscalar(x) or \
+                isinstance(x, np.ndarray) and x.shape == (1, )
```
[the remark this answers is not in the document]
duburcqa: `np.asanyarray(x).size == 1`

test/base/test_collector.py
```diff
@@ -27,16 +27,16 @@ def learn(self):
...
+    # if info is not provided from env, it will be a `Batch()`.
+    if not kwargs.get('info', Batch()).is_empty():
         n = len(kwargs['obs'])
         info = kwargs['info']
         for i in range(n):
             info[i].update(rew=kwargs['rew'][i])
         return {'info': info}
-        # or
-        # return Batch(info=info)
+        # or: return Batch(info=info)
```
youkaichao, line 37: So why not ``Batch(info=info)``?

Trinkle23897: Both are okay.

tianshou/data/collector.py
```diff
@@ -31,6 +31,11 @@ class Collector(object):
     :param BaseNoise action_noise: add a noise to continuous action. Normally
         a policy already has a noise param for exploration in training phase,
         so this is recommended to use in test collector for some purpose.
+    :param function reward_metric: to be used in multi-agent RL. The reward to
+        report is of shape [reward_length], but we need to return a single
+        scalar to monitor training. This function specifies what is the desired
+        metric, i.e. the reward of agent 1 or the average reward over all
+        agents. By default, the behavior is to select the reward of agent 1.
```
youkaichao: ``i.e.`` -> ``e.g.``

tianshou/data/collector.py
```diff
@@ -117,9 +117,7 @@ def __init__(self,
         self._action_noise = action_noise
 
         def _rew_metric(x):
-            flag = np.isscalar(x) or \
-                isinstance(x, np.ndarray) and x.shape == (1, )
-            assert flag, 'Please specify the reward_metric ' \
+            assert np.asanyarray(x) == 1, 'Please specify the reward_metric ' \
```
youkaichao: This is a value compare, not a shape compare.

tianshou/data/collector.py
```diff
@@ -87,68 +92,54 @@ class Collector(object):
...
-                    ListReplayBuffer() for _ in range(self.env_num)]
-            else:
-                raise TypeError('The buffer in data collector is invalid!')
+            self._cached_buf = [ListReplayBuffer()
+                                for _ in range(self.env_num)]
         self.stat_size = stat_size
         self._action_noise = action_noise
+
+        def _rew_metric(x):
+            assert np.asanyarray(x) == 1, 'Please specify the reward_metric ' \
```
duburcqa: `np.asanyarray(x).size`

Trinkle23897:
Should be okay now. Please have a review.

tianshou/data/collector.py
```diff
@@ -87,68 +93,57 @@ class Collector(object):
...
-                raise TypeError('The buffer in data collector is invalid!')
+            self._cached_buf = [ListReplayBuffer()
+                                for _ in range(self.env_num)]
         self.stat_size = stat_size
         self._action_noise = action_noise
+
+        def _rew_metric(x):
+            # this internal function is designed for single-agent RL
+            # for multi-agent RL, a reward_metric must be provided
+            assert np.asanyarray(x).size == 1, \
```
duburcqa: Could a simple averaging be the default or it does not make sense ?

tianshou/data/collector.py
```diff
@@ -87,68 +93,57 @@ class Collector(object):
...
-                self._cached_buf = [
-                    ListReplayBuffer() for _ in range(self.env_num)]
-            else:
-                raise TypeError('The buffer in data collector is invalid!')
+            self._cached_buf = [ListReplayBuffer()
+                                for _ in range(self.env_num)]
         self.stat_size = stat_size
         self._action_noise = action_noise
+
+        def _rew_metric(x):
```
duburcqa: Perhaps you could provide several "standard" reward metrics to avoid having to define it again and again for the user ? Plus I'm not a huge fun of nested method declaration when the user is supposed to have access to the method. It is impossible for the user to guess what the method is doing from its name doing `print(self._rew_metric)` since `_rew_metric` is impossible to import.

tianshou/data/collector.py
youkaichao: The default is hard to define.

duburcqa:
Good job! To me, Collector module is now easier to understand and maintain. I like the better integration with Batch, and the addition of a new unit test.

tianshou/data/collector.py
youkaichao: Okay, I can make it a static method.

tianshou/data/collector.py
duburcqa: OK no problem.

duburcqa (approved):

tianshou/data/collector.py
duburcqa: Yes that's good enough for now.

PR #126, opened 2020-07-10
Improve Batch

youkaichao:
    + [ ] exception-raising fix
    + [ ] new feature

  ```python
  import tianshou, torch, sys
  ```

### DIFF
--- a/tianshou/data/batch.py
+++ b/tianshou/data/batch.py
@@ -624,6 +624,9 @@ class Batch:
             raise TypeError("Object of type 'Batch' has no len()")
         return min(r)
 
+    def is_empty(self):
+        return len(self.__dict__.keys()) == 0
+
     @property
     def shape(self) -> List[int]:
         """Return self.shape."""
--- a/tianshou/data/batch.py
+++ b/tianshou/data/batch.py
@@ -518,12 +518,14 @@ class Batch:
                 raise TypeError(s)
 
     @staticmethod
-    def cat(batches: List['Batch']) -> 'Batch':
-        """Concatenate a :class:`~tianshou.data.Batch` object into a single
+    def cat(batches: List[Union[dict, 'Batch']]) -> 'Batch':
+        """Concatenate a list of :class:`~tianshou.data.Batch` object into a single
         new batch.
         """
         batch = Batch()
         for batch_ in batches:
+            if isinstance(batch_, dict):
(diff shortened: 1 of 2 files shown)


youkaichao:
1. make sure the key type of Batch is string, and add unit tests
2. add ``is_empty()`` function and unit tests
3. enable ``cat`` of either dict or Batch, just like ``stack``

tianshou/data/batch.py
```diff
@@ -554,7 +565,7 @@ def stack_(self,
                         self.__dict__[k][i] = val
 
     @staticmethod
-    def stack(batches: List['Batch'], axis: int = 0) -> 'Batch':
+    def stack(batches: List[Union[dict, 'Batch']], axis: int = 0) -> 'Batch':
```
duburcqa, line 568: Why to support `dict` ? Just in case ? Not supporting it would avoid having to check key type. But as you wish.

tianshou/data/batch.py
```diff
@@ -511,12 +518,14 @@ def cat_(self, batch: 'Batch') -> None:
                 raise TypeError(s)
 
     @staticmethod
-    def cat(batches: List['Batch']) -> 'Batch':
-        """Concatenate a :class:`~tianshou.data.Batch` object into a single
+    def cat(batches: List[Union[dict, 'Batch']]) -> 'Batch':
```
duburcqa, line 521: Same, is there a usecase for adding `dict` as supported input type ?

duburcqa:
> 3. enable cat of mixing dict and Batch, just like stack

Why to add this feature ? I didn't find any usecase personally. But I have no problem with it.

tianshou/data/batch.py
youkaichao: Because ``stack_`` actually supports ``dict`` and signature of ``stack`` ``stack_`` should be consistent.

tianshou/data/batch.py
youkaichao: To keep consistent with ``stack``.

youkaichao:
> > 1. enable cat of mixing dict and Batch, just like stack
> 
> Why to add this feature ? I didn't find any usecase personally. But I have no problem with it.

Well, the point is not to enable mixing, but to support both ``dict`` and ``Batch``.

PR #128, opened 2020-07-11
Improve Batch

youkaichao:
Keep improving Batch.

1. code beautify 
2. remove deprecated api ``append``
3. improve and implement Batch.cat_
4. Batch.update
5. fix some bugs (to be filled)
6. ``Batch.is_empty`` now correctly recognizes empty over empty Batch

### DIFF
--- a/tianshou/data/batch.py
+++ b/tianshou/data/batch.py
@@ -609,8 +609,8 @@ class Batch:
 
     @staticmethod
     def stack(batches: List[Union[dict, 'Batch']], axis: int = 0) -> 'Batch':
-        """Stack a list of :class:`~tianshou.data.Batch` object
-        into a single new batch.
+        """Stack a list of :class:`~tianshou.data.Batch` object into a single
+        new batch.
         """
         batch = Batch()
         batch.stack_(batches, axis)
--- a/tianshou/data/batch.py
+++ b/tianshou/data/batch.py
@@ -259,8 +259,7 @@ class Batch:
                         v_ = None
                         if not isinstance(v, np.ndarray) and \
                                 all(isinstance(e, torch.Tensor) for e in v):
-                            v_ = torch.stack(v)
-                            self.__dict__[k] = v_
+                            self.__dict__[k] = torch.stack(v)
                             continue
                         else:
                             v_ = np.asanyarray(v)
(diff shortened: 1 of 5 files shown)


youkaichao:
Following the standard behavior of ``np.concatenate`` and ``torch.cat``, ``Batch.cat_`` should accept a list of Batch.

Prior to this commit:
1. ``Batch.cat`` implements a one-by-one for-loop, which is very inefficient.
2. ``Batch.cat`` and ``Batch.cat_`` accepts different arguments.

This commit fixes them, and enables ``cat`` with multiple Batches, with an elegant solution to partially shared keys.

However, there has been several usage of ``a.cat_(b)`` in this repo. Therefore I keep this usage. Both ``a.cat_(b)`` and ``a.cat([b, c, d])`` are acceptable.

tianshou/data/batch.py
```diff
@@ -650,6 +651,20 @@ def empty(batch: 'Batch', index: Union[
...
+        if isinstance(batch, dict):
+            batch = Batch(batch)
+        if batch is not None:
+            for k, v in batch.items():
+                self.__dict__[k] = v
+        if kwargs is not None:
+            batch = Batch(kwargs)
+        if batch is not None:
+            for k, v in batch.items():
+                self.__dict__[k] = v
```
youkaichao: If ``kwargs`` is ``None``, ``batch `` will be used twice.

tianshou/data/batch.py
```diff
@@ -473,66 +470,109 @@ def to_torch(self,
...
     def cat(batches: List[Union[dict, 'Batch']]) -> 'Batch':
         """Concatenate a list of :class:`~tianshou.data.Batch` object into a single
-        new batch.
+        new batch. For keys that are not shared across all batches,
+        batches that do not have these keys will be padded by zeros with
+        appropriate shapes.
+        e.g.
+        a = Batch(a=shape(3,4), common=Batch(c=shape(3, 5)))
+        b = Batch(b=shape(4,3), common=Batch(c=shape(4, 5)))
+        c = Batch.cat([b, c])
```
Trinkle23897: c is not defined

Trinkle23897 (approved):

PR #130, opened 2020-07-11
Fix padding of inconsistent keys with Batch.stack and Batch.cat

youkaichao:
1. For keys that are not shared across all batches, batches that do not have these keys will be padded by zeros. E.g.

```python
>>> a = Batch(a=np.zeros([4, 4]), common=Batch(c=np.zeros([4, 5])))
>>> b = Batch(b=np.zeros([4, 6]), common=Batch(c=np.zeros([4, 5])))
>>> c = Batch.stack([a, b])
>>> c.a.shape
(8, 4)
>>> c.b.shape
(8, 6)
>>> c.common.c.shape
(8, 5)
```

2. simplify ``Batch.cat_``
3. raise exception for stacking with partial keys and axis!=0

### DIFF
--- a/tianshou/data/batch.py
+++ b/tianshou/data/batch.py
@@ -31,8 +31,10 @@ def _create_value(inst: Any, size: int, stack=True) -> Union[
         of (10, 3, 5), otherwise (10, 5)
     """
     has_shape = isinstance(inst, (np.ndarray, torch.Tensor))
-    is_scalar = isinstance(inst, Number) or \
-                issubclass(inst.__class__, np.generic) or (has_shape and not inst.shape)
+    is_scalar = \
+        isinstance(inst, Number) or \
+        issubclass(inst.__class__, np.generic) or \
+        (has_shape and not inst.shape)
     if not stack and is_scalar:
         # here we do not consider scalar types, following the
         # behavior of numpy which does not support concatenation
--- a/tianshou/data/batch.py
+++ b/tianshou/data/batch.py
@@ -3,7 +3,6 @@ import pprint
 import warnings
 import numpy as np
 from copy import deepcopy
-from functools import reduce
 from numbers import Number
 from typing import Any, List, Tuple, Union, Iterator, Optional
 
@@ -530,20 +529,26 @@ class Batch:
                 if isinstance(value, np.ndarray):
                     arrs = []
                     for i, e in enumerate(batches):
-                        shape = [lens[i]] + list(value.shape[1:])
-                        pad = np.zeros(shape, dtype=value.dtype)
-                        arrs.append(e.get(k, pad))
+                        v = e.get(k, None)
+                        if v is None:
+                            shape = [lens[i]] + list(value.shape[1:])
+                            v = np.zeros(shape, dtype=value.dtype)
+                        arrs.append(v)
                     self.__dict__[k] = np.concatenate(arrs)
                 elif isinstance(value, torch.Tensor):
                     arrs = []
                     for i, e in enumerate(batches):
-                        shape = [lens[i]] + list(value.shape[1:])
-                        pad = torch.zeros(shape,
-                                          dtype=value.dtype,
-                                          device=value.device)
-                        arrs.append(e.get(k, pad))
+                        v = e.get(k, None)
+                        if v is None:
+                            shape = [lens[i]] + list(value.shape[1:])
+                            v = torch.zeros(
+                                shape, dtype=value.dtype, device=value.device)
+                        arrs.append(v)
                     self.__dict__[k] = torch.cat(arrs)
                 else:
+                    # here we do not consider scalar types, following the
+                    # behavior of numpy which does not support concatenation
+                    # of zero-dimensional arrays (scalars)
                     raise TypeError(
                         f"cannot cat value with type {type(value)}, we only "
                         "support dict, Batch, np.ndarray, and torch.Tensor")
@@ -576,6 +581,9 @@ class Batch:
         """Stack a list of :class:`~tianshou.data.Batch` object into current
         batch.
         """
+        if len(batches) == 0:
+            return
+        batches = [x if isinstance(x, Batch) else Batch(x) for x in batches]
         if len(self.__dict__) > 0:
             batches = [self] + list(batches)
         keys_map = list(map(lambda e: set(e.keys()), batches))
@@ -593,18 +601,52 @@ class Batch:
                 if not issubclass(v.dtype.type, (np.bool_, np.number)):
                     v = v.astype(np.object)
                 self.__dict__[k] = v
-        keys_partial = reduce(set.symmetric_difference, keys_map)
+        keys_partial = set.union(*keys_map) - keys_shared
         _assert_type_keys(keys_partial)
         for k in keys_partial:
+            is_dict = False
+            value = None
             for i, e in enumerate(batches):
                 val = e.get(k, None)
                 if val is not None:
-                    try:
-                        self.__dict__[k][i] = val
-                    except KeyError:
-                        self.__dict__[k] = \
-                            _create_value(val, len(batches))
-                        self.__dict__[k][i] = val
+                    if isinstance(val, (dict, Batch)):
+                        is_dict = True
+                    else:  # np.ndarray / scalar or torch.Tensor
+                        value = val
+                    break
+            if is_dict:
+                self.__dict__[k] = Batch.stack(
+                    [e.get(k, Batch()) for e in batches], axis)
+            else:
+                if isinstance(value, np.ndarray) \
+                        or issubclass(value.__class__, np.generic) \
+                        or isinstance(value, Number):
+                    if isinstance(value, Number):
+                        value = np.asarray(value)
+                    arrs = []
+                    for i, e in enumerate(batches):
+                        v = e.get(k, None)
+                        if v is None:
+                            v = np.zeros_like(value)
+                        arrs.append(v)
+                    v = np.stack(arrs, axis)
+                    if not issubclass(v.dtype.type, (np.bool_, np.number)):
+                        v = v.astype(np.object)
+                    self.__dict__[k] = v
+                elif isinstance(value, torch.Tensor):
+                    arrs = []
+                    for i, e in enumerate(batches):
+                        v = e.get(k, None)
+                        if v is None:
+                            v = torch.zeros_like(value)
+                        arrs.append(v)
+                    self.__dict__[k] = torch.stack(arrs, axis)
+                else:  # fallback to object
+                    arrs = []
+                    for i, e in enumerate(batches):
+                        v = e.get(k, None)
+                        arrs.append(v)
+                    self.__dict__[k] = np.array(arrs)
 
     @staticmethod
     def stack(batches: List[Union[dict, 'Batch']], axis: int = 0) -> 'Batch':
--- a/tianshou/data/batch.py
+++ b/tianshou/data/batch.py
@@ -535,11 +535,11 @@ class Batch:
                 val = e.get(k, None)
                 if val is not None:
                     try:
-                        self.__dict__[k][sum_lens[i]:sum_lens[i+1]] = val
+                        self.__dict__[k][sum_lens[i]:sum_lens[i + 1]] = val
                     except KeyError:
                         self.__dict__[k] = \
                             _create_value(val, sum_lens[-1], stack=False)
-                        self.__dict__[k][sum_lens[i]:sum_lens[i+1]] = val
+                        self.__dict__[k][sum_lens[i]:sum_lens[i + 1]] = val
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


tianshou/data/batch.py
```diff
@@ -593,23 +601,69 @@ def stack_(self,
                 if not issubclass(v.dtype.type, (np.bool_, np.number)):
                     v = v.astype(np.object)
                 self.__dict__[k] = v
-        keys_partial = reduce(set.symmetric_difference, keys_map)
+        keys_partial = set.union(*keys_map) - keys_shared
```
duburcqa: I think ` - ` operator lack of clarity.

tianshou/data/batch.py
```diff
@@ -593,23 +601,69 @@ def stack_(self,
...
             for i, e in enumerate(batches):
                 val = e.get(k, None)
                 if val is not None:
-                    try:
-                        self.__dict__[k][i] = val
-                    except KeyError:
-                        self.__dict__[k] = \
-                            _create_value(val, len(batches))
-                        self.__dict__[k][i] = val
+                    if isinstance(val, (dict, Batch)):
```
duburcqa: Why replacing something working and efficient by such a complicated nested if/else code ?

tianshou/data/batch.py
```diff
@@ -593,23 +601,69 @@ def stack_(self,
...
+                self.__dict__[k] = Batch.stack(
+                    [e.get(k, Batch()) for e in batches], axis)
+            else:
+                if isinstance(value, np.ndarray) \
+                        or issubclass(value.__class__, np.generic) \
+                        or isinstance(value, Number):
+                    if isinstance(value, Number):
+                        value = np.asarray(value)
+                    arrs = []
+                    for i, e in enumerate(batches):
```
duburcqa: What you are doing is almost duplicated for each if case, maybe you could find a way to factorize this. Duplicated blocks make it harder to maintain.

tianshou/data/batch.py
Trinkle23897: Because the current code of stack_ method has different behavior when dealing with inconsistent keys:
```python
In [2]: a = Batch(a=[1], b=[2], c=[3]) 
   ...: b = Batch(a=[4], b=[5], d=[6]) 
   ...: c = Batch(c=[3], b=[5], d=[9])

In [3]: Batch.cat([a,b,c])
Out[3]: 
Batch(
    b: array([2, 5, 5]),
    a: array([1, 4, 0]),
    c: array([3, 0, 3]),
    d: array([0, 6, 9]),
)

In [4]: Batch.stack([a,b,c])
Out[4]: 
Batch(
    b: array([[2],
              [5],
              [5]]),
)
```
We expect to get the stacked data with four keys.

tianshou/data/batch.py
```diff
@@ -530,20 +529,26 @@ def cat_(self,
...
                     self.__dict__[k] = np.concatenate(arrs)
                 elif isinstance(value, torch.Tensor):
                     arrs = []
                     for i, e in enumerate(batches):
-                        shape = [lens[i]] + list(value.shape[1:])
-                        pad = torch.zeros(shape,
-                                          dtype=value.dtype,
-                                          device=value.device)
-                        arrs.append(e.get(k, pad))
+                        v = e.get(k, None)
```
duburcqa: Please avoid almost duplicated code as much as possible.

tianshou/data/batch.py
duburcqa: It is not related to this part of the code, otherwise this would not work
```python
In [1]: from tianshou.data import Batch

In [2]: a = Batch(a=[1], b=[2], c=[3])

In [3]: b = Batch(a=[4], b=[5], d=[6])

In [4]: Batch.stack([a,b])
Out[4]: 
Batch(
    a: array([[1],
              [4]]),
    b: array([[2],
              [5]]),
    d: array([[0],
              [6]]),
    c: array([[3],
              [0]]),
)
```

duburcqa: You can check by yourself that simply adding your modification line 604 `set.union(*keys_map) - keys_shared` does the trick:
```python
In [1]: from tianshou.data import Batch

In [2]: a = Batch(a=[1], b=[2], c=[3])

In [3]: b = Batch(a=[4], b=[5], d=[6])

In [4]: c = Batch(c=[3], b=[5], d=[9])

In [5]: Batch.stack([a,b,c])
Out[5]: 
Batch(
    b: array([[2],
              [5],
              [5]]),
    c: array([[3],
              [0],
              [3]]),
    a: array([[1],
              [4],
              [0]]),
    d: array([[0],
              [6],
              [9]]),
)
```

Trinkle23897: Batch.stack([a,b]) works but Batch.stack([a,b,c]) does not work. Maybe there is a bug?

duburcqa: Yes it is, see my comment just above :)

youkaichao: Try three Batches, the desired behavior is to remain all keys, but previous code only reserves some keys.

duburcqa: > Try three Batches, the desired behavior is to remain all keys, but previous code only reserves some keys.

How many times I need to tell you to read my comment ?

> You can check by yourself that simply adding your modification line 604 set.union(*keys_map) - keys_shared does the trick

duburcqa: Also, I strongly suggest to simplify your `cat` implementation to follow the same pattern than the current implementation of `stack`, that is clearer and easy to maintain in my opinion.

youkaichao: In the first place, I try to re-implement ``Batch.cat_``, and the "duplicate" code is required (#128 ). Then I remember to re-implement ``Batch.stack`` and copied the code. I'll check later if they are necessary, or you can help me check it :)

Trinkle23897: > Also, I strongly suggest to simplify your `cat` implementation to follow the same pattern than the current implementation of `cat`, that is clearer and easy to maintain in my opinion.

Strongly AGREE!!!

youkaichao: > Also, I strongly suggest to simplify your `cat` implementation to follow the same pattern than the current implementation of `cat`, that is clearer and easy to maintain in my opinion.

I'll try. Or you can help me. ``cat`` is more complicated than ``stack``. The shapes of stack arguments are the same, but they differ in cat.

duburcqa: The way you are dealing all this changes is impossible to follow and I'm thinking to giving up.

You should focus more on how to make existing things works rather than reimplementing stuff. I know its may be frustrating and more difficult, but that's the only way to collaborate with people. Keep in mind other people are not 100% involve in your project.

So it is up to you, but more than 2 PR (<500 lines) a day is clearer impossible to follow.

youkaichao: The reason I want to re-implement is because I do not understand what ``keys_partial = reduce(set.symmetric_difference, keys_map)`` does when having more than two sets. I thought that you depend on some magic that I do not know, so I re-implement it  =_=

youkaichao: Emm, I'm sorry for this.

duburcqa: > Emm, I'm sorry for this.

Don't be. You did nothing wrong but your best I'm pretty sure. Plus, replacing `keys_partial = reduce(set.symmetric_difference, keys_map)` by `keys_partial = set.union(*keys_map) - keys_shared` is VERY GOOD and fixes the bug with `stack` !

I'm not saying that you are doing anything wrong. Just try to keep in mind that open-source is great because it involves a community. So it means several things: many people are happy to spend times to improve the project, BUT, you also need to compromise and sometimes refrain your enthusiasm of changing many stuffs. Other people can only follow small steps. Trust me I know it is frustrating to spend time trying to fix code developed by others instead of doing something new, and that it is time consuming, but it is the only way to get and keep people involved.

youkaichao: Thank you. I have learned so much from you and from this project. ♡

I'll fix them tomorrow :)

duburcqa: Last but not last, you are doing a very good job so far. It just seems you need to learn take time to step back a little bit and get the full picture.

youkaichao:
@duburcqa When I take a further look at the code, I find that ``axis`` arguments are really annoying. Current code actually does not deal with the ``axis`` argument.
```python
try:
    self.__dict__[k][i] = val
except KeyError:
    self.__dict__[k] = \
        _create_value(val, len(batches))
    self.__dict__[k][i] = val
```

The question is, do we really need to support the ``axis`` argument?

Besides testcases, there are only one usage in buffer.py, which is actually not used by any script now.

duburcqa:
> The question is, do we really need to support the axis argument?
Besides testcases, there are only one usage in buffer.py, which is actually not used by any script now.

I guess if you can get it over with than you can remove it. I don't think I use it anywhere in my own code. So I encourage you doing it if it helps making things simpler.

tianshou/data/batch.py
duburcqa: > Thank you. I have learned so much from you and from this project. ♡
I'll fix them tomorrow :)

And it is just the beginning! There is a long way ahead of us

youkaichao:
> > The question is, do we really need to support the axis argument?
> > Besides testcases, there are only one usage in buffer.py, which is actually not used by any script now.
> 
> I guess if you can get it over with than you can remove it. I don't think I use it anywhere in my own code. So I encourage you doing it if it helps making things simpler.

It makes the world much simpler ...
If we support ``axis``, we would need much more code than these 6 lines, but without any practical usage.
What's your opinion, @Trinkle23897

youkaichao:
An example of failure case:
```python
In [109]: a = Batch(a=np.zeros((3, 4))); b=Batch(b=np.zeros((5, 6)))
     ...: c = Batch.stack([a, b], axis=1)
     ...: c.a.shape # should be (3, 2, 4)
Out[109]: (2, 3, 4)

In [110]: c.b.shape # should be (5, 2, 6)
Out[110]: (2, 5, 6)
```

Trinkle23897:
> > > The question is, do we really need to support the axis argument?
> > > Besides testcases, there is only one usage in buffer.py, which is actually not used by any script now.
> > 
> > 
> > I guess if you can get it over with than you can remove it. I don't think I use it anywhere in my own code. So I encourage you to do it if it helps to make things simpler.
> 
> It makes the world much simpler ...
> If we support `axis`, we would need much more code than these 6 lines, but without any practical usage.
> What's your opinion, @Trinkle23897

In buffer, current implementation secures the consistency of shape when performing Batch.stack, because when we store the data into the buffer, it will automatically create those padding. So if you do not will to implement, you can raise an exception at the end of Batch.stack_.

tianshou/data/batch.py
```diff
@@ -24,28 +23,45 @@ def _is_batch_set(data: Any) -> bool:
...
+    has_shape = isinstance(inst, (np.ndarray, torch.Tensor))
+    is_scalar = \
+        isinstance(inst, Number) or \
+        issubclass(inst.__class__, np.generic) or \
+        (has_shape and not inst.shape)
+    if not stack and is_scalar:
+        # here we do not consider scalar types, following the
+        # behavior of numpy which does not support concatenation
+        # of zero-dimensional arrays (scalars)
+        raise TypeError(f"cannot cat {inst} with which is scalar")
```
duburcqa, line 42: Perhaps it would be better to raise this exception at `cat` and `stack` level ? I know it is kind of code duplication, but in the particular case of raising an exception it seems appropriate to simplify the trace, since `_create_value` is not an "official" method of Tianshou and only used to factorize the code. Moreover, only `cat` can raise this exception. But as you want.

duburcqa (approved):

duburcqa:
@youkaichao  I just wanted to stress the fact that you did a very nice job with this PR. As you can see, it was more work, but at the end the quality is much higher. With far less modifications, it fixes some bugs, and implement new features, without messing with code base and without any code duplication. That's what I call high-quality contribution. :+1:

tianshou/data/batch.py
youkaichao: Yeah, the trace will start with ``_create_value ``, but followed by ``cat``. This should be okay. I have seen many exceptions (like numpy) starting with an internal method. Users should be able to follow the trace to find the root cause.

youkaichao:
> @youkaichao I just wanted to stress the fact that you did a very nice job with this PR. As you can see, it was more work, but at the end the quality is much higher. With far less modifications, it fixes some bugs, and implement new features, without messing with code base and without any code duplication. That's what I call high-quality contribution. 👍

Yes, I feel the code becomes much better. Thank you for the comment, otherwise these lines of complicated code will remain and make the maintenance much harder.

I too enjoy the process of collaborating with you in this project.

duburcqa:
> otherwise these lines of complicated code will remain and make the maintenance much harder.

Yes, that's my greatest fear when I see comment in any project saying "I will do it later in another PR". Often people forgot and it is never done, and things get messy without really realizing it. Then at some point it is too late and, it is necessary to start over from scratch over and over again.

PR #134, opened 2020-07-13
Asynchronous sampling vector environment

### LINKED ISSUE #103: Async Sampling

duburcqa:
Fix #103


duburcqa:
ping @youkaichao

youkaichao:
Roger that. Keep following this pr.

duburcqa:
@youkaichao If you want to help me on that, do not hesitate. I'm quite busy right now. I can find time for reviews, but not to code something new, for at least the next 2 weeks.

youkaichao:
Ok, I'll take this pr.

youkaichao:
@duburcqa  please invite me to your repo so that I can work on your branch.

youkaichao:
The head is now at b99f48423f1adb39353310cdc376c9022599c8cb . I think we should only modify ``Collector.collect``.

duburcqa:
> The head is now at b99f484 . I think we should only modify Collector.collect.

Yes I think so, but are you sure `AsyncVectorEnv` is not useful ?

youkaichao:
> are you sure AsyncVectorEnv is not useful ?

Emmm, my bad. After a second thought, I thin it is useful.

duburcqa:
> Emmm, my bad. After a second thought, I thin it is useful.

Yes, I think both are needed.

youkaichao:
After #157 , this should be done today ✌️

duburcqa:
Nice !

tianshou/env/vecenv.py
```diff
@@ -180,6 +180,93 @@ def close(self) -> List[Any]:
...
+             ) -> Tuple[np.ndarray, np.ndarray, np.ndarray, np.ndarray]:
+        """
+        Provide the given action to the environments specified by the id.
+        If action is None, fetch unfinished step() calls instead.
+        """
+        if id is not None:
+            raise ValueError("cannot specify the id of environments"
+                             " during step() for AsyncVectorEnv")
+        if action is not None:
+            assert len(action) + len(self.waiting_id) <= self.env_num
```
duburcqa: I'm afraid it is not working like that. Every action is specific to a given environment and cannot be executed by all of them indifferently. Indeed, the action has been computed based on a given observation of the environment, thus this exact same environment must be used to perform a step using this action.

youkaichao: Nice analysis! I should record the env id of returned observations last time, and use that env id.

youkaichao: @duburcqa  please take a look at ef09ac2 😊

tianshou/env/vecenv.py
```diff
@@ -180,6 +180,101 @@ def close(self) -> List[Any]:
...
+        # all environments are ready in the beginning
+        self.ready_id = list(range(self.env_num))
+
+    def reset(self, id: Optional[Union[int, List[int]]] = None) -> np.ndarray:
+        if id is None:
+            id = range(self.env_num)
+            # reset all environments
+            self.ready_id = list(range(self.env_num))
+        elif np.isscalar(id):
+            assert id in self.ready_id, f'can only reset ' \
```
duburcqa: why this check is only for scalar id ?

tianshou/env/vecenv.py
```diff
@@ -180,6 +180,101 @@ def close(self) -> List[Any]:
...
+             ) -> Tuple[np.ndarray, np.ndarray, np.ndarray, np.ndarray]:
+        """
+        Provide the given action to the environments. For the first call,
+        action should be the action for each environment; for the consecutive
+        call, the action sequence should correspond to the ``env_id`` filed
+        in the last returned ``info``
+        If action is None, fetch unfinished step() calls instead.
+        """
+        if id is not None:
+            raise ValueError("cannot specify the id of environments"
```
duburcqa: I would rather accept `id` argument to be a subset of : "``env_id`` filled in the last returned ``info``"

tianshou/env/vecenv.py
```diff
@@ -180,6 +180,101 @@ def close(self) -> List[Any]:
...
+        return super().close()
+
+    def step(self,
+             action: Optional[np.ndarray],
+             id: Optional[Union[int, List[int]]] = None
+             ) -> Tuple[np.ndarray, np.ndarray, np.ndarray, np.ndarray]:
+        """
+        Provide the given action to the environments. For the first call,
+        action should be the action for each environment; for the consecutive
+        call, the action sequence should correspond to the ``env_id`` filed
```
duburcqa: type: "filled"

tianshou/env/vecenv.py
```diff
@@ -180,6 +180,101 @@ def close(self) -> List[Any]:
...
+        self.step(None)
+        return super().close()
+
+    def step(self,
+             action: Optional[np.ndarray],
+             id: Optional[Union[int, List[int]]] = None
+             ) -> Tuple[np.ndarray, np.ndarray, np.ndarray, np.ndarray]:
+        """
+        Provide the given action to the environments. For the first call,
+        action should be the action for each environment; for the consecutive
```
duburcqa: This is a nice default behavior, but it is a but restrictive. Why the user should have to send actions for every ready env ? That's the all point of `id` arg. (See my comment below.)

tianshou/env/vecenv.py
duburcqa: Much better now !

tianshou/env/vecenv.py
youkaichao: Nice catch! Fixed in c6caf95 .

tianshou/env/vecenv.py
youkaichao: There are two alternatives, one is the current code, the other is as you said, let id be a subset of env_id filled in the last returned info. In the latter case, ``Collector.collect`` would need to pass this argument again and again, and I don't think this has practical significance. The whole point of asynchronous simulation is to step those ready environments as soon as possible, so I take the former alternative, assume the observation returned in the last step all yield corresponding action.

If you have some usecase of the latter alternative, please tell me.

duburcqa: I see your point. Indeed it is not intended to be used in `Collector.collect`. But I like to use `Tianshou.data` for many stuffs far from RL. Part of my work is simulation VS reality analysis. For this, I'm not using any collector. Just raw Batch and VectorEnv. Then I run simulations by hand step by step to carefully analyse what is happening. 

To wrap up, try to think outside the box. Tianshou is desgined for RL, but part of it goes far beyond. Especially Batch and VecEnv. Making them as versatile as possible may lead to unexpected yet very interesting use cases, from which Tianshou could get better if we seize the opportunity.

In this particular case, it is not a lot of effort but makes it significantly more flexible.

youkaichao: That's reasonable, suggestion taken in ecec577 . I call this "partial async simulation".

duburcqa:
Well done! I have no more comment right now.

youkaichao:
Here I present a dirty implementation first to request for comment.

The dirty things are those functions added in batch.py.

Currently, I need them because:

1. support creating elements during indexing: ``x = Batch(a=Batch()); x[1, 2] = Batch(a=[5, 6])`` => ``x = Batch(a=[0, 5, 6, ...])``

2. support ignoring empty batches during indexing: ``x = Batch(a=[1, 2, 3]); x[1, 2] = Batch(a=Batch())`` => ``x = Batch(a=[1, 2, 3])``

3. copy reserved keys only and ignore non-reserved keys: ``x = Batch(a=[1, 2, 3], b=Batch()); y = Batch(a=[4, 5, 6], b=[0, 1, 2]);`` => ``x = Batch(a=[1, 2, 3], b=[0, 1, 2])``

These subtleties come from the fact that, in async simulation, the data is initially empty (because we do not know the shape), and the first returned data are not full shape (only a subset of them), so we have to do indexing on empty Batches.

Finally, sometimes the testcase of async collector works but sometimes not. Seems to related with randomness and maybe improper disposal of async simulation.

duburcqa:
> These subtleties come from the fact that, in async simulation, the data is initially empty (because we do not know the shape), and the first returned data are not full shape (only a subset of them), so we have to do indexing on empty Batches.

Are you sure? You have access to `venv.envs_num` so you know the size to reserve.

duburcqa:
> Finally, sometimes the testcase of async collector works but sometimes not. Seems to related with randomness and maybe improper disposal of async simulation.

You could set the random seed, yet it should be more robust.

youkaichao:
> Are you sure? You have access to venv.envs_num so you know the size to reserve.

That is only the first dimension, not the whole shape.

youkaichao:
I find similar code in ``ReplayBuffer._add_to_buffer``:

```python
        try:
            value[self._index] = inst
        except KeyError:
            for key in set(inst.keys()).difference(value.__dict__.keys()):
                value.__dict__[key] = _create_value(inst[key], self._maxsize)
            value[self._index] = inst
```

Maybe it means setting index with empty Batch is a common requirement and we can support something like ``value[(index, pad_length)] = inst``, where ``pad_length`` means how long to pad when ``value[key]`` is empty but ``inst[key]`` not.

duburcqa:
> I find similar code in ReplayBuffer._add_to_buffer:

Yes bu the whole point is that it is not part of Batch, since in my opinion it is not a good practice to allow allocation using assignment operator. I find it confusing and it is not numpy compliant.

duburcqa:
> That is only the first dimension, not the whole shape.

Yes, but it you know this dimension, and you have access to a first returned data, you have everything to reserve the right size.

duburcqa:
I'm not against custom behaviour, but I would rather a dedicate method in `collector` or `buffer` rather than make it standard in `Batch`, if it is borderline (as it is the case for `ReplayBuffer._add_to_buffer`). I think the changes you are proposing fall into this category.

youkaichao:
> I would rather a dedicate method in collector or buffer rather than make it standard in Batch, if it is borderline

That's why I call it a dirty implementation :)  

There are two possible fixes:

1. collector and buffer use their custom code to manipulate Batch, do not modify Batch, but result in duplicate code.

2. add new behavior for Batch so that collector and buffer can make use of Batch api.

> in my opinion it is not a good practice to allow allocation using assignment operator. I find it confusing and it is not numpy compliant.

I think that's not a good reason, though. Reserved keys are unique features of Batch, not numpy. It can be expected that they differ with some behavior.

youkaichao:
In terms of the behavior of Batch, I want to mention another point of setitem:

```python
x = Batch(a=np.ones((3, 4)), b=np.ones((3, 4)))
y=Batch(b=[0, 1, 2, 3])
x[0] = y
x
Batch(
    a: array([[0., 0., 0., 0.],
              [1., 1., 1., 1.],
              [1., 1., 1., 1.]]),
    b: array([[0., 1., 2., 3.],
              [1., 1., 1., 1.],
              [1., 1., 1., 1.]]),
)
```

In the above example, do we really need to set ``x.a[0]`` to zero ?

duburcqa:
> I think that's not a good reason, though. Reserved keys are unique features of Batch, not numpy. It can be expected that they differ with some behavior.

Unique feature is clearly not the same than different behaviour is the same situation.

duburcqa:
> collector and buffer use their custom code to manipulate Batch, do not modify Batch, but result in duplicate code.

Or you could add a specialisation of `Batch` that inherits from `Batch` and have a more complex/borderline behavior than you can use everywhere since `isinstance` should work with both indifferently.

duburcqa:
> In the above example, do we really need to set x.a[0] to zero ?

Good point. Ideally it should fail but we cannot afford it. Personally I rely on the current behavior (hence why it is like this), because It enables "sparse" assignment.

youkaichao:
> Personally I rely on the current behavior (hence why it is like this)

Can you provide some use cases so we understand the rationale?

In #142 , we specify the standard behavior of aggregating heterogeneous batches. Now we are talking about assignment with heterogeneous batches. This is worth discussing and not a custom behavior.

From the rationale of reserved keys, I think assignment with heterogeneous batches should be re-considered and the behavior in current release 0.2.5 is too restrictive.

duburcqa:
> Can you provide some use cases so we understand the rationale?

`Info` output of my environment is a dict whose field are the original mixture of reward that is used compute the aggregated one, for instance: `{'stability': 0.1, 'duration': 0.2, 'deviation': -0.1}`. Whose components are changing over iteration. For instance at some point you may get: `{'avoidance': 1.0, 'periodicity': 0.2}`, and for the last step `{'terminal': 2.0}`. In such a case, if any already existing component is not provided at a given step, it means that is not used for aggregated reward computation, which is equivalent to setting them to 0.

> In #142 , we specify the standard behavior of aggregating heterogeneous batches. Now we are talking about assignment with heterogeneous batches. This is worth discussing and not a custom behavior.
>
> From the rationale of reserved keys, I think assignment with heterogeneous batches should be re-considered and the behavior in current release 0.2.5 is too restrictive.

Yes you can open an issue.

youkaichao:
> ... Whose components are changing over iteration. ... Whose components are changing over iteration ...

This is easy to fix, the components are changing, but you should know how many possible components you have, so just return a dict with all components, and set those missing components to zero. No need to support this at the level of Batch when it can be easily dealt with by users.

duburcqa:
> This is easy to fix, the components are changing, but you should know how many possible components you have, so just return a dict with all components, and set those missing components to zero. No need to support this at the level of Batch when it can be easily dealt with by users.

Yes I know, but I'm doing it because having consistent fields for `info` is not part of gym env specification so you cannot rely on it. We already discussed this point with @Trinkle23897, and there already a unit checking exactly this.

youkaichao:
Discussion opened in #163 . @Trinkle23897  @duburcqa

youkaichao:
@Trinkle23897 @duburcqa  should be ok now, please have a look when you are available :)

tianshou/data/collector.py
```diff
@@ -320,3 +369,24 @@ def sample(self, batch_size: int) -> Batch:
...
+            vs = source.get(k, Batch())
+            if isinstance(vs, Batch) and vs.is_empty():
+                # case 2
+                # use __dict__ to avoid many type checks
+                source.__dict__[k] = _create_value(v[0], size)
+        else:
+            # target[k] is reserved
+            # case 1
+            continue
+        source.__dict__[k][indices] = v
```
duburcqa, line 392: try/except should be more efficient, since the other cases seldom occur.
```
    for k, v in target.items():
        try:
            source.__dict__[k][indices] = v
        except KeyError:
            if not isinstance(v, Batch) or not v.is_empty():
                # target[k] is non-reserved
                vs = source.get(k, Batch())
                if isinstance(vs, Batch) and vs.is_empty():
                    # case 2
                    # use __dict__ to avoid many type checks
                    source.__dict__[k] = _create_value(v[0], size)
            source.__dict__[k][indices] = v
```

But it's not a big deal.

youkaichao: I feel uncomfortable with explicitly invoking an exception and then catching it so I write it this way 😄

duburcqa: I few months ago I would have said exactly the same, but it turns out try/except is way more efficient in practice if the exception is very seldom raised, as it should be the case here. You should have a look at [this post](https://stackoverflow.com/a/7604717/4820605) for instance. But it is up to you !

It is related to Python principle "EAFP": Easier to ask for forgiveness than permission. It is amazing to see how Python reasoning is far from C++.

youkaichao: > it turns out try/except is way more efficient in practice if the exception is very seldom raised

That's an interesting point. The post is interesting, too. If someday I meet efficiency bottleneck then I'll consider something like this, but if not, I prefer the straightforward way to make it clear for future maintenance.

duburcqa:
I approve this PR (but I cannot actually approve it since I opened it :sweat_smile:).

tianshou/data/collector.py
```diff
@@ -139,6 +150,7 @@ def reset_env(self) -> None:
         """Reset all of the environment(s)' states and reset all of the cache
         buffers (if need).
         """
+        self._ready_env_ids = np.array(range(self.env_num))
```
duburcqa: np.arange(self.env_num)

youkaichao (dismissed):

tianshou/env/vecenv/asyncenv.py
```diff
@@ -0,0 +1,103 @@
...
+        """
+        :param wait_num: used in asynchronous simulation if the time cost of
+            ``env.step`` varies with time and synchronously waiting for all
+            environments to finish a step is time-wasting. In that case, we
+            can return when ``wait_num`` environments finish a step and keep
+            on simulation in these environments. If ``None``, asynchronous
+            simulation is disabled; else, ``1 <= wait_num <= env_num``.
+        """
+        super().__init__(env_fns)
+        self.wait_num = wait_num or len(env_fns)
```
Trinkle23897, line 28: So if `wait_num` is None, this AsyncVenv is equal to the original `SubprocVenv`. Why not add this directly into the SubprocVenv?
In other words, the sync mode is a special case of async mode, thus we can unify the collector to use async mode for all venv instead of sync mode, and implement async in all venv.

Trinkle23897: Maybe we can create another issue and pr

youkaichao: Good point, and this can go further: why not just make it an ability of BaseVecEnv so that async simulation directly works for any vec envs? This seems worthwhile, but is a breaking change. Maybe we need to abstract Worker and change the interface of vec env.

Trinkle23897 (approved):

youkaichao (approved):

tianshou/env/vecenv/asyncenv.py
duburcqa: Having a unified VecEnv would be awesome and it may greatly ease the implementation of other classes/methods such as Collector.

duburcqa: > In other words, the sync mode is a special case of async mode, thus we can unify the collector to use async mode for all venv instead of sync mode, and implement async in all venv.

That's what I was thinking x)

duburcqa: @Trinkle23897 But I think adding this feature in a dedicated class as we did is a good start I think, since it may be break changing as said.

PR #122, opened 2020-07-09
Add multi-agent example: tic-tac-toe

youkaichao:

PR #137, opened 2020-07-14
Standardized behavior of Batch.cat and misc code refactor

### LINKED ISSUE #131: Remove some useless kwargs or not
For some functions (like ``Collector.__init__``), @Trinkle23897 keep ``**kwargs`` arguments for api stability, so that adding a new argument with default values can work with older versions of tianshou.

However, I tend to remove these unused ``kwargs`` because they may hide some bugs and make it difficult for IDE to report unexpected arguments.

We need some discussions to determine what to do next. Suggestions are welcome especially by @duburcqa . Must we keep the ability of old versions of tianshou to work with code written for new versions of tianshou?

### LINKED ISSUE #138: The length of empty Batch
Currently, ``len(Batch())`` will raise an error.

The numpy behavior is:
```python
a = np.array([])
len(a) # 0
b = np.array(3)
len(b) # exception
```

So, len of an empty Batch should be 0. len of a Batch with scalars should raise an error.

Besides, if we want to support ``cat`` with empty Batches, we should also make ``len(Batch()) == 0``, otherwise we cannot tell ``Batch()`` from ``Batch(a=1)``, where ``Batch()`` can be used in ``cat`` but the scalar ``Batch(a=1)`` cannot.

@Trinkle23897  @duburcqa

### LINKED ISSUE #139: Standardize the behavior of Batch aggregation (stack/cat) when dealing with reserved keys
Currently, we use ``Batch()`` to indicate that a key reserves the place in Batch and it will have value later on. It works fine in most cases, but now we have noticed that it is problematic. 

The critic problem is: how to treat ``Batch(c=Batch())``? There are two opinions:

- ``Batch(c=Batch())`` means hierarchical key reservation, and should be treated as empty. But the problem is, we want to enable the concatenation of Batches with other empty Batches, which is seemingly natural. But considering ``Batch.cat([Batch(a=np.random.rand(3, 4)), Batch(a=Batch(c=Batch()))])``, we would want to treat ``Batch(c=Batch())`` as non-empty.

- ``Batch(c=Batch())`` is not empty, and there is no need to support hierarchical key reservation. This makes the implementation straightforward, but does not support hierarchical key reservation. Unfortunately, there are some use cases of hierarchical key reservation in Tianshou.

I think the critical problem is how to reserve keys and support hierarchical key reservation.

@Trinkle23897  @duburcqa

youkaichao:
Misc code refactor and add Batch.condense

These changes are picked from #122 to be an independent PR.


tianshou/data/batch.py
```diff
@@ -674,6 +675,14 @@ def empty(batch: 'Batch', index: Union[
         """
         return deepcopy(batch).empty_(index)
 
+    def condense(self):
```
duburcqa: Why is this useful? Plus I don't find this name very informative.

youkaichao: When ``cat`` does not support to concatenate a list of batches with some fields filled by empty Batch() (the default behavior of ``numpy.concatenate``), but we need this feature for multi-agent RL. To be consistent with ``numpy``, we can remove empty fields first, and then ``cat`` them.

The functionality is to remove fields with empty Batch, it is ok if you have another name.

duburcqa: Why not to support empty field in `cat` instead ? Using the "key reservation" paradigm (empty Batch are not regular keys). It is a feature that it not part of numpy.

tianshou/data/batch.py
```diff
@@ -376,8 +376,9 @@ def __setitem__(self, index: Union[
             except KeyError:
                 if isinstance(val, Batch):
                     self.__dict__[key][index] = Batch()
-                elif isinstance(val, np.ndarray) and \
-                        issubclass(val.dtype.type, (np.bool_, np.number)):
+                elif isinstance(val, torch.Tensor) or \
```
duburcqa, line 415: I noticed this yesterday but I didn't care to fix it x) Nice thing that you did it !

tianshou/policy/base.py
```diff
@@ -149,7 +151,7 @@ def compute_nstep_return(
         target_q_fn: Callable[[ReplayBuffer, np.ndarray], torch.Tensor],
         gamma: float = 0.99,
         n_step: int = 1,
-        rew_norm: bool = False
+        rew_norm: bool = False,
```
duburcqa, line 154: typo I guess.

tianshou/policy/base.py
```diff
@@ -180,16 +182,17 @@ def compute_nstep_return(
...
+            bfr = rew[:min(len(buffer), 1000)]  # avoid large buffer
             mean, std = bfr.mean(), bfr.std()
             if np.isclose(std, 0):
                 mean, std = 0, 1
         else:
             mean, std = 0, 1
         returns = np.zeros_like(indice)
         gammas = np.zeros_like(indice) + n_step
-        done, rew, buf_len = buffer.done, buffer.rew, len(buffer)
+        done, rew, buf_len = buffer.done, rew, len(buffer)
```
duburcqa: rew = rew ?

tianshou/data/batch.py
youkaichao: I tried, but that's very annoying. Take a code snippet for example:
```python
b1 = Batch(a=np.random.rand(3, 4), common=Batch(c=np.random.rand(3, 5)))
b2 = Batch(b=torch.rand(4, 3), common=Batch(c=np.random.rand(4, 5)))
test = Batch.cat([b1, b2])
```
For partial key 'a', 'a' is not a key of b2, so it is perfectly ok to fill zero in b2. If we support b1.a is Batch() in ``cat``, what is the output ``test.a`` supposed to be? Furthermore, if both b1.a and b2.a is Batch(), what is the output ``test.a`` supposed to be?

If these are made clear, I'm happy to support ``cat`` with empty Batch().

tianshou/data/batch.py
youkaichao: The exception comes up when I write MARL scripts, so I fixed it by the way.

tianshou/data/batch.py
youkaichao: For shared keys, it is also very annoying. Currently, we are passing ``cat`` to ``np.concatenate`` for shared keys. The support for ``cat`` with Batch() would make the code much complicated, essentially we need to duplicate the code of dealing with partial keys.

youkaichao: One possible solution would be to re-define shared keys, like ``keys_map = [set(k for k, v in batch.items() if not (isinstance(v, Batch) and v.is_empty())) for batch in batches]``, omitting keys with empty Batch() when calculating shared keys.

Trinkle23897: My opinion is

> If we support `b1.a` is Batch() in `cat`, what is the output `test.a` supposed to be? 

`Batch()`, since `b1.a` is `Batch()` and `b2.a` does not exist.

> Furthermore, if both `b1.a` and `b2.a` is Batch(), what is the output `test.a` supposed to be?

`Batch()`.

But it should define that `len(Batch()) == 0` I guess?

duburcqa: > One possible solution would be to re-define shared keys, like keys_map = [set(k for k, v in batch.items() if not (isinstance(v, Batch) and v.is_empty())) for batch in batches], omitting keys with empty Batch() when calculating shared keys.

Exactly.  You just need to compute 3 set of keys: 
- shared keys -> Not empty for all batches
- partial keys -> Empty for at least one of the batches AND not empty for  at least one of the batches
- reserved keys -> Empty for at least one of the batches

duburcqa: > If we support b1.a is Batch() in cat, what is the output test.a supposed to be?

`b1.a` is `np.random.rand(3, 4)`, and `b2.a` is `Batch()`, so it is the (newly refined) "partial keys" case.

> if both b1.a and b2.a is Batch(), what is the output test.a supposed to be?

if both `b1.a` and `b2.a` are `Batch()`, then it is the (newly refined) "reserved keys" case. So you should only reserve the key in the concatenated batch.

duburcqa: I think the 3 cases are pretty clear, natural, and quite straightforward to implement. You just need to refine the computation of "shared keys" and "partial keys", and add a new 5 lines of code "reserved keys" handling.

youkaichao: > > One possible solution would be to re-define shared keys, like keys_map = [set(k for k, v in batch.items() if not (isinstance(v, Batch) and v.is_empty())) for batch in batches], omitting keys with empty Batch() when calculating shared keys.
> 
> Exactly. You just need to compute 3 set of keys:
> 
> * shared keys -> Not empty for all batches
> * partial keys -> Empty for at least one of the batches AND not empty for  at least one of the batches
> * reserved keys -> Empty for at least one of the batches

Okay, reserved keys are treated the same as partial keys.

@Trinkle23897 
> But it should define that len(Batch()) == 0 I guess?

No, we can first filter out empty batches by ``batches = [b for b in batches if len(b.__dict__)]``.

youkaichao: > I think the 3 cases are pretty clear, natural, and quite straightforward to implement. You just need to refine the computation of "shared keys" and "partial keys", and add a new 5 lines of code "reserved keys" handling.

What about ``stack``? The same semantic meaning may apply, too.

tianshou/policy/base.py
youkaichao: A trailing comma is okay, though. Future modifications then will not include this comma.

tianshou/data/batch.py
duburcqa: Yes it should.

youkaichao:
Currently, ``cat`` only supports partially empty Batch, that is, a non-empty Batch but has some fields with empty Batch.

To enable ``cat`` with empty Batch like ``Batch(a=Batch(), b=Batch(c=Batch()))``, we have to know its length, which should be resolved after #138 .

duburcqa:
> To enable cat with empty Batch like Batch(a=Batch(), b=Batch(c=Batch())), we have to know its length, which should be resolved after #138 .

`is_empty` is provided in such case. No need to implement change or implement anything else.

youkaichao:
```python
Batch.cat([Batch(a=np.random.rand(3, 4)), Batch(b=Batch(), a=Batch(c=Batch()))])
```
@duburcqa What about this? Maybe this should cause an error? Then we should distinguish between Batch() and Batch(c=Batch()), where the former is used to reserve keys but the latter should not be treated as empty.

This raises a natural question: is it ok to use Batch() to reserve keys? Or maybe we should explicitly reserve the keys? (e.g. have a field named ``self.reserved_keys``).

Or, must we reserve any keys? Maybe we can reserve no keys, return a special token to indicate the absence of a key, and let the caller to deal with it.

One more issue: ``None`` is also a valid object (``Batch(a=None)`` works just fine), but we are using ``b.get(k, None) is None`` to judge the existence of a key.

duburcqa:
> @duburcqa What about this? 

`Batch(c=Batch())` is not an empty Batch -> `a=Batch(c=Batch())` is not a key reservation -> the shape of `Batch(b=Batch(), a=Batch(c=Batch()))` is not compatible with the shape of `Batch(a=np.random.rand(3, 4))` -> it should raise an exception. I see no ambiguity here.

> This raises a natural question: is it ok to use Batch() to reserve keys? Or maybe we should explicitly reserve the keys? (e.g. have a field named self.reserved_keys).
Or, must we reserve any keys? Maybe we can reserve no keys, return a special token to indicate the absence of a key, and let the caller to deal with it.
One more issue: None is also a valid object (Batch(a=None) works just fine), but we are using b.get(k, None) is None to judge the existence of a key.

I dont know. But those design issues should be discussed in another dedicated issue + PR.

youkaichao:
> I see no ambiguity here.

Currently, ``Batch(c=Batch())`` is empty. And maybe someone would want to use ``Batch(b=Batch(), a=Batch(c=Batch()))`` for hierarchical key reservation?

duburcqa:
> Currently, Batch(c=Batch()) is empty. 

Yes, but I don't think it should (at least for this PR).

> And maybe someone would want to use Batch(b=Batch(), a=Batch(c=Batch())) for hierarchical key reservation?

Maybe, but I think it is another issue since it requires changing of paradigm. But it my opinion, it is irrelevant to support hierarchical key reservation. I may be wrong.

Trinkle23897:
Hierarchical key reservation is a must for the current collector implementation, since it will create something like self.data is a `Batch(obs=..., ..., policy=Batch(_state=Batch()), ...)`

youkaichao:
> Currently, Batch(c=Batch()) is empty.

Now we support ``is_empty(recursive=bool)`` to identify either ``Batch()`` or ``Batch(c=Batch())``.

> Hierarchical key reservation is a must for the current collector implementation

Hierarchical key reservation is stays.

With 6d2cda6 , now ``Batch.cat_`` behaves exactly as described [here](https://github.com/thu-ml/tianshou/issues/139#issuecomment-659092494).

youkaichao:
> One more issue: None is also a valid object (Batch(a=None) works just fine), but we are using b.get(k, None) is None to judge the existence of a key.

2541bec fixes it.

duburcqa:
> Now we support is_empty(recursive=bool) to identify either Batch() or Batch(c=Batch()).

Torch is using  `recurse` not `recursive`.

tianshou/data/batch.py
```diff
@@ -14,11 +14,17 @@
...
+    # where each element is a dict/Batch object
     if isinstance(data, (list, tuple)):
         if len(data) > 0 and all(isinstance(e, (dict, Batch)) for e in data):
             return True
     elif isinstance(data, np.ndarray) and data.dtype == np.object:
-        if all(isinstance(e, (dict, Batch)) for e in data.tolist()):
+        # ``for e in data`` will just unpack the first dimension,
+        # but data.tolist() will flatten ndarray of objects
+        # so do not use data.tolist()
+        if all(isinstance(e, (dict, Batch)) for e in data):
```
duburcqa, line 27: Nice catch !

tianshou/data/batch.py
```diff
@@ -502,49 +546,92 @@ def to_torch(self, dtype: Optional[torch.dtype] = None,
...
 
     def cat_(self,
-             batches: Union['Batch', List[Union[dict, 'Batch']]]) -> None:
+             batches: Union['Batch', List[Union[dict, 'Batch']]],
+             lens: Optional[List[int]] = None
+             ) -> None:
         """Concatenate a list of (or one) :class:`~tianshou.data.Batch` objects
         into current batch.
+        If x=Batch(a=Batch(a=np.random.randn(3, 4)), b=np.random.randn(3, 4)),
+        y = Batch(a=Batch(a=Batch()), b=np.random.randn(3, 4)), if we want to
```
duburcqa: Reserving keys `a=Batch(a=Batch())` should not be necessary.

tianshou/data/batch.py
```diff
@@ -502,49 +546,92 @@ def to_torch(self, dtype: Optional[torch.dtype] = None,
...
         batches = [x if isinstance(x, Batch) else Batch(x) for x in batches]
-        if len(self.__dict__) > 0:
+        if lens is None:
+            # remove empty Batch, and infer lens
+            batches = [x for x in batches if not x.is_empty()]
+            try:
+                lens = [0 if x.is_empty(recursive=True) else len(x)
+                        for x in batches]
+            except TypeError as e:
+                e2 = Exception(
```
duburcqa: Never use `Exception`. Too generic. `ValueError` is more appropriate.

tianshou/data/batch.py
```diff
@@ -502,49 +546,92 @@ def to_torch(self, dtype: Optional[torch.dtype] = None,
...
         """
         if isinstance(batches, Batch):
             batches = [batches]
         if len(batches) == 0:
             return
         batches = [x if isinstance(x, Batch) else Batch(x) for x in batches]
-        if len(self.__dict__) > 0:
+        if lens is None:
+            # remove empty Batch, and infer lens
+            batches = [x for x in batches if not x.is_empty()]
```
duburcqa: Why not to use only the `try...catch` black ?

tianshou/data/batch.py
```diff
@@ -502,49 +546,92 @@ def to_torch(self, dtype: Optional[torch.dtype] = None,
                 v.to_torch(dtype, device)
 
     def cat_(self,
-             batches: Union['Batch', List[Union[dict, 'Batch']]]) -> None:
+             batches: Union['Batch', List[Union[dict, 'Batch']]],
+             lens: Optional[List[int]] = None
```
duburcqa: I don't like such the idea of using hint.

youkaichao: A better solution is welcome. But it seems unavoidable for the case I give:

```python
x=Batch(a=Batch(a=np.random.randn(3, 4)), b=np.random.randn(3, 4))
y = Batch(a=Batch(a=Batch()), b=np.random.randn(3, 4))
Batch.cat([x, y])
```

tianshou/data/batch.py
youkaichao: This is hierarchical key reservation, and is indeed required for tianshou:

> Hierarchical key reservation is a must for the current collector implementation, since it will create something like self.data is a `Batch(obs=..., ..., policy=Batch(_state=Batch()), ...)`

tianshou/data/batch.py
duburcqa: Yes, this issue was already discussed in the past. It is possible. You need to do something like this:

```
def cat(batches):
    def cat_impl(batches, lens):
        [...] # actual stack implementation

    lens = [...] # compute len here
    return cat_impl(batches, lens)
```

This way, it hides the implementation of `cat` that is used internally, because the `lens` parameter should not appears at top level in my opinion. It is quite easy to adapt your code to use this pattern, `cat_impl` is your current implementation of `cat`.

youkaichao: Agree that ``lens `` should not be exposed to users. But creating a nested function would make the indentation too deep. I changed it into an internal function ``__cat``.

tianshou/data/batch.py
duburcqa: That's not right. It should be possible to do this:
```
>>> Batch.stack([Batch(a=np.array([1])), Batch(a=np.array([2]), b=np.array([3]))])
Batch(
    a: array([[1],
              [2]]),
    b: array([[0],
              [3]]),
)
```

This is very likely to append in practice (and it happens in some project of my company). For example, concatenating environment `info`, the keys of the last step (when done flag is True) can be completely different, or a least some new ones can be present, for instance in the case of a terminal reward.

tianshou/data/batch.py
youkaichao: Can you elaborate on that? I do not get it.

tianshou/data/batch.py
duburcqa: By the way, take advantage of this structure to make `lens` argument required and not optional in `cat_impl`, and move the computation of the initial value outside `cat_impl`. It will make the complete implementation clearer and simpler.

duburcqa: > But creating a nested function would make the indentation too deep. I changed it into an internal function __cat.

As you wish. It is just an implementation detail.

tianshou/data/batch.py
duburcqa: I was just wonderting if this line is necessary since you are using recursive `is_empty` in the try/catch block.

youkaichao: Oh, this is necessary. ``is_empty()`` and ``is_empty(recursive=True)`` mean different things and have to be dealt with differently.

For ``is_empty() == True``, this is ``Batch()``, and will be ignored because this is a top-level cat.

For ``is_empty(recursive=True) == True``, this is nested empty Batch but not Batch(), and we have to treat it as length zero and keep it.

tianshou/data/batch.py
youkaichao: > That's not right. It should be possible to do this:

I don't get it. It is already supported now, ``Batch.stack([Batch(a=np.array([1])), Batch(a=np.array([2]), b=np.array([3]))])`` works as expected.

duburcqa: So the doc is wrong. Discussed in the related issue.

tianshou/data/batch.py
duburcqa: OK nice.

duburcqa: Maybe add a small comment at the end of those lines to make it clear.

tianshou/data/batch.py
```diff
@@ -501,50 +545,97 @@ def to_torch(self, dtype: Optional[torch.dtype] = None,
             elif isinstance(v, Batch):
                 v.to_torch(dtype, device)
 
-    def cat_(self,
-             batches: Union['Batch', List[Union[dict, 'Batch']]]) -> None:
+    def __cat(self,
```
duburcqa, line 545: - take advantage of the new code architecture (cad combining `cat_` and `__cat`) to make `lens` argument required and not optional
- move the computation of the initial value of `lens` in `cat_` itself. 

It will make the complete implementation clearer and simpler.

tianshou/data/batch.py
youkaichao: Solved in #139 .

duburcqa: Mark resolved after fixing doc.

tianshou/data/batch.py
youkaichao: Fixed in 6287326 .

tianshou/data/batch.py
youkaichao: The doc is this [comment](https://github.com/thu-ml/tianshou/issues/139#issuecomment-659092494) is already fixed.

tianshou/data/batch.py
```diff
@@ -639,7 +601,40 @@ def __cat(self,
 
     def cat_(self,
              batches: Union['Batch', List[Union[dict, 'Batch']]]) -> None:
-        return self.__cat(batches)
+        """Concatenate a list of (or one) :class:`~tianshou.data.Batch` objects
+        into current batch.
+
+        If x=Batch(a=Batch(a=np.random.randn(3, 4)), b=np.random.randn(3, 4)),
+        y = Batch(a=Batch(a=Batch()), b=np.random.randn(3, 4)), if we want to
+        concatenate x and y, we want to pad y.a.a with zeros. Without ``lens``
```
duburcqa: `without ``lens``
        as a hint, when we concatenate x.a and y.a, we would not be able to
        know how to pad y.a. So we first compute the variable ``lens`` to the
        internal method ``Batch.__cat`` a hint.`

Not relevant here.

tianshou/data/batch.py
```diff
@@ -639,7 +601,40 @@ def __cat(self,
 
     def cat_(self,
              batches: Union['Batch', List[Union[dict, 'Batch']]]) -> None:
-        return self.__cat(batches)
+        """Concatenate a list of (or one) :class:`~tianshou.data.Batch` objects
+        into current batch.
+
+        If x=Batch(a=Batch(a=np.random.randn(3, 4)), b=np.random.randn(3, 4)),
+        y = Batch(a=Batch(a=Batch()), b=np.random.randn(3, 4)), if we want to
```
duburcqa: I don'k like much this example since it implies that `a=Batch(a=Batch())` has to be specified. Perhaps add an extra field to x: `x=Batch(a=Batch(a=np.random.randn(3, 4)), b=np.random.randn(3, 4), c=np.random.randn(3, 2))`

tianshou/data/batch.py
youkaichao: Indeed this should not appear in the doc string of ``cat_``, which can be seen by users. Moved to the doc string of ``__cat`` in f64faf5 .

tianshou/data/batch.py
duburcqa: Sure, but it has to be fixed in this PR, line 137.

tianshou/data/batch.py
youkaichao: I think it is okay. It is a corner case that requires some thinking. 

``x=Batch(a=Batch(a=np.random.randn(3, 4)), b=np.random.randn(3, 4), c=np.random.randn(3, 2))`` is straightforward and actually previous code works for it, by treating 'c' as partial keys.

So we had better keep it to let future maintainers (including us) understand the necessity of ``batch_holder.__cat(v, lens=lens)``.

duburcqa: As you wish.

tianshou/data/batch.py
youkaichao: Oh, my bad. Fixed in b795493 .

youkaichao:
> Torch is using `recurse` not `recursive`.

Fixed in 02f9d1a .

duburcqa (dismissed):

duburcqa:
Nice job ! Very high quality PR again :+1: @Trinkle23897 Ready to be merged.

NB: @Trinkle23897 you should not allow every collaborators to merge PR, or clearly state if you agree if I merge some PRs.

Trinkle23897:
> NB: @Trinkle23897 you should not allow every collaborator to merge PR, or clearly state if you agree if I merge some PRs.

2020-07-16 16-34-37 的屏幕截图

I don't know how to set it since the minimum role for approval is "write". I open the " At least 1 approving review by reviewers with write access." choice. Maybe set it to 2 is better for our three?

duburcqa:
> I don't know how to set it since the minimum role for approval is "write".

Settings -> branches -> branch protected rules -> master -> proected matching branches -> Require review from Code Owners + Restrict who can dismiss pull request reviews 

> Maybe set it to 2 is better for our three?

Why not

tianshou/data/batch.py
```diff
@@ -501,50 +544,97 @@ def to_torch(self, dtype: Optional[torch.dtype] = None,
...
+    def __cat(self,
+              batches: Union['Batch', List[Union[dict, 'Batch']]],
+              lens: List[int]) -> None:
+        """
+        If x=Batch(a=Batch(a=np.random.randn(3, 4)), b=np.random.randn(3, 4)),
+        y = Batch(a=Batch(a=Batch()), b=np.random.randn(3, 4)), if we want to
+        concatenate x and y, we want to pad y.a.a with zeros. Without ``lens``
+        as a hint, when we concatenate x.a and y.a, we would not be able to
+        know how to pad y.a. So we need ``Batch.cat_`` to compute the ``lens``
+        to give ``Batch.__cat`` a hint.
```
Trinkle23897:
```python
In [27]: b=Batch(a=np.zeros([3,4]),b=np.zeros([5,4]),c={})

In [28]: c=Batch(a=np.zeros([4,4]),b=np.zeros([5,4]),c=np.zeros([1,4]))

In [29]: Batch.cat([b,c]).c.shape
Out[29]: (4, 4)
```
Please specify this undefined behavior in the documentation or unittest. (choose the smallest size)

tianshou/data/batch.py
```diff
@@ -691,26 +798,53 @@ def __len__(self) -> int:
...
+        >>>Batch().is_empty()
+        True
+        >>>Batch(a=Batch(), b=Batch(c=Batch())).is_empty()
+        False
+        >>>Batch(a=Batch(), b=Batch(c=Batch())).is_empty(recurse=True)
+        True
+        >>>Batch(d=1).is_empty()
+        False
+        >>>Batch(a=np.float64(1.0)).is_empty()
+        False
```
Trinkle23897: Should add 4 space before `>>>` and another space after `>>>`:
```python
    >>> Batch().is_empty()
    True
```
I think it is good to render the documentation and take a look before the commit.

test/base/test_batch.py
```diff
@@ -109,6 +114,11 @@ def test_batch():
     assert isinstance(batch5.b, Batch)
     assert np.allclose(batch5.b.index, [1])
 
+    # None is a valid object and can be stored in Batch
+    a = Batch.stack([Batch(a=None), Batch(b=None)])
+    assert a.a[0] is None and a.a[1] is None
+    assert a.b[0] is None and a.b[1] is None
+
```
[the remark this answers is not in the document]
youkaichao: ``set`` object is a normal object, and this is expected. In fact, this feature is used in multi-agent RL to store legal actions (store list of sets as np.ndarray with np.object type).

tianshou/data/batch.py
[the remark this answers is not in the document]
youkaichao: ``Batch.cat_`` is an example.

youkaichao: This should go into the doc string of ``len``.

tianshou/data/batch.py
youkaichao: Maybe we can add the doc render as a test, too, just like pep8 check.

Trinkle23897: Agree

tianshou/data/batch.py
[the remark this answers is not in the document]
youkaichao: Examples added in doc string.

Trinkle23897 (approved):

Trinkle23897:
@duburcqa I've changed the number to 2.

duburcqa (approved):

PR #142, opened 2020-07-16
write tutorials to specify the standard of Batch

### LINKED ISSUE #138: The length of empty Batch

youkaichao:
1. write tutorials to specify the standard of Batch
2. bugfix for shape of Batch(a=1), shape property do not raise exception
3. type conversions are now consistent for all setters (constructor, __setattr__, and __setitem__)
4. scalars are converted to numpy scalars automatically


docs/tutorials/batch.rst
```diff
@@ -0,0 +1,449 @@
...
+    False
+    >>> Batch(a=np.float64(1.0)).is_empty()
+    False
+
+.. note::
+
+    Do not get confused with ``Batch.is_empty`` and ``Batch.empty``. ``Batch.empty`` and its in-place variant ``Batch.empty_`` are used to set some values to zeros. Check the api documentation for further details.
+
+Length and Shape
+-----------------
```
Trinkle23897: Delete a ‘-‘

docs/tutorials/batch.rst
```diff
@@ -0,0 +1,449 @@
...
+The above definition precisely defines the structure of the result of stack/concatenate batches. The values are relatively easy to define: for any key chain ``k`` that is applicable to ``b``, ``b[k]`` is the stack/concatenate of ``[bi[k] for bi in S]`` (if ``k`` is not applicable to ``bi``, appropriate size of zeros or ``None`` are filled automatically). If ``bi[k]`` are all ``Batch()``, then the aggregation result is also an empty ``Batch()``.
+
+Conceptually, how to aggregate batches is well done. And it is enough to understand the behavior of ``Batch`` objects during aggregation. Implementation is another story, though. Fortunately, Tianshou users do not have to worry about it. Just have the conceptual image in mind and you are all set!
+
+.. note::
+
+    ``Batch.stack`` and ``Batch.stack_`` also support ``axis`` argument so that one can stack batches besides the first dimension. But be cautious, if there are keys that are not shared across all batches, ``stack`` with ``axis != 0`` is undefined, and will cause an exception.
+
+Miscellaneous Notes
+--------------------
```
Trinkle23897: Del a ‘-‘

tianshou/data/batch.py
```diff
@@ -728,23 +529,7 @@ def stack_(self,
...
-            >>> c.b.shape
-            (2, 4, 6)
-            >>> c.common.c.shape
-            (2, 4, 5)
-
-        .. note::
-
-            If there are keys that are not shared across all batches, ``stack``
-            with ``axis != 0`` is undefined, and will cause an exception.
+        batches that do not have these keys will be padded by zeros.
```
Trinkle23897: I think this doc can maintain here instead of delete. Same of cat and empty

youkaichao: batch.py is already too long. We'd better keep it as short as possible. Given that we have a tutorial for Batch, I assume most users will learn that tutorial. So no need to duplicate the doc twice.

docs/tutorials/batch.rst
```diff
@@ -0,0 +1,449 @@
...
+             'sensory':sensory}}
+
+The real problem is how to **manipulate them**, like merging multiple hierarchical named tensors, adding two hierarchical named tensors, especially when some tensors are missing. ``Batch`` is designed to easily create, store, and manipulate these hierarchical named tensors.
+
+What Does Batch Contain
+------------------------
+
+The content of ``Batch`` objects can be defined by `regular expressions <https://en.wikipedia.org/wiki/Regular_expression>`_ as the following rules.
+
+.. code-block:: bash
```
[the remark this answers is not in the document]
youkaichao: I just want to create a shadow block without grammar highlight. This is a workaround. Happy to change it if there is any alternative.

youkaichao:
@duburcqa  ready for review now, it is done. I think the explanation of Aggregation of Heterogeneous Batches is crystally clear now :)

docs/tutorials/batch.rst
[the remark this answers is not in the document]
youkaichao: Solved. Plain indentation works just fine.

tianshou/data/batch.py
```diff
@@ -29,6 +29,30 @@ def _is_batch_set(data: Any) -> bool:
...
 
 
+def is_scalar_value(value: Any) -> bool:
+    # check if the value is a scalar
+    # 1. python bool object, number object: isinstance(value, Number)
+    # 2. numpy scalar: isinstance(value, np.generic)
+    # 3. python object rather than dict / Batch / tensor
+    # the check of dict / Batch is omitted because this only checks a value.
+    # a dict / Batch will eventually check their values
+    is_scalar = isinstance(value, Number)
```
duburcqa: Take advantage of short circuiting.

tianshou/data/batch.py
```diff
@@ -29,6 +29,30 @@ def _is_batch_set(data: Any) -> bool:
...
+    is_scalar = is_scalar or (isinstance(value, (np.ndarray, torch.Tensor))
+                              and not value.shape)
+    return is_scalar
+
+
+def is_number_value(value: Any) -> bool:
+    # isinstance(value, Number) checks 1, 1.0, np.int(1), np.float(1.0), etc.
+    # isinstance(value, np.nummber) checks np.int32(1), np.float64(1.0), etc.
+    # isinstance(value, np.bool_) checks np.bool_(True), etc.
+    is_scalar = isinstance(value, Number)
```
duburcqa: Take advantage of short circuiting.

tianshou/data/batch.py
```diff
@@ -29,6 +29,30 @@ def _is_batch_set(data: Any) -> bool:
...
+    # the check of dict / Batch is omitted because this only checks a value.
+    # a dict / Batch will eventually check their values
+    is_scalar = isinstance(value, Number)
+    is_scalar = is_scalar or isinstance(value, np.generic)
+    is_scalar = is_scalar or (isinstance(value, (np.ndarray, torch.Tensor))
+                              and not value.shape)
+    return is_scalar
+
+
+def is_number_value(value: Any) -> bool:
```
duburcqa: same

tianshou/data/batch.py
```diff
@@ -29,6 +29,30 @@ def _is_batch_set(data: Any) -> bool:
     return False
 
 
+def is_scalar_value(value: Any) -> bool:
```
duburcqa: I would rather emphasize that it is an internal method prefixing with `_`.

duburcqa: And I not to simply call it `is_scalar` ?

tianshou/data/batch.py
duburcqa: And I not to simply call it `is_number` ?

tianshou/data/batch.py
```diff
@@ -29,6 +29,30 @@ def _is_batch_set(data: Any) -> bool:
...
+    # check if the value is a scalar
+    # 1. python bool object, number object: isinstance(value, Number)
+    # 2. numpy scalar: isinstance(value, np.generic)
+    # 3. python object rather than dict / Batch / tensor
+    # the check of dict / Batch is omitted because this only checks a value.
+    # a dict / Batch will eventually check their values
+    is_scalar = isinstance(value, Number)
+    is_scalar = is_scalar or isinstance(value, np.generic)
```
duburcqa: why not to use np.asanyarray(value).size == 1 instead of anything else ? Is it really slower ?

tianshou/data/batch.py
duburcqa: I tend to agree with @Trinkle23897 (but not fully): Tutorials are not meant to be used as documentation. BUT documentation is not meant to be used as tutorial. So, try to find a good compromise. Only the required information to understand what each methods do and how they work must be available in module file, but those information have to be present (and one or two example never harms anyone, but not more).

tianshou/data/batch.py
youkaichao: short circuiting is okay. I write this because the line is too long, so I break it into multiple lines. Efficiency should not be a issue, though. Short circuiting still works.

tianshou/data/batch.py
```diff
@@ -859,6 +654,9 @@ def shape(self) -> List[int]:
                 try:
                     data_shape.append(v.shape)
                 except AttributeError:
+                    if is_scalar_value(v):
+                        data_shape.append([])
+                        continue
                     raise TypeError("No support for 'shape' method with "
```
duburcqa: Is it really possible to reach this point ? 

Anyway, I would rather ALWAYS return `data_shape.append([])` for any exception, instead of raising an exception. Because I don't think `shape` should raise an exception under any condition:
`Batch(a=1.0).shape` must not fail, neither `Batch(a={1.0, 2.0}).shape` does.

tianshou/data/batch.py
youkaichao: Fixed in 44034b6 .

tianshou/data/batch.py
duburcqa: Indeed. Ok then.

tianshou/data/batch.py
youkaichao: Fixed in 94f69b1 .

tianshou/data/batch.py
youkaichao: > I don't think shape should raise an exception under any condition:

That sounds reasonable. Even numpy scalars have shapes.

```python
>>> np.float64(1.0).shape
()
```

So yes, I agree with this change. What about @Trinkle23897 ?

tianshou/data/batch.py
youkaichao: > why not to use np.asanyarray(value).size == 1 

That's a cool solution! Haven't thought about that yet. It seems we can bypass many type-checks with this trick! ``np.asanyarray`` can do a lot for us.

Trinkle23897: Btw, asanyarray is mentioned in reward scalar

tianshou/data/batch.py
Trinkle23897: Agree

tianshou/data/batch.py
Trinkle23897: Agree with @duburcqa. You can make the doc shorter but please don’t erase the examples at all. It is easier for users to look at instead of an extra click to the Batch tutorial.

tianshou/data/batch.py
youkaichao: After a second thought, I still vote for the current implementation.

If we use ``np.asanyarray(value).size == 1``, additionally we have to check ``not np.asanyarray(value).shape`` to avoid ``[[1]]``

Furthermore, the current implementation makes it clear what types of data are considered scalars. But ``np.asanyarray(value).size == 1`` leaves all these to numpy and may confuse future maintainers.

Third, ``np.asanyarray`` does not come for free, though. It must actually does these type check, too.

tianshou/data/batch.py
youkaichao: Fixed in 6ec46c1 .

tianshou/data/batch.py
youkaichao: Fixed in 17d907a .

youkaichao:
Modifications to batch.py is a byproduct of this PR. The focus is to write that tutorial. So please take a look at batch.rst file if possible 💗 @Trinkle23897  @duburcqa 

If you do not have time to compile the tutorial, I have prepared a pdf version here :)

[Understand Batch — Tianshou 0.2.4 documentation.pdf](https://github.com/thu-ml/tianshou/files/4940894/Understand.Batch.Tianshou.0.2.4.documentation.pdf)

docs/tutorials/batch.rst
```diff
@@ -0,0 +1,474 @@
...
+
+    2. Type consistency: If ``bi[k]`` is not ``Batch()`` (the last key in the key chain is not a reserved key), then the type of ``b[k]`` should be the same as ``bi[k]``.
+
+The key chain applicability rises from the motivation of reserved keys. The type consistency requirement rises from the fact that, if we have a scalar/tensor value, that position in the aggregated ``Batch`` object should also be a scalar/tensor.
+
+If there exists ``b`` that satisfies these rules, it is clear that adding more reserved keys into ``b`` will not break these rules and there will be infinitely many ``b`` that can satisfy these rules. Among them, there will be an object with the least number of keys, and that is the answer to aggregating :math:`S`.
+
+The above definition precisely defines the structure of the result of stacking/concatenating batches. The values are relatively easy to define: for any key chain ``k`` that applies to ``b``, ``b[k]`` is the stack/concatenation of ``[bi[k] for bi in S]`` (if ``k`` does not apply to ``bi``, the appropriate size of zeros or ``None`` are filled automatically). If ``bi[k]`` are all ``Batch()``, then the aggregation result is also an empty ``Batch()``.
```
Trinkle23897: Too long. I got lost after reading it. Can you make it more simple?

tianshou/data/batch.py
duburcqa: > If we use np.asanyarray(value).size == 1, additionally we have to check not np.asanyarray(value).shape to avoid [[1]]

Indeed, that's the point of this approach, containing a single value, same goes for any array with shape [1, 1, ...., 1] x)

> Furthermore, the current implementation makes it clear what types of data are considered scalars. But np.asanyarray(value).size == 1 leaves all these to numpy and may confuse future maintainers.

I don't think so. It just means: it must be possible to cast this think in a numpy array having a single element. It is pretty clear to me, knowing that it is in the name of the function itself and the docstring.

> Third, np.asanyarray does not come for free, though. It must actually does these type check, too.

Of course, there is no free lunch.

duburcqa:
> Modifications to batch.py is a byproduct of this PR. The focus is to write that tutorial. So please take a look at batch.rst file if possible heartpulse @Trinkle23897 @duburcqa

I'm afraid that maybe it is good, but it is way too long to read it for now. 5 or 6 pages is the maximum. No way someone new with the lib will spend so much time reading this whereas it is NOT relevant for most user. Basic user does not care how Batch works, only the learning algorithms.

- Start with pictures and short statement as must as possible, and the pictures are too big.
- Use the minimum number of examples, and introduce them sometime before the explanations to draw the attention.
- Remove everything that is not necessary to understand how it works and what it does. There are many "note" that you could remove. Find another place to put such details or simply provide an example.
- Delete things that I don't necessary to understand how it works, such as length and shape, and maybe aggregation.

youkaichao:
> Delete things that I don't necessary to understand how it works, such as length and shape, and maybe aggregation.

Agree, indeed the tutorial is too long. I have divided the tutorial into two parts, basic and advanced. Length, shape, and aggregation can go to the advanced part. Now the tutorial has 9 pages: 4 pages for basic usages and 5 pages for advanced topics. Code snippets are by default hidden and clickable when users want to read.
[Understand Batch — Tianshou 0.2.4 documentation.pdf](https://github.com/thu-ml/tianshou/files/4941571/Understand.Batch.Tianshou.0.2.4.documentation.pdf)

tianshou/data/batch.py
youkaichao: If we use ``np.asanyarray(value).size`` to check whether it is a scalar, maybe we can take a step further, and just store the data with np.ndarray. This can remove many unnecessary type checks.

Current status:
```python
a = Batch(a=1)
type(a.a) # int
```

Proposed:
```python
a = Batch(a=1)
type(a.a) # np.ndarray
print(a.a) # array(1)
```

duburcqa: > If we use np.asanyarray(value).size to check whether it is a scalar

I'm not saying we must do it though, just that it must be considered.

> maybe we can take a step further, and just store the data with np.ndarray. This can remove many unnecessary type checks.

Personally I don't see any reason not to convert scalars into numpy array systematically if it is actually un-noticeable for the user. Remove unnecessary type check if a good thing.

duburcqa:
It is much better, this time is read it. Here are my comments:
[Understand.Batch.Tianshou.0.2.4.documentation-duburcqa-comments.pdf](https://github.com/thu-ml/tianshou/files/4941643/Understand.Batch.Tianshou.0.2.4.documentation-1.pdf)

The common thread: go STRAIGHT TO THE POINT! Avoid saying obvious things at all cost.

angtft


But **nice job overall**. It is pretty clear and well written :+1:

Trinkle23897:
@duburcqa The file link is broken.

duburcqa:
> @duburcqa The file link is broken.

Fixed.

youkaichao:
> Avoid saying obvious things at all cost.

It is arguable what is obvious, though. It varies with readers. I prefer to explain as clear as I can. If some readers find it obvious, they can skip some content and just skim the whole tutorial.

Trinkle23897:
> It is arguable what is obvious, though. It varies with readers. I prefer to explain as clear as I can. If some readers find it obvious, they can skip some content and just skim the whole tutorial.

Let me define "obvious": a beginner of RL can quick understand the material.

duburcqa:
> It is arguable what is obvious, though. It varies with readers. 

Indeed, and I'm not saying that I know exactly what is obvious or not. Just to be very cautious about that.

> I prefer to explain as clear as I can. If some readers find it obvious, they can skip some content and just skim the whole tutorial.

I don't share this opinion. Personally I don't read anything in the first place if it is too long. I prefer something shorter not going through the details, so that I can quickly get the whole picture without getting lost in information.

youkaichao:
7.5 pages in 316914b : 3 pages for basic and 4.5 pages for advanced. Most advice is taken.  Aggregation of Heterogeneous Batches is kept because I do not know how to be more concise :(

[Understand Batch — Tianshou 0.2.4 documentation.pdf](https://github.com/thu-ml/tianshou/files/4941800/Understand.Batch.Tianshou.0.2.4.documentation.pdf)

duburcqa:
Updated comments: [Understand.Batch.Tianshou.0.2.4.documentation.pdf](https://github.com/thu-ml/tianshou/files/4941816/Understand.Batch.Tianshou.0.2.4.documentation.pdf)

youkaichao:
8 pages now: 3 for basic and 5 for advanced. The layout in pdf is malformed, but it is ok in html pages.

Move figure to the top; Trim the description of aggregation; add a figure for aggregation.

[Understand Batch — Tianshou 0.2.4 documentation.pdf](https://github.com/thu-ml/tianshou/files/4941907/Understand.Batch.Tianshou.0.2.4.documentation.pdf)

youkaichao:
I find many inconsistent type conversions and delegate all of them to ``Batch.__init__`` to both eliminate duplicate code and keep type consistency in 67068dc and b000381 .

tianshou/data/batch.py
```diff
@@ -308,47 +132,35 @@ def __init__(self,
...
-            else:
-                value = np.array(value)
-                if not issubclass(value.dtype.type, (np.bool_, np.number)):
-                    value = value.astype(np.object)
-        elif isinstance(value, dict) or isinstance(value, np.ndarray) \
-                and value.dtype == np.object and _is_batch_set(value):
-            value = Batch(value)
-        self.__dict__[key] = value
+        """self.key = value"""
+        another = Batch({key: value})
```
duburcqa: nice !

tianshou/data/batch.py
```diff
@@ -389,20 +201,13 @@ def __setitem__(self, index: Union[
...
-            self.__dict__[index] = value
+            self.__setattr__(index, value)
             return
-        if not isinstance(value, (dict, Batch)):
-            if _is_batch_set(value):
-                value = Batch(value)
-            else:
-                raise TypeError("Batch does not supported value type "
-                                f"{type(value)} for item assignment.")
+        value = Batch(a=value).a
```
duburcqa: There is no better way to do it ? it is not very elegant to create a unless field a and to extract it right after. Maybe it would be nice to create a new internal method use to cast value in Batch-compatible data type, that would be called both here and in `__init__`.

tianshou/data/batch.py
```diff
@@ -389,20 +201,13 @@ def __setitem__(self, index: Union[
...
+        if isinstance(value, (np.ndarray, torch.Tensor)):
```
duburcqa, line 210: it this really necessary to be that much restrictive ? I don't like the idea to add such a constraint for advanced users.

youkaichao: I think setters of Batch should be consistent. For advanced users, they can directly manipulate ``self.__dict__``.

youkaichao: > I don't like the idea to add such a constraint for advanced users.

This constraint is put on indexing with slice or int, not string. ``obj['a']=tensor`` still works, but ``obj[1] = tensor`` is really undefined, because ``obj[1]`` always returns a Batch object.

tianshou/data/batch.py
youkaichao: Fixed in c7ed637 .

youkaichao:
Any further comments for thhe tutorial?

Trinkle23897:
> Any further comments for thhe tutorial?

New version, please have a look

[Understand Batch — Tianshou 0.2.4 documentation.pdf](https://github.com/thu-ml/tianshou/files/4943140/Understand.Batch.Tianshou.0.2.4.documentation.pdf)

I have no other comments

duburcqa (approved):

tianshou/data/batch.py
```diff
@@ -29,6 +29,43 @@ def _is_batch_set(data: Any) -> bool:
     return False
 
 
+def _is_scalar(value: Any) -> bool:
+    # check if the value is a scalar
+    # 1. python bool object, number object: isinstance(value, Number)
```
duburcqa, line 34: Does this comment still make sense ?

tianshou/data/batch.py
```diff
@@ -29,6 +29,43 @@ def _is_batch_set(data: Any) -> bool:
...
+    # 2. numpy scalar: isinstance(value, np.generic)
+    # 3. python object rather than dict / Batch / tensor
+    # the check of dict / Batch is omitted because this only checks a value.
+    # a dict / Batch will eventually check their values
+    value = np.asanyarray(value)
+    return value.size == 1 and not value.shape
```
duburcqa, line 40: Maybe your previous implementation was more appropriate after all, if you need to specify `and not value.shape`. I don't know.

tianshou/data/batch.py
youkaichao: The comment conceptually descibes what is a scalar, it can differ from the implementation details. I think it is ok.

Trinkle23897 (approved):

PR #147, opened 2020-07-18
Improve to_torch/to_numpy converters

### LINKED ISSUE #152: cuda tensor will raise an exception in np.asanyarray
```python
In [2]: a=Batch(a={})

In [3]: b=Batch(a=torch.tensor([3,4], device='cuda'))

In [4]: Batch.cat([a,b])
```
will raise an exception. Same as Batch.stack:
```python
~/github/tianshou-new/tianshou/data/batch.py in _is_scalar(value)
     37     # the check of dict / Batch is omitted because this only checks a value.
     38     # a dict / Batch will eventually check their values
---> 39     value = np.asanyarray(value)
     40     return value.size == 1 and not value.shape
     41 

~/.local/lib/python3.6/site-packages/numpy/core/_asarray.py in asanyarray(a, dtype, order)
    136 
    137     """
--> 138     return array(a, dtype, copy=False, order=order, subok=True)
    139 
    140 

~/.local/lib/python3.6/site-packages/torch/tensor.py in __array__(self, dtype)
    490     def __array__(self, dtype=None):
    491         if dtype is None:
--> 492             return self.numpy()
    493         else:
    494             return self.numpy().astype(dtype, copy=False)

```

duburcqa:
- Do not force cast `list` and `tuple`.
- Do not silently skip cast if no converter is found
- Add fallback for `to_numpy`

### DIFF
--- a/test/base/test_batch.py
+++ b/test/base/test_batch.py
@@ -35,6 +35,10 @@ def test_batch():
         Batch(a=[np.zeros((3, 2)), np.zeros((3, 3))])
     with pytest.raises(TypeError):
         Batch(a=[torch.zeros((2, 3)), torch.zeros((3, 3))])
+    with pytest.raises(TypeError):
+        Batch(a=[torch.zeros(3,3), np.zeros([3,3])])
+    with pytest.raises(TypeError):
+        Batch(a=[1, np.zeros([3,3]), torch.zeros(3,3)])
     batch = Batch(a=[torch.ones(3), torch.ones(3)])
     assert torch.allclose(batch.a, torch.ones(2, 3))
     batch = Batch(obs=[0], np=np.zeros([3, 4]))
--- a/test/base/test_batch.py
+++ b/test/base/test_batch.py
@@ -41,6 +41,7 @@ def test_batch():
         Batch(a=[1, np.zeros((3, 3)), torch.zeros((3, 3))])
     batch = Batch(a=[torch.ones(3), torch.ones(3)])
     assert torch.allclose(batch.a, torch.ones(2, 3))
+    Batch(a=[])
     batch = Batch(obs=[0], np=np.zeros([3, 4]))
     assert batch.obs == batch["obs"]
     batch.obs = [1]
--- a/test/base/test_batch.py
+++ b/test/base/test_batch.py
@@ -36,9 +36,9 @@ def test_batch():
     with pytest.raises(TypeError):
         Batch(a=[torch.zeros((2, 3)), torch.zeros((3, 3))])
     with pytest.raises(TypeError):
-        Batch(a=[torch.zeros(3,3), np.zeros([3,3])])
+        Batch(a=[torch.zeros((3, 3)), np.zeros((3, 3))])
     with pytest.raises(TypeError):
-        Batch(a=[1, np.zeros([3,3]), torch.zeros(3,3)])
+        Batch(a=[1, np.zeros((3, 3)), torch.zeros((3, 3))])
     batch = Batch(a=[torch.ones(3), torch.ones(3)])
     assert torch.allclose(batch.a, torch.ones(2, 3))
     batch = Batch(obs=[0], np=np.zeros([3, 4]))
--- a/test/base/test_batch.py
+++ b/test/base/test_batch.py
@@ -3,6 +3,7 @@ import copy
 import pickle
 import pytest
 import numpy as np
+from itertools import starmap
 
 from tianshou.data import Batch, to_torch
 
@@ -307,7 +308,7 @@ def test_batch_over_batch_to_torch():
     assert batch.b.d.dtype == torch.float32
 
 
-def test_utils_to_torch():
+def test_utils_to_torch_numpy():
     batch = Batch(
         a=np.float64(1.0),
         b=Batch(
@@ -326,25 +327,32 @@ def test_utils_to_torch():
     data_list = [float('nan'), 1]
     data_list_torch = to_torch(data_list)
     assert data_list_torch.dtype == torch.float64
-    data_list_2 = [np.zeros((3, 3)), np.zeros((3, 3))]
+    data_list_2 = [np.random.rand(3, 3), np.random.rand(3, 3)]
     data_list_2_torch = to_torch(data_list_2)
     assert data_list_2_torch.shape == (2, 3, 3)
+    assert np.allclose(to_numpy(to_torch(data_list_2)), data_list_2)
     data_list_3 = [np.zeros((3, 2)), np.zeros((3, 3))]
     data_list_3_torch = to_torch(data_list_3)
     assert isinstance(data_list_3_torch, list)
-    assert isinstance(data_list_3_torch[0], torch.Tensor)
+    assert all(isinstance(e, torch.Tensor) for e in data_list_3_torch)
+    assert all(starmap(np.allclose,
+               zip(to_numpy(to_torch(data_list_3)), data_list_3)))
     data_list_4 = [np.zeros((2, 3)), np.zeros((3, 3))]
     data_list_4_torch = to_torch(data_list_4)
     assert isinstance(data_list_4_torch, list)
-    assert isinstance(data_list_4_torch[0], torch.Tensor)
+    assert all(isinstance(e, torch.Tensor) for e in data_list_4_torch)
+    assert all(starmap(np.allclose,
+               zip(to_numpy(to_torch(data_list_4)), data_list_4)))
     data_list_5 = [np.zeros(2), np.zeros((3, 3))]
     data_list_5_torch = to_torch(data_list_5)
     assert isinstance(data_list_5_torch, list)
-    assert isinstance(data_list_5_torch[0], torch.Tensor)
+    assert all(isinstance(e, torch.Tensor) for e in data_list_5_torch)
     data_array = np.zeros((3, 2, 2))
     data_tensor = to_torch(data_array[[]])
     assert isinstance(data_tensor, torch.Tensor)
     assert data_tensor.shape == (0, 2, 2)
+    assert np.allclose(to_numpy(to_torch(data_array)), data_array)
+
 
 def test_batch_pickle():
     batch = Batch(obs=Batch(a=0.0, c=torch.Tensor([1.0, 2.0])),
@@ -451,7 +459,7 @@ if __name__ == '__main__':
     test_batch()
     test_batch_over_batch()
     test_batch_over_batch_to_torch()
-    test_utils_to_torch()
+    test_utils_to_torch_numpy()
     test_batch_pickle()
     test_batch_from_to_numpy_without_copy()
     test_batch_standard_compatibility()
--- a/test/base/test_batch.py
+++ b/test/base/test_batch.py
@@ -333,10 +333,14 @@ def test_utils_to_torch():
     data_list_3_torch = to_torch(data_list_3)
     assert isinstance(data_list_3_torch, list)
     assert isinstance(data_list_3_torch[0], torch.Tensor)
-    data_list_4 = [np.zeros(2), np.zeros((3, 3))]
+    data_list_4 = [np.zeros((2, 3)), np.zeros((3, 3))]
     data_list_4_torch = to_torch(data_list_4)
     assert isinstance(data_list_4_torch, list)
     assert isinstance(data_list_4_torch[0], torch.Tensor)
+    data_list_5 = [np.zeros(2), np.zeros((3, 3))]
+    data_list_5_torch = to_torch(data_list_5)
+    assert isinstance(data_list_5_torch, list)
+    assert isinstance(data_list_5_torch[0], torch.Tensor)
     data_array = np.zeros((3, 2, 2))
     data_tensor = to_torch(data_array[[]])
     assert isinstance(data_tensor, torch.Tensor)
--- a/test/base/test_batch.py
+++ b/test/base/test_batch.py
@@ -29,6 +29,10 @@ def test_batch():
     assert b.a == 3
     with pytest.raises(AssertionError):
         Batch({1: 2})
+    with pytest.raises(TypeError):
+        Batch(a=[np.zeros((2, 3)), np.zeros((3, 3))])
+    with pytest.raises(TypeError):
+        Batch(a=[torch.zeros((2, 3)), torch.zeros((3, 3))])
     batch = Batch(a=[torch.ones(3), torch.ones(3)])
     assert torch.allclose(batch.a, torch.ones(2, 3))
     batch = Batch(obs=[0], np=np.zeros([3, 4]))
--- a/test/base/test_batch.py
+++ b/test/base/test_batch.py
@@ -32,6 +32,8 @@ def test_batch():
     with pytest.raises(TypeError):
         Batch(a=[np.zeros((2, 3)), np.zeros((3, 3))])
     with pytest.raises(TypeError):
+        Batch(a=[np.zeros((3, 2)), np.zeros((3, 3))])
+    with pytest.raises(TypeError):
         Batch(a=[torch.zeros((2, 3)), torch.zeros((3, 3))])
     batch = Batch(a=[torch.ones(3), torch.ones(3)])
     assert torch.allclose(batch.a, torch.ones(2, 3))
--- a/test/base/test_batch.py
+++ b/test/base/test_batch.py
@@ -347,10 +347,13 @@ def test_utils_to_torch_numpy():
     data_list_5_torch = to_torch(data_list_5)
     assert isinstance(data_list_5_torch, list)
     assert all(isinstance(e, torch.Tensor) for e in data_list_5_torch)
-    data_array = np.zeros((3, 2, 2))
-    data_tensor = to_torch(data_array[[]])
-    assert isinstance(data_tensor, torch.Tensor)
-    assert data_tensor.shape == (0, 2, 2)
+    data_array = np.random.rand(3, 2, 2)
+    data_empty_tensor = to_torch(data_array[[]])
+    assert isinstance(data_empty_tensor, torch.Tensor)
+    assert data_empty_tensor.shape == (0, 2, 2)
+    data_empty_array = to_numpy(data_empty_tensor)
+    assert isinstance(data_empty_array, np.ndarray)
+    assert data_empty_array.shape == (0, 2, 2)
     assert np.allclose(to_numpy(to_torch(data_array)), data_array)
 
 
--- a/test/base/test_batch.py
+++ b/test/base/test_batch.py
@@ -323,8 +323,10 @@ def test_utils_to_torch():
     assert batch_torch_float.a.dtype == torch.float32
     assert batch_torch_float.b.c.dtype == torch.float32
     assert batch_torch_float.b.d.dtype == torch.float32
-    array_list = [float('nan'), 1.0]
-    assert to_torch(array_list).dtype == torch.float64
+    data_list = [float('nan'), 1]
+    data_list_torch = to_torch(data_list)
+    assert data_list_torch[0].dtype == torch.float64
+    assert data_list_torch[1].dtype == torch.int64
 
 
 def test_batch_pickle():
--- a/test/base/test_batch.py
+++ b/test/base/test_batch.py
@@ -325,9 +325,22 @@ def test_utils_to_torch():
     assert batch_torch_float.b.d.dtype == torch.float32
     data_list = [float('nan'), 1]
     data_list_torch = to_torch(data_list)
-    assert data_list_torch[0].dtype == torch.float64
-    assert data_list_torch[1].dtype == torch.int64
-
+    assert data_list_torch.dtype == torch.float64
+    data_list_2 = [np.zeros((3, 3)), np.zeros((3, 3))]
+    data_list_2_torch = to_torch(data_list_2)
+    assert data_list_2_torch.shape == (2, 3, 3)
+    data_list_3 = [np.zeros((3, 2)), np.zeros((3, 3))]
+    data_list_3_torch = to_torch(data_list_3)
+    assert isinstance(data_list_3_torch, list)
+    assert isinstance(data_list_3_torch[0], torch.Tensor)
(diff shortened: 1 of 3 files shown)


duburcqa:
Maybe a more clever mechanism could be implemented for `list`/`tuple`:

If all elements have the same type and shape (except for the first dim), than it can be converted as a whole into torch.tensor/np.ndarray, otherwise each element is handled separately.

What do you guys think ? @youkaichao @Trinkle23897

tianshou/data/utils.py
```diff
@@ -36,14 +40,16 @@ def to_torch(x: Union[torch.Tensor, dict, Batch, np.ndarray],
...
-    elif isinstance(x, np.ndarray) and \
-            isinstance(x.item(0), (np.number, np.bool_, Number)):
-        x = torch.from_numpy(x).to(device)
-        if dtype is not None:
-            x = x.type(dtype)
+    elif isinstance(x, (list, tuple)):
+        x = [to_torch(x_i, dtype, device) for x_i in x]
+    else:  # fallback
+        x = np.asanyarray(x)
+        if isinstance(x.item(0), (np.number, np.bool_, Number)):
```
youkaichao: The docstring in NumPy says:

> a.item(*args)
> Copy an element of an array to a standard Python scalar and return it.
> ...

It seems x.item(0) will not be instance of np.number or np.bool_

youkaichao:
> If all elements have the same type and shape (except for the first dim), than it can be converted as a whole into torch.tensor/np.ndarray, otherwise each element is handled separately.

I'm ok with this. If they finally get into a Batch, they will be converted to a whole array anyway. If not, it makes sense to handle them separately.

tianshou/data/utils.py
```diff
@@ -17,6 +17,10 @@ def to_numpy(x: Union[
             x[k] = to_numpy(v)
     elif isinstance(x, Batch):
         x.to_numpy()
+    elif isinstance(x, (list, tuple)):
```
youkaichao, line 20: Update type annotation, too.

Trinkle23897:
> > If all elements have the same type and shape (except for the first dim), than it can be converted as a whole into torch.tensor/np.ndarray, otherwise each element is handled separately.
> 
> I'm ok with this. If they finally get into a Batch, they will be converted to a whole array anyway. If not, it makes sense to handle them separately.

Agree with Kaichao

youkaichao:
I think the point is, previous code returns one merged tensor for ``to_torch([np.zeros((3, 4)), np.zeros((3, 4))])``. Now you want to return two separate tensors to avoid merging them. Right? @duburcqa

test/base/test_batch.py
```diff
@@ -322,8 +322,10 @@ def test_utils_to_torch():
     assert batch_torch_float.a.dtype == torch.float32
     assert batch_torch_float.b.c.dtype == torch.float32
     assert batch_torch_float.b.d.dtype == torch.float32
-    array_list = [float('nan'), 1.0]
-    assert to_torch(array_list).dtype == torch.float64
+    data_list = [float('nan'), 1]
+    data_list_torch = to_torch(data_list)
+    assert data_list_torch[0].dtype == torch.float64
```
youkaichao: Is this platform dependent? Or it surely returns data type with ``torch.float64``? The same for the following line ``torch.int64``.

duburcqa: It is not platform dependent. Default native Python and numpy types are using 64 bits, while torch is doing the same for integers, but using 32bits by default for floating points, on any platforms.

duburcqa:
> I think the point is, previous code returns one merged tensor for to_torch([np.zeros((3, 4)), np.zeros((3, 4))]). Now you want to return two separate tensors to avoid merging them. Right? @duburcqa

Exactly, it is a step forward supporting action space `gym.Space.Tuple`.

Trinkle23897:
A bug:
```python
In [3]: d=np.zeros([3,6,6])

In [4]: to_torch(d[[]])
~/github/tianshou-new/tianshou/data/utils.py in to_torch(x, dtype, device)
     41         x = to_torch(np.asanyarray(x), dtype, device)
     42     elif isinstance(x, np.ndarray) and \
---> 43             isinstance(x.item(0), (np.number, np.bool_, Number)):
     44         x = torch.from_numpy(x).to(device)
     45         if dtype is not None:

IndexError: index 0 is out of bounds for size 0
```
However, 
```python
In [6]: to_torch(Batch(d=d)[[]])
Out[6]: 
Batch(
    d: tensor([], size=(0, 6, 6), dtype=torch.float64),
)
```

So, do not use ``x.item(0)`` to find the type, use ``issubclass(x.dtype.type, (np.bool_, np.number))`` as in batch.py

duburcqa:
> A bug: [...]

Yes I noticed this. Thank you !

Trinkle23897:
> > A bug: [...]
> 
> Yes I noticed this. Thank you!

So could you please add this into the testcase?

duburcqa:
> So could you please add this into the testcase?

Sure ! on it

tianshou/data/utils.py
```diff
@@ -18,7 +18,11 @@ def to_numpy(x: Union[
     elif isinstance(x, Batch):
         x.to_numpy()
     elif isinstance(x, (list, tuple)):
-        x = [to_numpy(x_i) for x_i in x]
+        x = np.asanyarray(x)
+        if x.dtype == np.object:
+            x = [to_numpy(e) for e in x]
+        else:
+            x = to_numpy(x)
```
youkaichao: > List of object are converted individually, as a single tensor otherwise.

Why is that? You say you want to support tuple observation, and it needs ``x = [to_numpy(x_i) for x_i in x]``.

youkaichao: The following code fails immediately:
```python
a = np.asanyarray([torch.ones((3, 4)), torch.ones((3, 4))])
```

duburcqa:
Ready for review.

tianshou/data/batch.py
```diff
@@ -116,15 +115,17 @@ def _parse_value(v: Any):
...
-        v_ = _to_array_with_correct_type(v)
-        if v_.dtype == np.object and _is_batch_set(v):
-            v = Batch(v)  # list of dict / Batch
-        else:
-            # normal data list (main case)
-            # or actually a data list with objects
-            v = v_
+            return torch.stack(v)
+        try:
+            v_ = _to_array_with_correct_type(v)
```
duburcqa: Fixes a "bug" of `Batch`:

```python
>>> Batch(a=[np.zeros((3, 2)), np.zeros((3, 3))])
Batch(
    a: array([array([[0., 0.],
              [0., 0.],
              [0., 0.]]),
              array([[0., 0., 0.],
              [0., 0., 0.],
              [0., 0., 0.]])], dtype=object),
)
```

is now working now.

test/base/test_batch.py
```diff
@@ -323,9 +323,28 @@ def test_utils_to_torch():
...
+    assert isinstance(data_list_4_torch, list)
+    assert isinstance(data_list_4_torch[0], torch.Tensor)
+    data_list_5 = [np.zeros(2), np.zeros((3, 3))]
+    data_list_5_torch = to_torch(data_list_5)
+    assert isinstance(data_list_5_torch, list)
+    assert isinstance(data_list_5_torch[0], torch.Tensor)
+    data_array = np.zeros((3, 2, 2))
+    data_tensor = to_torch(data_array[[]])
+    assert isinstance(data_tensor, torch.Tensor)
+    assert data_tensor.shape == (0, 2, 2)
```
Trinkle23897: Also with `np.allclose(to_numpy(data_tensor), data_array)`

tianshou/data/batch.py
Trinkle23897: But how about `Batch(a=[torch.zeros((3, 2)), torch.zeros((3, 3))])`?

duburcqa: Fixed:
```python
>>> Batch(a=[torch.zeros((3, 2)), torch.zeros((3, 3))])
Batch(
    a: array([tensor([[0., 0.],
               [0., 0.],
               [0., 0.]]),
              tensor([[0., 0., 0.],
               [0., 0., 0.],
               [0., 0., 0.]])], dtype=object),
)
```

youkaichao:
Personally, I don't think this is the best way to support gym.Space.Tuple in either the observation space or the action space.

A nice workaround is to wrap the original environment to return dict state and accept dict action, which is natively supported by tianshou.data.Batch.

Take observation space as an example, if an environment returns a tuple observation with two items, one is image observation with shape [224, 224, 3], the other is a vector observation with shape [10]. 

By writing an environment wrapper, the observation is ``{'img':img, 'vec':vec}``. Everything else works without modifications.
```python
for i in range(100):
    buffer.add(obs={'img':np.zeros((224, 224, 3)), 'vec':np.zeros((10,))})
buffer.obs.img # tensor of shape [100, 224, 224, 3], ready to feed in neural network policy
buffer.obs.vec # tensor of shape [100, 10], ready to feed in neural network policy
```

But if we support tuple observation as this pr, the observation in the buffer is a 2-d object array. The policy has to unpack the observation array and construct separate tensors for image observation and vector observations.
```python
for i in range(100):
    buffer.add(obs=[np.zeros((224, 224, 3)), np.zeros((10,))])
buffer.obs # tensor of shape [100, 2], with data type np.object
img = np.stack(buffer.obs[:, 0]) # user has to do this explicitly!
vec = np.stack(buffer.obs[:, 1]) # user has to do this explicitly!
```

In addition, ``stack`` always create a new copy of data, while slicing does not. So the latter solution of directly supporting tuple observation keeps creating new data, costs much memory!

```python
a = np.zeros((3, 4))
a
Out[104]: 
array([[0., 0., 0., 0.],
       [0., 0., 0., 0.],
       [0., 0., 0., 0.]])
b = a[:, 0] # sliced objects share memory 
b[0] = 1
In[107]: 
a
Out[107]: 
array([[1., 0., 0., 0.],
       [0., 0., 0., 0.],
       [0., 0., 0., 0.]])

d = np.stack([a, a]) # stacking objects require additional memory cost
d
Out[109]: 
array([[[1., 0., 0., 0.],
        [0., 0., 0., 0.],
        [0., 0., 0., 0.]],
       [[1., 0., 0., 0.],
        [0., 0., 0., 0.],
        [0., 0., 0., 0.]]])
d[:] = 0
d
Out[111]: 
array([[[0., 0., 0., 0.],
        [0., 0., 0., 0.],
        [0., 0., 0., 0.]],
       [[0., 0., 0., 0.],
        [0., 0., 0., 0.],
        [0., 0., 0., 0.]]])
a
Out[112]: 
array([[1., 0., 0., 0.],
       [0., 0., 0., 0.],
       [0., 0., 0., 0.]])

```

duburcqa:
@youkaichao yes I know that. That's why I'm not planning to implement anything else regarding tuple support. Batch is obviously not design for that at this point.

As I said, it is just a step forward. Currently, it only consists in increasing the versatility of generic helpers, that are not indended to be used only internally in conjunction with Batch.

test/base/test_batch.py
Trinkle23897: I mean to test both to_numpy and to_torch that won't fail on this (0, 2, 2) case

duburcqa: Let me know if it is fine now.

Trinkle23897:
I don't know whether we should support this: 
```python
In [15]: d=Batch(a=[np.zeros([3,3]), torch.zeros([3,3])])

In [16]: d.to_torch() 
```

duburcqa:
> I don't know whether we should support this: [...]

Yes I noticed this. So I propose to completely forbidden arrays of tensors.

tianshou/data/batch.py
```diff
@@ -114,16 +113,24 @@ def _assert_type_keys(keys):
...
+                # normal data list (main case)
+                # or actually a data list with objects
+                v = v_
+        except (ValueError, RuntimeError):
+            v_ = np.empty(len(v), dtype=np.object)
+            for i, e in enumerate(v):
+                if not isinstance(e, torch.Tensor):
+                    v_[i] = _to_array_with_correct_type(e)
+                else:
+                    v_[i] = e
```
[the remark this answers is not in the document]
duburcqa: It should be clearer now.

duburcqa:
It should be ready now.

tianshou/data/batch.py
```diff
@@ -52,17 +52,19 @@ def _is_number(value: Any) -> bool:
 
 def _to_array_with_correct_type(v: Any) -> np.ndarray:
     # convert the value to np.ndarray
-    # convert to np.object data type if neither bool nor number
     v = np.asanyarray(v)
```
youkaichao, line 61: ``np.asanyarray`` fails in the presence of cuda tensor ``torch.ones(1, 2, device='cuda')``. Maybe this pr can fix it by the way? #152

tianshou/data/batch.py
```diff
@@ -114,17 +118,20 @@ def _assert_type_keys(keys):
...
+            v_ = _to_array_with_correct_type(v)
+            if v_.dtype == np.object and _is_batch_set(v):
+                v = Batch(v)  # list of dict / Batch
+            else:
+                # normal data list (main case)
+                # or actually a data list with objects
+                v = v_
+        except (ValueError, RuntimeError):
+            raise TypeError("Batch does not support non-stackable list/tuple "
+                            "of tensors as value yet.")
```
youkaichao: Somehow I'm not comfortable with such a large try-catch block. Maybe we can just leave the exceptions as they are. Such a large try-catch block may catch some unintended exceptions and mislead users.

tianshou/data/batch.py
duburcqa: Good point. But I checked the code of this PR and it should raise an appropriate exception in any case already.

Trinkle23897: But it should not raise an exception.

tianshou/data/batch.py
youkaichao: > Finally, it does nothing more than raising an exception, so the trace is preserved.

My point is that, not only the trace but also the exception message should remain, so that users know what causes the exception. For example, torch.cat gives message like "RuntimeError: stack expects each tensor to be equal size, but got [3, 4] at entry 0 and [3, 5] at entry 1", this message is valuable for debugging, but will be replaced by "Batch does not support non-stackable list/tuple of tensors as value yet." in this pr.

tianshou/data/batch.py
duburcqa: > But it should not raise an exception.

To be clearer: It should raise an appropriate exception **when appropriate**. :smile: 

For instance: 
```
Batch(a=[torch.tensor([1.0], device='cpu'), torch.tensor([2.0], device='cuda')])
```
raises an exception, since it is actually non-stackable.

youkaichao: > But it should not raise an exception.

Agree. ``_to_array_with_correct_type `` should deal with cuda tensors correctly, rather than raising an exception.

duburcqa: > Agree. _to_array_with_correct_type should deal with cuda tensors correctly, rather than raising an exception.

It is just wrong. `_to_array_with_correct_type` is never dealing with torch.Tensor objects in a relevant scenario (except for raising exceptions). So it makes no sense to do anything else at `_to_array_with_correct_type` level. It will be handle by `_parse_value` later anyway.

tianshou/data/batch.py
duburcqa: > "RuntimeError: stack expects each tensor to be equal size, but got [3, 4] at entry 0 and [3, 5] at entry 1", this message is valuable for debugging, but will be replaced by "Batch does not support non-stackable list/tuple of tensors as value yet." in this pr.

So it is basically the same but less specific. If you really want it I can forward the message to ease debugging.

duburcqa: > I can forward the message to ease debugging.

Done !

tianshou/data/batch.py
youkaichao: > _to_array_with_correct_type is never dealing with torch.Tensor objects in a relevant scenario (except for raising exceptions).

Ok, that's reasonable. tensors are handled by ``_parse_value``.

tianshou/data/batch.py
youkaichao: > I can forward the message to ease debugging.

With this, I'm ok now. You can merge as you like. Anyway, I still don't get the point of this big try-catch block.

duburcqa: Ok I will split it if it is clearer for you. No problem.

duburcqa: If we are to collaborate, we have to make sure everybody understands and is ok with each other's PR.

youkaichao: > Ok I will split it if it is clearer for you. No problem.

It looks much better if you just catch the ``v_ = _to_array_with_correct_type(v)`` statement or do it inside ``_to_array_with_correct_type``. It does not make sense to catch other statements, I think.

```python
        try:
            if not isinstance(v, np.ndarray) and \
                    all(isinstance(e, torch.Tensor) for e in v):
                return torch.stack(v) # no need to catch here, users clearly know what caused the error when an exception happens here
            v_ = _to_array_with_correct_type(v)
            if v_.dtype == np.object and _is_batch_set(v):
                # no need to catch here, users clearly know what caused the error when an exception happens here
                v = Batch(v)  # list of dict / Batch
            else:
                # normal data list (main case)
                # or actually a data list with objects
                v = v_ # no need to catch here, users clearly know what caused the error when an exception happens here
        except (ValueError, RuntimeError):
            raise TypeError("Batch does not support non-stackable list/tuple "
                            "of tensors as value yet.")
```

duburcqa: It should be fine now. Let me know.

tianshou/data/batch.py
```diff
@@ -36,8 +37,11 @@ def _is_scalar(value: Any) -> bool:
     # 3. python object rather than dict / Batch / tensor
     # the check of dict / Batch is omitted because this only checks a value.
     # a dict / Batch will eventually check their values
-    value = np.asanyarray(value)
-    return value.size == 1 and not value.shape
+    if isinstance(value, (np.ndarray, torch.Tensor)):
+        return value.size == 1 and not value.shape
```
youkaichao: for torch.Tensor, use ``value.numel()``
```python
import torch
a = torch.ones((3, 4))
a.size
Out[15]: <function Tensor.size>
a.numel()
Out[16]: 12
```

tianshou/data/batch.py
```diff
@@ -36,8 +37,11 @@ def _is_scalar(value: Any) -> bool:
...
+    else:
+        value = np.asanyarray(value)
+        return _is_scalar(value)
```
youkaichao: Maybe just adding two lines is ok. Why make a recursive call?

Trinkle23897: Maybe something like:
```python
if not isinstance(value, torch.Tensor):
    value = np.asanyarray(value)
return value.size == 1 and not value.shape
```
?

tianshou/data/batch.py
duburcqa: Indeed. I thought it was the same for both. FIxed

tianshou/data/batch.py
```diff
@@ -113,25 +121,29 @@ def _assert_type_keys(keys):
...
+            v_ = _to_array_with_correct_type(v)
+        except ValueError as e:
+            raise TypeError("Batch does not support heterogeneous list/tuple"
+                            " of tensors as unique value yet.") from e
+        if _is_batch_set(v):
+            v = Batch(v)  # list of dict / Batch
+        else:
+            # None, scalar, normal data list (main case)
+            # or an actual list of objects
+            v = v_
```
youkaichao: Nice usage of ``raise ... from``!

tianshou/data/batch.py
duburcqa: It should be OK now.

NB: To be accurate, it is not "recursive" but rather "nested to itself". Since recursive means: a procedure that can repeat itself indefinitely

tianshou/data/batch.py
duburcqa: I knew you would like it :smile:

tianshou/data/batch.py
youkaichao: > It should be OK now.

Forgot to push?

youkaichao (dismissed):
Nice work👍

Trinkle23897:
This case can bypass the check:
```python
In [10]: d=Batch(a=[1, np.zeros([3,3]), np.zeros([3,3]), torch.zeros(3,3)])

In [11]: d
Out[11]: 
Batch(
    a: array([1, array([[0., 0., 0.],
              [0., 0., 0.],
              [0., 0., 0.]]),
              array([[0., 0., 0.],
              [0., 0., 0.],
              [0., 0., 0.]]),
              tensor([[0., 0., 0.],
               [0., 0., 0.],
               [0., 0., 0.]])], dtype=object),
)

In [12]: Batch.cat([d,d])
Out[12]: 
Batch(
    a: array([1, array([[0., 0., 0.],
              [0., 0., 0.],
              [0., 0., 0.]]),
              array([[0., 0., 0.],
              [0., 0., 0.],
              [0., 0., 0.]]),
              tensor([[0., 0., 0.],
               [0., 0., 0.],
               [0., 0., 0.]]),
              1, array([[0., 0., 0.],
              [0., 0., 0.],
              [0., 0., 0.]]),
              array([[0., 0., 0.],
              [0., 0., 0.],
              [0., 0., 0.]]),
              tensor([[0., 0., 0.],
               [0., 0., 0.],
               [0., 0., 0.]])], dtype=object),
)
```
I'm not sure this case is well enough. I think it is too "corner-case".

Trinkle23897 (dismissed):

duburcqa:
> I'm not sure this case is well enough.

Indeed, I though it was forbidden. I will fix it and add a test case.

youkaichao (dismissed):

duburcqa:
It is fine now.

Trinkle23897:
I think `d=Batch(a=[])` should not raise an error.

duburcqa:
> I think d=Batch(a=[]) should not raise an error.

Indeed. Fixed. Added unit test.

Trinkle23897 (approved):

youkaichao (approved):

PR #154, opened 2020-07-20
buffer update bug fix

### LINKED ISSUE #153: buffer update fails when stack_num is not 0
```python
>>> from tianshou.data.buffer import ReplayBuffer
>>> import numpy as np
>>> buf1 = ReplayBuffer(4, stack_num=2)
>>> for i in range(5):
...     buf1.add(**{'obs':np.array([i]), 'act':float(i), 'rew':i*i, 'done': False})
... 
>>> buf2 = ReplayBuffer(4, stack_num=2)
>>> buf2.update(buf1)
>>> print(buf1)
ReplayBuffer(
    obs: array([[4],
                [1],
                [2],
                [3]]),
    act: array([4., 1., 2., 3.]),
    rew: array([16,  1,  4,  9]),
    done: array([False, False, False, False]),
    obs_next: Batch(),
    info: Batch(),
    policy: Batch(),
)
>>> print(buf2)
ReplayBuffer(
    obs: array([[[1],
                 [1]],
         
                [[1],
                 [2]],
         
                [[2],
                 [3]],
         
                [[3],
                 [4]]]),
    act: array([1., 2., 3., 4.]),
    rew: array([ 1,  4,  9, 16]),
    done: array([False, False, False, False]),
    obs_next: Batch(),
    info: Batch(),
    policy: Batch(),
)
```

I think this is due to the reason that update() uses advanced slice and calls get() in __getitem__ function, which returns stacked results. I will open a PR to fix this if necessary.

ChenDRAG:
fix #153


tianshou/data/buffer.py
```diff
@@ -163,7 +163,19 @@ def update(self, buffer: 'ReplayBuffer') -> None:
             return
         i = begin = buffer._index % len(buffer)
         while True:
-            self.add(**buffer[i])
+            obs_next = None if isinstance(
```
duburcqa: This is quite ugly and has several duplicates. I would rather patch `add` method if possible.

Trinkle23897 (approved):

duburcqa (approved):

PR #155, opened 2020-07-21
Fix shape inconsistency in A2CPolicy and PPOPolicy

### LINKED ISSUE #101: Does the current PPO support discrete actions?
Hi, how an exciting work you have done in tianshou! But there still some doubts while I use the code in my experiments. I have found that there are no different in the continues version PPO and discrete version PPO. However, your experimental results show that the discrete version PPO work well in CartPole-v0. But when I change the env to MountainCar and Acrobot it does not work at all.

Trinkle23897:
* The original `r - v`'s shape in A2C is wrong.
* The shape of log_prob is different: [bsz] in Categorical and [bsz, 1] in Normal. Should manually make the shape to be consistent with other tensors.


tianshou/policy/modelfree/a2c.py
```diff
@@ -97,9 +97,10 @@ def forward(self, batch: Batch,
     def learn(self, batch: Batch, batch_size: int, repeat: int,
               **kwargs) -> Dict[str, List[float]]:
         self._batch = batch_size
-        r = batch.returns
+        r = batch.returns[:, None]
```
duburcqa: I don't like this notation. Rather use `np.newaxis` instead of None for `numpy`, or `unsqueeze(-1)` for `torch`. Or `reshape((*v.shape, 1))` if it must supports both `torch` and `numpy`.

tianshou/policy/modelfree/ddpg.py
```diff
@@ -143,8 +143,7 @@ def forward(self, batch: Batch,
 
     def learn(self, batch: Batch, **kwargs) -> Dict[str, float]:
         current_q = self.critic(batch.obs, batch.act)
-        target_q = to_torch_as(batch.returns, current_q)
-        target_q = target_q[:, None]
+        target_q = batch.returns[:, None]
```
duburcqa: same.

tianshou/policy/modelfree/ppo.py
```diff
@@ -133,8 +133,7 @@ def learn(self, batch: Batch, batch_size: int, repeat: int,
         batch.v = torch.cat(v, dim=0)  # old value
         batch.act = to_torch_as(batch.act, v[0])
         batch.logp_old = torch.cat(old_log_prob, dim=0)
-        batch.returns = to_torch_as(
-            batch.returns, v[0]).reshape(batch.v.shape)
+        batch.returns = to_torch_as(batch.returns, v[0])[:, None]
```
duburcqa: same.

tianshou/policy/modelfree/sac.py
```diff
@@ -140,7 +140,7 @@ def _target_q(self, buffer: ReplayBuffer,
     def learn(self, batch: Batch, **kwargs) -> Dict[str, float]:
         # critic 1
         current_q1 = self.critic1(batch.obs, batch.act)
-        target_q = to_torch_as(batch.returns, current_q1)[:, None]
+        target_q = batch.returns[:, None]
```
duburcqa: same.

tianshou/policy/modelfree/a2c.py
Trinkle23897: `None` is supported for both np and torch

duburcqa: Yes I know, hence my comment. But it is very ugly and confusing. I have never seen anybody using it directly in real project, and I don't consider doing it as a good practice.

Trinkle23897: okay

duburcqa (dismissed):

tianshou/policy/modelfree/a2c.py
```diff
@@ -105,11 +105,12 @@ def learn(self, batch: Batch, batch_size: int, repeat: int,
             for b in batch.split(batch_size):
                 self.optim.zero_grad()
                 dist = self(b).dist
-                v = self.critic(b.obs)
+                v = self.critic(b.obs).squeeze()
```
duburcqa: You should specify the axis if possible. This way the behaviour is independent of the size of the other dimensions, making the output more predictable.

tianshou/policy/modelfree/ddpg.py
```diff
@@ -142,9 +142,8 @@ def forward(self, batch: Batch,
         return Batch(act=actions, state=h)
 
     def learn(self, batch: Batch, **kwargs) -> Dict[str, float]:
-        current_q = self.critic(batch.obs, batch.act)
-        target_q = to_torch_as(batch.returns, current_q)
-        target_q = target_q[:, None]
+        current_q = self.critic(batch.obs, batch.act).squeeze()
```
duburcqa: same

tianshou/policy/modelfree/a2c.py
Trinkle23897: This is a tensor with shape `(bsz, 1)`. I'd like to squeeze all the shape.

Btw, another critical issue: Categorical distribution outputs [bsz] while Normal distribution outputs [bsz, 1], so at this time we cannot determine whether this log_prob tensor is squeezed or not. Thus I use squeeze() in all log_prob.

Trinkle23897:
Should be okay now :)

youkaichao:
It should be better to document somewhere the shape of these variables. ndarray with shape of ``(n,)`` is very error-prone and annoying because of mis-broadcasting.

tianshou/policy/modelfree/a2c.py
duburcqa: There is no side effect to also squeeze `bsz` part ? because if any of the dim has size 1, it will be removed. Are you sure it cannot be an issue ?

duburcqa:
> It should be better to document somewhere the shape of these variables.

I agree.

tianshou/policy/modelfree/a2c.py
Trinkle23897: Okay.

duburcqa (approved):

youkaichao (approved):

PR #157, opened 2020-07-23
unify single-env and multi-env in collector

youkaichao:
Unify the implementation with multi-environments (wrap a single environment in a multi-environment with one envs) to greatly simplify the code.

This changed the behavior of single-environment.
Prior to this pr, for single environment, ``collector.collect(n_step=n)`` will step n steps.
After this pr, for single environment, ``collector.collect(n_step=n)`` will step m episodes until the steps are greater than n.

That is to say, collectors now always collect **full episodes**.

### DIFF
--- a/test/base/test_collector.py
+++ b/test/base/test_collector.py
@@ -98,7 +98,7 @@ def test_collector_with_dict_state():
     policy = MyPolicy(dict_state=True)
     c0 = Collector(policy, env, ReplayBuffer(size=100), preprocess_fn)
     c0.collect(n_step=3)
-    c0.collect(n_episode=3)
+    c0.collect(n_episode=2)
     env_fns = [lambda x=i: MyTestEnv(size=x, sleep=0, dict_state=True)
                for i in [2, 3, 4, 5]]
     envs = VectorEnv(env_fns)
@@ -126,9 +126,10 @@ def test_collector_with_ma():
     policy = MyPolicy()
     c0 = Collector(policy, env, ReplayBuffer(size=100),
                    preprocess_fn, reward_metric=reward_metric)
+    # n_step=3 will collect a full episode
     r = c0.collect(n_step=3)['rew']
-    assert np.asanyarray(r).size == 1 and r == 0.
-    r = c0.collect(n_episode=3)['rew']
+    assert np.asanyarray(r).size == 1 and r == 4.
+    r = c0.collect(n_episode=2)['rew']
     assert np.asanyarray(r).size == 1 and r == 4.
     env_fns = [lambda x=i: MyTestEnv(size=x, sleep=0, ma_rew=4)
                for i in [2, 3, 4, 5]]
--- a/test/base/test_collector.py
+++ b/test/base/test_collector.py
@@ -46,6 +46,19 @@ class Logger:
         else:
             return Batch()
 
+    @staticmethod
+    def single_preprocess_fn(**kwargs):
+        # same as above, without tfb
+        if not kwargs.get('info', Batch()).is_empty():
+            n = len(kwargs['obs'])
+            info = kwargs['info']
+            for i in range(n):
+                info[i].update(rew=kwargs['rew'][i])
+            return Batch(info=info)
+            # or: return {'info': info}
+        else:
+            return Batch()
+
 
 def test_collector():
     writer = SummaryWriter('log/collector')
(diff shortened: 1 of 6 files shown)


youkaichao:
This is a prerequisite for #134 .

Trinkle23897 (dismissed):

youkaichao:
``Collector.collect`` should only reflect the stat of current call. If users want to track the stat (such as moving average) of ``Collector.collect``, they can achieve it with minimal efforts, but this should not be done by Tianshou anyway. So I commited d3e23ac .

duburcqa:
To me it is a great improvement. Reducing the number of branches is not easy but often it makes everything simpler and clearer. And especially this part of the code was unnecessarily complicated and hard to understand for no reason (probably an artefact of the past), since a single env is nothing conceptually different from a vector venv with a single instance. Good you took the time to clean that part of the code !

youkaichao:
> And especially this part of the code was unnecessarily complicated and hard to understand for no reason (probably an artefact of the past), since a single env is nothing conceptually different from a vector venv with a single instance.

I realized this when I want to support sync simulation, and get support from @Trinkle23897 . It is indeed an artefact of the past because he wants to support partial episodes, but later he realized that it is hardly necessary, so I removed this part of code.

youkaichao:
ready for review now @Trinkle23897  @duburcqa

test/base/test_collector.py
```diff
@@ -60,16 +70,16 @@ def test_collector():
...
-    c0.collect(n_step=3, log_fn=logger.log)
+                   logger.preprocess_fn)
+    c0.collect(n_step=3)
     assert np.allclose(c0.buffer.obs[:3], [0, 1, 0])
     assert np.allclose(c0.buffer[:3].obs_next, [1, 2, 1])
-    c0.collect(n_episode=3, log_fn=logger.log)
+    c0.collect(n_episode=3)
     assert np.allclose(c0.buffer.obs[:8], [0, 1, 0, 1, 0, 1, 0, 1])
     assert np.allclose(c0.buffer[:8].obs_next, [1, 2, 1, 2, 1, 2, 1, 2])
     c0.collect(n_step=3, random=True)
```
Trinkle23897: It should be changed since single-env does not exactly collect n_episode and n_step. Check for other cases nearby.

tianshou/data/collector.py
```diff
@@ -274,110 +242,70 @@ def collect(self,
...
-                    self.data.obs_next = obs_next
-                if n_episode != 0 and cur_episode >= n_episode:
+            for i in range(self.env_num):
+                self._cached_buf[i].add(**self.data[i])
+                if self.data.done[i]:
+                    if n_step or np.isscalar(n_episode) or \
+                            episode_count[i] < n_episode[i]:
+                        episode_count[i] += 1
+                        reward_total += np.asarray(
+                            self._cached_buf[i].rew).sum(axis=0)
```
duburcqa: What is the type of `self._cached_buf[i].rew` ? To me it can only be a Batch or `np.ndarray`, so using `asarray` is unecessary. Not to mention that Batch already support `np.sum` natively.

tianshou/data/collector.py
```diff
@@ -81,9 +73,9 @@ class Collector(object):
         #   clear the buffer
         collector.reset_buffer()
 
-    For the scenario of collecting data from multiple environments to a single
-    buffer, the cache buffers will turn on automatically. It may return the
-    data more than the given limitation.
+    Collected data always consist of full episodes. So if only ``n_step``
+    argument is give, the collector may return the data more than the
+    ``n_step`` limitation.
```
Trinkle23897: add n_episode after n_step

tianshou/data/collector.py
```diff
@@ -195,24 +168,22 @@ def _reset_state(self, id: Union[int, List[int]]) -> None:
...
                 ) -> Dict[str, float]:
         """Collect a specified number of step or episode.
 
         :param int n_step: how many steps you want to collect.
-        :param n_episode: how many episodes you want to collect (in each
-            environment).
-        :type n_episode: int or list
+        :param n_episode: how many episodes you want to collect. If it is
+            an int, it means to collect at lease ``n_episode`` episodes; if
+            it is list, it means to collect exactly ``n_episode[i]`` episodes
```
Trinkle23897: a list

tianshou/data/collector.py
youkaichao: It is a list, because the cache buffer is based on list.

Trinkle23897: It bypasses the setattr method

duburcqa: Ah yes it remember now.

duburcqa (dismissed):

Trinkle23897 (dismissed):

Trinkle23897 (approved):

jiang-yuan:
According to my experiments:
Even though I change the step_number within an episode (e.g. There is a 30 steps episode, within which I update the net every 5, 10, or 15 steps), the result may be different.
I do not understand. What is the reason for 'he realized that it is hardly necessary'?

Trinkle23897:
We do some experiments on Bipedal-walker and Lunar-lander, the final result seems slightly different with each other. @imoneoi

youkaichao:
> Even though I change the step_number within an episode (e.g. There is a 30 steps episode, within which I update the net every 5, 10, or 15 steps), the result may be different.

The point of this pr is to enforce that there will only be full episodes in the replay buffer. If fully online learning really matters for you, I think you can use a single environment and try to write a collector with simple for-loops.

```python
for xxx:
    data = []
    while len(data) < n:
        data.append(env.step())
    policy.learn(data)
```

In my opinion, if you need a replay buffer, fully online learning (learning after every several steps) will not matter so much. Current implementation (Tianshou 0.2.5) allows users to perform online learning at the episode level (learning after every several episodes) and I think it is enough.

However, feel free to provide feedback and to discuss :)

PR #159, opened 2020-07-23
Improve PER

Trinkle23897:
1. use segment tree to rewrite the previous PrioReplayBuffer code, add the test
2. enable all Q-learning algorithms to use PER


tianshou/data/utils/segtree.py
```diff
@@ -0,0 +1,87 @@
+import numpy as np
+from typing import Union, Optional
+
+
+class SegmentTree:
```
duburcqa, line 32: If this needs to be as fast as possible, consider using `numpy.jit`. You may get a significant speed-up almost for free.

Trinkle23897: It is `numba.jit(nopython=True)` instead of `np.jit`. I'm considering that we can use numba to rewrite some part of our codebase later because it is not a good deal to use numba only in this SegmentTree.

tianshou/data/utils/segtree.py
```diff
@@ -0,0 +1,117 @@
...
+    :param int size: the size of segment tree.
+    :param str operation: the operation of segment tree. Choose one of "sum",
+        "min" and "max", defaults to "sum".
+    """
+
+    def __init__(self, size: int,
+                 operation: str = "sum") -> None:
+        bound = 1
+        while bound < size:
+            bound <<= 1
```
youkaichao: I prefer the pythonic way ``bound  *= 2``.
``bound <<= 1`` seems to tell everyone you were a professional c programmer.

tianshou/data/utils/segtree.py
```diff
@@ -0,0 +1,117 @@
...
+            f"Unknown operation {operation}."
+        (self._op, self._init_value) = {
+            "sum": (np.sum, 0.),
+            "min": (np.min, np.inf),
+            "max": (np.max, -np.inf),
+        }[operation]
+        self._value = np.zeros([bound << 1]) + self._init_value
+
+    def __len__(self):
+        return self._bound
```
youkaichao: Why is it ``self._bound`` rather than ``size``? I think ``self._bound`` is something internal. For users, they just know they have ``size`` elements.

tianshou/data/utils/segtree.py
```diff
@@ -0,0 +1,117 @@
...
+        if isinstance(index, int) and isinstance(value, float):
+            index, value = np.array([index]), np.array([value])
+        assert isinstance(index, np.ndarray) and isinstance(value, np.ndarray)
+        assert ((0 <= index) & (index < self._bound)).all()
+        index = index + self._bound
+        self._value[index] = value
+        while index[0] > 1:
+            index >>= 1
+            self._value[index] = self._op(
+                [self._value[index << 1], self._value[index << 1 | 1]], axis=0)
```
youkaichao: Leave some comments to show how duplicate values are dealt with.

tianshou/data/utils/segtree.py
```diff
@@ -0,0 +1,117 @@
...
+
+    1. Find out the smallest n which safisfies ``size <= 2^n``, and let \
+    ``bound = 2^n``. This is to ensure that all leaf nodes are in the same \
+    depth inside the segment tree.
+    2. Store the original value to leaf nodes in ``[bound:bound * 2]``, and \
+    the union of elementary to internal nodes in ``[1:bound]``. The internal \
+    node follows the rule: \
+    ``value[i] = operation(value[i * 2], value[i * 2 + 1])``.
+    3. Update a node takes O(log(bound)) time complexity.
+    4. Query an interval [l, r] with the default operation takes O(log(bound))
```
youkaichao: I don't think these implementation details are necessary in the doc of SegmentTree. Padding the array to size of 2^n is just an implementation detail.

A more proper way is to say something like the following:

Store an array ``arr`` with size ``n`` in a segment tree, support value update and fast query of ``min/max/sum arr[left:right]`` in O(log(n)) time.

tianshou/data/utils/segtree.py
youkaichao: ``np.min([a, b], axis=0)`` will create a new array ``x = np.array([a, b])``. Maybe you are talking about ``np.minimum(a, b)`` which does element-wise comparison.

tianshou/data/utils/segtree.py
```diff
@@ -0,0 +1,117 @@
...
+        bound = 1
+        while bound < size:
+            bound <<= 1
+        self._bound = bound
+        assert operation in ["sum", "min", "max"], \
+            f"Unknown operation {operation}."
+        (self._op, self._init_value) = {
+            "sum": (np.sum, 0.),
+            "min": (np.min, np.inf),
+            "max": (np.max, -np.inf),
```
youkaichao: ``np.ndarray.__add__``, ``np.minimum`` and ``np.maximum``.

tianshou/data/utils/segtree.py
youkaichao: Are you sure it is ``index << 1 | 1`` rather than ``index << 1 + 1``?

tianshou/data/utils/segtree.py
```diff
@@ -0,0 +1,117 @@
...
+        if not isinstance(value, np.ndarray):
+            value = np.array([value])
+            single = True
+        assert (value <= self._value[1]).all()
+        index = np.ones(value.shape, dtype=np.int)
+        index = self.__class__._get_prefix_sum_idx(
+            index, value, self._bound, self._value)
+        return index.item() if single else index
+
+    # numba version, 5x speed up
```
duburcqa: Nice !

duburcqa (dismissed):

tianshou/data/utils/segtree.py
Trinkle23897: That's why I use axis=0

youkaichao: I mean,  ``np.minimum(a, b)`` is way more efficient than ``np.min([a, b], axis=0)``, although the latter is correct.

tianshou/data/utils/segtree.py
```diff
@@ -0,0 +1,117 @@
...
+                    value: Union[float, np.ndarray]) -> None:
+        """Insert or overwrite a (or some) value(s) in this segment tree. The
+        duplicate values are handled as numpy array, in other words, we only
+        keep the last value and ignore the previous same value.
+        """
+        if isinstance(index, int) and isinstance(value, float):
+            index, value = np.array([index]), np.array([value])
+        assert ((0 <= index) & (index < self._bound)).all()
+        index = index + self._bound
+        self._value[index] = value
```
youkaichao: Users just have _size elements. So you should check if the index exceeds self._size rather than self._bound
``assert ((0 <= index) & (index < self._size)).all()``

tianshou/data/utils/segtree.py
```diff
@@ -0,0 +1,117 @@
...
+        return self._value[index + self._bound]
+
+    def __setitem__(self, index: Union[int, np.ndarray],
+                    value: Union[float, np.ndarray]) -> None:
+        """Insert or overwrite a (or some) value(s) in this segment tree. The
+        duplicate values are handled as numpy array, in other words, we only
+        keep the last value and ignore the previous same value.
+        """
+        if isinstance(index, int) and isinstance(value, float):
+            index, value = np.array([index]), np.array([value])
```
youkaichao: Is it necessary to convert ``value``? To me it seems value is ok if it is float.

Trinkle23897: Good catch!

tianshou/data/utils/segtree.py
```diff
@@ -0,0 +1,122 @@
...
+                result = self._op(result, self._value[start ^ 1])
+            if end % 2 == 1:
+                result = self._op(result, self._value[end ^ 1])
+            start, end = start // 2, end // 2
+        return result
+
+    def get_prefix_sum_idx(
+            self, value: Union[float, np.ndarray]) -> Union[int, np.ndarray]:
+        """Return the index ``i`` which satisfies
+        ``sum(value[:i]) <= value < sum(value[:i + 1])``.
```
youkaichao: If there are multiple indexes that satisfy this, return the largest, smallest, or whichever one?

youkaichao:
Discussion: is it necessary to support segment tree with min / max operator? By definition, segment tree supports any binary commutative operators ``op(a, op(b, c)) = op(op(a, b), c)``. But the sum tree we use is not an exact segment tree: we require elements in the tree are non-negative.

In my opinion, we can just use a dedicated sum tree, without support min / max tree because min / max tree has no practical usage in priorized experience replay.

I'm open to different opinions.

Trinkle23897:
> Discussion: is it necessary to support the segment tree with min/max operator? By definition, segment tree supports any binary commutative operators `op(a, op(b, c)) = op(op(a, b), c)`. But the sum tree we use is not an exact segment tree: we require elements in the tree that are non-negative.

This is only the constraint for `get_prefix_sum_idx`, but for the normal usage of Segment Tree (`op(value[l:r])`), that doesn't exist.

tianshou/data/utils/segtree.py
youkaichao: This must be the smallest one. Suppose we are using a buffer with capacity of 100, and there are just three elements now, we have to assure the returned index is smaller than 3, otherwise we are trying to access an transition tuple that does not exist.

test/base/test_buffer.py
```diff
@@ -112,9 +115,116 @@ def test_update():
...
+            naive[index] = value
+            tree[index] = value
+            for i in range(actual_len):
+                for j in range(i, actual_len):
+                    try:
+                        ref = realop(naive[i:j])
+                    except ValueError:
+                        continue
+                    out = tree.reduce(i, j)
+                    assert np.allclose(ref, out), (i, j, ref, out)
```
youkaichao: At first glance, I think this is wrong. But later I figure out this is ``assert np.allclose(ref, out)``, and the error message is ``(i, j, ref, out)``. It works, I admit. But it is rather confusing.

test/base/test_buffer.py
```diff
@@ -112,9 +115,116 @@ def test_update():
...
+            index = np.random.randint(actual_len)
+            value = np.random.rand()
+            naive[index] = value
+            tree[index] = value
+            for i in range(actual_len):
+                for j in range(i, actual_len):
+                    try:
+                        ref = realop(naive[i:j])
+                    except ValueError:
+                        continue
```
youkaichao: Why raise an error here?

test/base/test_buffer.py
```diff
@@ -112,9 +115,116 @@ def test_update():
...
+            index = np.random.choice(actual_len, size=4)
+            value = np.random.rand(4)
+            naive[index] = value
+            tree[index] = value
+            assert np.allclose(realop(naive), tree.reduce())
+            for i in range(10):
+                left = right = 0
+                while left >= right:
+                    left = np.random.randint(actual_len)
+                    right = np.random.randint(actual_len)
```
youkaichao: Bad code. Just random two numbers and let ``left`` be the smaller one.

test/base/test_buffer.py
```diff
@@ -112,9 +115,116 @@ def test_update():
...
+                assert np.allclose(realop(naive[left:right]),
+                                   tree.reduce(left, right))
+        # large test
+        actual_len = 16384
+        tree = SegmentTree(actual_len, op)
+        naive = np.zeros([actual_len]) + init
```
youkaichao: Change all occurrence to ``np.full``.

test/base/test_buffer.py
```diff
@@ -112,9 +115,116 @@ def test_update():
...
+            index = np.random.choice(actual_len, size=64)
+            value = np.random.rand(64)
+            naive[index] = value
+            tree[index] = value
+            assert np.allclose(realop(naive), tree.reduce())
+            for i in range(10):
+                left = right = 0
+                while left >= right:
+                    left = np.random.randint(actual_len)
+                    right = np.random.randint(actual_len)
```
youkaichao: The same. Bad code.

test/base/test_buffer.py
```diff
@@ -112,9 +115,116 @@ def test_update():
...
+
+        def sample_npbuf():
+            return np.random.choice(size, bsz, p=naive / naive.sum())
+
+        def sample_tree():
+            scalar = np.random.rand(bsz) * tree.reduce()
+            return tree.get_prefix_sum_idx(scalar)
+
+        print('npbuf', timeit(sample_npbuf, setup=sample_npbuf, number=1000))
+        print('tree', timeit(sample_tree, setup=sample_tree, number=1000))
```
youkaichao, line 214: It is better to make it a separate function?

test/base/test_buffer.py
```diff
@@ -112,9 +115,116 @@ def test_update():
...
+
+
 if __name__ == '__main__':
+    test_priortized_replaybuffer()
     test_replaybuffer()
     test_ignore_obs_next()
     test_stack()
+    test_segtree()
     test_priortized_replaybuffer(233333, 200000)
```
youkaichao: Why test ``test_priortized_replaybuffer `` twice?

tianshou/data/buffer.py
```diff
@@ -387,68 +384,48 @@ def add(self,
...
-                f"batch_size should be less than {len(self)}, \
-                    or set replace=True")
+            scalar = np.random.rand(batch_size) * self.weight.reduce()
+            indice = self.weight.get_prefix_sum_idx(scalar)
         batch = self[indice]
-        batch["impt_weight"] = (self._size * p) ** (-self._beta)
+        # impt_weight
+        # original formula: ((p_j/p_sum*N)**(-beta))/((p_min/p_sum*N)**(-beta))
+        # simplified formula: (p_j/p_min)**(-beta)
+        batch.weight = (batch.weight / self._min_prio) ** (-self._beta)
```
youkaichao, line 428: In the doc of ``sample()``, describe what does the ``weight`` field in the returned batch mean. It is not the importance sampling weight, but the weight on loss functions to unbias the importance sampling procedure.

tianshou/policy/base.py
```diff
@@ -213,4 +214,11 @@ def compute_nstep_return(
         returns = to_torch_as(returns, target_q)
         gammas = to_torch_as(gamma ** gammas, target_q)
         batch.returns = target_q * gammas + returns
+        # prio buffer update
+        if isinstance(buffer, PrioritizedReplayBuffer):
+            batch.update_weight = buffer.update_weight
```
youkaichao, line 219: Mounting a function ``buffer.update_weight`` to be a field of batch objects is a hack and should be avoided. Since it has been here for a while and this PR is already very large, I will open a new PR to deal with it.

The solution may be something like the following: add a function ``BasePolicy .update`` and a function ``BasePolicy.post_process_fn``. The update of weight into the buffer can be done in ``BasePolicy.post_process_fn``. Trainer functions just have to call ``BasePolicy.update``.

```python
def update(self, buffer, batch_size):
    batch, indices = buffer.sample(batch_size)
    self.process_fn(batch, buffer, indices)
    self.learn(batch)
    self.post_process_fn(batch, buffer, indices)
```

tianshou/policy/modelfree/td3.py
```diff
@@ -119,16 +118,22 @@ def learn(self, batch: Batch, **kwargs) -> Dict[str, float]:
         # critic 1
         current_q1 = self.critic1(batch.obs, batch.act).flatten()
         target_q = batch.returns.flatten()
-        critic1_loss = F.mse_loss(current_q1, target_q)
+        td1 = current_q1 - target_q
+        critic1_loss = (td1.pow(2) * batch.weight).mean()
+        # critic1_loss = F.mse_loss(current_q1, target_q)
```
youkaichao, line 123: Make sure ``[n] * [n, 1] ==> [n, n]`` does not happen again :) The same for several occurrence above.

test/base/test_buffer.py
Trinkle23897: realop(naive[1:1])

test/base/test_buffer.py
Trinkle23897: The left cannot be equal to the right.

tianshou/policy/modelfree/td3.py
Trinkle23897: Before this line, we ensure `current_q` and `target_q` are both 1-d array (same as batch.weight), so that that cannot happen.

test/base/test_buffer.py
Trinkle23897: Because during my debugging, I once passed the large test but failed in the small test. This is related to the index issue: I forgot to deep-copy the index in `__setitem__`: `index += self._bound`. Later this small test failed and I modified to `index = index + self._bound`.

test/base/test_buffer.py
youkaichao: That's easy, ``left = np.random.randint(actual_len)``, ``right = np.random.randint(left + 1, actual_len + 1)``.

test/base/test_buffer.py
youkaichao: Okay, then put these two tests in nearby lines.

youkaichao (dismissed):
A nice PR to improve the efficiency of prioritized buffer!

youkaichao (dismissed):

Trinkle23897:
@duburcqa should be okay now and please have a check.

youkaichao (approved):

PR #160, opened 2020-07-24
Yet another 3 fix

### LINKED ISSUE #162: Potential Bug
    + [√ ] exception-raising bug
    + [ ] RL algorithm bug
    + [ ] new feature request
- [ √] I have visited the [source website](https://github.com/thu-ml/tianshou/)
- [ √] I have searched through the [issue tracker](https://github.com/thu-ml/tianshou/issues) for duplicates
- [ √] I have mentioned version numbers, operating system and environment, where applicable:
  ```python
  import tianshou, torch, sys
  # 0.2.4 1.4.0 3.7.6 | packaged by conda-forge | (default, Jun  1 2020, 18:57:50) [GCC 7.5.0] linux
  ```
Hi, first thanks for the excellent work! 

One thing I found may has potential bug. 
When I was using a2c policy, I found that you calculate [a_loss](https://github.com/thu-ml/tianshou/blob/dev/tianshou/policy/modelfree/a2c.py#L111) in this way, which may not work if the action has multiple dimension, say if the shape of output of dist.log_prob(a) is [bsz, n] while the shape of (r-v) is [bsz, 1] or [bsz] and neither of them can do dot product with the dis.log_prob(a). In my opinion, it would be better to use dist.log_prob(a).transpose(0,1) so that it will do multiplication among the same dimension.

Trinkle23897:

1. DQN learn should keep eps=0
2. Add a warning of env.seed in VecEnv
3. fix #162 of multi-dim action

### DIFF
--- a/tianshou/utils/net/common.py
+++ b/tianshou/utils/net/common.py
@@ -36,7 +36,10 @@ class Net(nn.Module):
     def forward(self, s, state=None, info={}):
         """s -> flatten -> logits"""
         s = to_torch(s, device=self.device, dtype=torch.float32)
-        s = s.flatten(1)
+        if len(s.shape) > 1:
+            s = s.flatten(1)
+        else:
+            s = s.unsqueeze(-1)
         logits = self.model(s)
         return logits, state
 
--- a/tianshou/utils/net/common.py
+++ b/tianshou/utils/net/common.py
@@ -36,10 +36,7 @@ class Net(nn.Module):
     def forward(self, s, state=None, info={}):
         """s -> flatten -> logits"""
         s = to_torch(s, device=self.device, dtype=torch.float32)
-        if len(s.shape) > 1:
-            s = s.flatten(1)
-        else:
-            s = s.unsqueeze(-1)
+        s = s.reshape(s.size(0), -1)
         logits = self.model(s)
         return logits, state
 
--- a/docs/tutorials/cheatsheet.rst
+++ b/docs/tutorials/cheatsheet.rst
@@ -48,6 +48,17 @@ where ``env_fns`` is a list of callable env hooker. The above code can be writte
     env_fns = [lambda x=i: MyTestEnv(size=x) for i in [2, 3, 4, 5]]
     venv = SubprocVectorEnv(env_fns)
 
+.. warning::
+
+    If you use your own environment, please make sure the ``seed`` method is set up properly, e.g.,
+
+    ::
+
+        def seed(self, seed):
+            np.random.seed(seed)
+
+    Otherwise, the outputs of these envs will be the same with each other.
+
 .. _preprocess_fn:
 
 Handle Batched Data Stream in Collector
--- a/docs/tutorials/cheatsheet.rst
+++ b/docs/tutorials/cheatsheet.rst
@@ -57,7 +57,7 @@ where ``env_fns`` is a list of callable env hooker. The above code can be writte
         def seed(self, seed):
             np.random.seed(seed)
 
-    Otherwise, the outputs of these envs will be the same with each other.
+    Otherwise, the outputs of these envs may be the same with each other.
 
 .. _preprocess_fn:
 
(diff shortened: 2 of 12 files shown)


Trinkle23897:
It's no difference with the previous version because `self(batch).act` is not used in `learn()`, since `eps` only changes `act`.
But pointing it out is much better.

tianshou/utils/net/common.py
```diff
@@ -36,7 +36,10 @@ def __init__(self, layer_num, state_shape, action_shape=0, device='cpu',
     def forward(self, s, state=None, info={}):
         """s -> flatten -> logits"""
         s = to_torch(s, device=self.device, dtype=torch.float32)
-        s = s.flatten(1)
+        if len(s.shape) > 1:
```
duburcqa: it is quite ugly x) what is the shape of s ?

Trinkle23897: It is not determined, either [bsz] or [bsz, ?] or [bsz, ?, ?] or [bsz, ?, ?, ?]. I want to change the shape as (bsz, -1), so it's better to use reshape

youkaichao (approved):

youkaichao:
Reshape is good, and would be better if you can avoid shape [bsz] by all means (reward has this shape) because of the [confusing behavior](https://stackoverflow.com/questions/40866598/broadcasting-n-and-with-n-1-arrays-in-numpy).

tianshou/utils/net/common.py
duburcqa: ok so why not `s.flatten(1, -1)` ?

duburcqa:
> Reshape is good, and would be better if you can avoid shape [bsz] by all means (reward has this shape) because of the confusing behavior.

I don't agree, usually I prefer to use `.flatten(1, -1)` since it makes it clear that you want to flatten but preserve batch processing.

Trinkle23897:
> > Reshape is good, and would be better if you can avoid shape [bsz] by all means (reward has this shape) because of the confusing behavior.
> 
> I don't agree, usually I prefer to use `.flatten(1, -1)` since it makes it clear that you want to flatten but preserve batch processing.

But 1-dim tensor cannot apply flatten(1):

```python
In [6]: b=torch.rand(3)

In [7]: b.flatten(1,-1)
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
<ipython-input-7-0fd10ca82bb9> in <module>
----> 1 b.flatten(1,-1)

IndexError: Dimension out of range (expected to be in range of [-1, 0], but got 1)
```

duburcqa:
> But 1-dim tensor cannot apply flatten(1):

OK so your example before edit "`[bsz, ?]` or `[bsz, ?, ?]`" was wrong. So yes if `[bsz]` can happen (which I highly doubt, since it means 0-dim input without batch-processing) then `reshape` must be used.

youkaichao:
>  if [bsz] can happen (which I highly doubt, since it means 0-dim input without batch-processing)

reward is the case.

duburcqa:
> reward is the case.

The reward is used as input of its own neural network ? In which algorithm ?

youkaichao:
> The reward is used as input of its own neural network

I'm not talking about using reward as input, though. My suggestion is to reshape reward to be [bsz, 1] in ``Policy.learn`` :)

Trinkle23897:
> > The reward is used as input of its own neural network
> 
> I'm not talking about using reward as input, though. My suggestion is to reshape reward to be [bsz, 1] in `Policy.learn` :)

I don't agree to reshape this as [bsz, 1] since it had already made a bug previously.

duburcqa:
> I'm not talking about using reward as input, though. My suggestion is to reshape reward to be [bsz, 1] in Policy.learn :)

We are talking about [this](https://github.com/thu-ml/tianshou/pull/160#discussion_r459922060) initially.

Trinkle23897:
Okay. This is because I write a small test with MyTestEnv under the `test/base/` folder. I found that the obs in this env cannot be processed through the current network (because its shape [bsz]), so I modify this line of code.
I know it rarely happens :)

duburcqa (approved):

PR #165, opened 2020-07-25
Replay buffer allows stack_num = 1

duburcqa:
I think it is more convenient to consider "no stacking" = 1 instead of 0, as it is the case now. It is useful in my case because it enables to increase number of observation steps provided to the model without having to do an exception for "no stack" case.

### DIFF
--- a/tianshou/data/buffer.py
+++ b/tianshou/data/buffer.py
@@ -104,7 +104,7 @@ class ReplayBuffer:
 
     :param int size: the size of replay buffer.
     :param int stack_num: the frame-stack sampling argument, should be greater
-        than 1, defaults to 0 (no stacking).
+        than or equal to 1, defaults to 1 (no stacking).
     :param bool ignore_obs_next: whether to store obs_next, defaults to
         ``False``.
     :param bool sample_avail: the parameter indicating sampling only available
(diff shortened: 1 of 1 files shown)


tianshou/data/buffer.py
```diff
@@ -169,7 +170,7 @@ def update(self, buffer: 'ReplayBuffer') -> None:
             return
         i = begin = buffer._index % len(buffer)
         origin = buffer._get_stack_num()
-        buffer._set_stack_num(0)
+        buffer._set_stack_num(1)
```
Trinkle23897: By the way, could you please change L259 to this style? I have no more comment.

duburcqa: you mean using `_get_stack_num` ? Why just this one ? It also appears lines 208, 209 and 216.

tianshou/data/buffer.py
```diff
@@ -104,27 +104,27 @@ class ReplayBuffer:
 
     :param int size: the size of replay buffer.
     :param int stack_num: the frame-stack sampling argument, should be greater
-        than 1, defaults to 0 (no stacking).
+        than or equal to 1, defaults to 1 (no stacking).
```
Trinkle23897, line 107: I’m not sure it is mentioned in the tutorials. Could you please have a double-check?

tianshou/data/buffer.py
Trinkle23897: Exactly. Because there is just L259 a place that use self._stack instead of _get_stack_num

tianshou/data/buffer.py
duburcqa: According to [this](https://github.com/thu-ml/tianshou/search?q=stack_num&unscoped_q=stack_num), it is not.

tianshou/data/buffer.py
Trinkle23897: Okay. Same as 208 and others. My bad.

duburcqa: Ok so I propose to use Python `@property` decorator. You can have a look to the changes.

Trinkle23897 (approved):

youkaichao (approved):

youkaichao:
Nice little pr to resolve the legacy inconsistency 👍 Maybe this should be mentioned in release notes.

PR #168, opened 2020-07-27
doc update and do not force save 'policy' in np format

### LINKED ISSUE #167: why is the value of `collect_per_step` passed into `n_episode`?
https://github.com/thu-ml/tianshou/blob/bd9c3c7f8d144448c44a350828b2c5222298bd8e/tianshou/trainer/onpolicy.py#L87

I see that the parameter `collect_per_step` in `onpolicy_trainer()` function refers to "the number of frames the collector would collect before the network update", but its value is finally passed into the `n_episode` parameter in `collect()`, I get quite confused about this. Do I misunderstand anything?

Trinkle23897:
fix #167

### DIFF
--- a/examples/pong_a2c.py
+++ b/examples/pong_a2c.py
@@ -23,7 +23,7 @@ def get_args():
     parser.add_argument('--gamma', type=float, default=0.9)
     parser.add_argument('--epoch', type=int, default=100)
     parser.add_argument('--step-per-epoch', type=int, default=1000)
-    parser.add_argument('--collect-per-step', type=int, default=100)
+    parser.add_argument('--collect-per-step', type=int, default=10)
     parser.add_argument('--repeat-per-collect', type=int, default=1)
     parser.add_argument('--batch-size', type=int, default=64)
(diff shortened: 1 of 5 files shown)


youkaichao (approved):

youkaichao:
Since 49fbe4b points out the problem of hyper-parameters, maybe it is time that we take hyper-parameters seriously and tune them carefully. Or leave it in the roadmap as a TODO.

Trinkle23897:
> Since [49fbe4b](https://github.com/thu-ml/tianshou/commit/49fbe4b4a4f5c9e72d53378b063e9f35900dc840) points out the problem of hyper-parameters, maybe it is time that we take hyper-parameters seriously and tune them carefully. Or leave it in the roadmap as a TODO.

Already. The first step is to optimize the training procedure (e.g., asyncVenv is a good first step) and I think it can be further enhanced. Then we will tune these hyperparameters on several scenarios including common mujoco and atari envs.

duburcqa (approved):

PR #170, opened 2020-07-28
Dueling DQN

### LINKED ISSUE #54: Why is dqn playing breakout so bad? I trained 400 epochs and the final result did not exceed 6 points.
I trained 400 epochs and the final result did not exceed 6 points. What parameters need to be adjusted?

yingchengyang:
    + [ ] exception-raising fix
    + [x] new feature

- [ ] I have visited the [source website](https://github.com/thu-ml/tianshou)
  ```python
  import tianshou, torch, sys
  ```


Trinkle23897:
Should be okay now, please have a check.
I also change the tic-tac-toe's hyperparameter since gamma=0.1 is quite unstable with Q-value noise.

tianshou/utils/net/common.py
```diff
@@ -9,35 +10,74 @@ class Net(nn.Module):
...
-        self.model = [
-            nn.Linear(input_size, hidden_layer_size),
-            nn.ReLU(inplace=True)]
+
+        def miniblock(inp, oup):
+            ret = [nn.Linear(inp, oup)]
+            if layer_norm:
+                ret += [nn.LayerNorm(oup)]
+            ret += [nn.ReLU(inplace=True)]
+            return ret
```
youkaichao: It would be better to have a static function rather than a function inside a function. Static functions can be reused elsewhere, but inner functions cannot.

tianshou/utils/net/common.py
```diff
@@ -9,35 +10,74 @@ class Net(nn.Module):
...
+        self.dueling = dueling
+        self.softmax = softmax
         input_size = np.prod(state_shape)
         if concat:
             input_size += np.prod(action_shape)
-        self.model = [
-            nn.Linear(input_size, hidden_layer_size),
-            nn.ReLU(inplace=True)]
+
+        def miniblock(inp, oup):
```
duburcqa: I would have duplicated thus block instead of wrapping it in a small method specifically here, because to me it is more readable but as you want.

tianshou/utils/net/common.py
```diff
@@ -9,35 +10,74 @@ class Net(nn.Module):
...
     please refer to :ref:`build_the_network`.
 
-    :param concat: whether the input shape is concatenated by state_shape
+    :param bool concat: whether the input shape is concatenated by state_shape
         and action_shape. If it is True, ``action_shape`` is not the output
         shape, but affects the input shape.
+    :param bool dueling: whether to use dueling network to calculate Q values
+        (for Dueling DQN), defaults to False.
+    :param bool layer_norm: whether to use layer norm before ReLU, defaults to
+        False.
```
youkaichao: For multi-layer perceptrons, batch normalization is more popular. Or maybe this can be a module so that users can use whatever they like (such as ``norm_layer=torch.nn.BatchNorm1d``)

tianshou/utils/net/common.py
```diff
@@ -9,35 +10,74 @@ class Net(nn.Module):
...
         self.model = nn.Sequential(*self.model)
 
     def forward(self, s, state=None, info={}):
         """s -> flatten -> logits"""
         s = to_torch(s, device=self.device, dtype=torch.float32)
         s = s.reshape(s.size(0), -1)
         logits = self.model(s)
+        if self.dueling is not None:  # Dueling DQN
+            q, v = self.Q(logits), self.V(logits)
+            logits = q - q.mean(dim=1).reshape(-1, 1) + v
```
duburcqa: Rather use `keepdim` optional argument of `mean` instead of `reshape`.

tianshou/utils/net/common.py
Trinkle23897: This is aligned with stable-baselines: https://github.com/hill-a/stable-baselines/blob/master/stable_baselines/deepq/policies.py#L114
It's just a demo network, actually, I want users to create their own.

tianshou/utils/net/common.py
Trinkle23897: I don't think so. You can have a look at the left side: https://github.com/thu-ml/tianshou/pull/170/commits/0d1f0a4a3c8b79311fc2e32f90b52b59c034668a

tianshou/utils/net/common.py
Trinkle23897: But this method is only used in the init.

tianshou/utils/net/common.py
duburcqa: Yes I see, and I had no problem with it. But it is just a matter of taste.

tianshou/utils/net/common.py
duburcqa: `stable-baselines` is already dead. I don't know if it worth the effort to align with it.

duburcqa (dismissed):

tianshou/utils/net/common.py
Trinkle23897: Yes, you are right lol
But I want to use their rl-baseline-zoo for the reference.

Trinkle23897: > For multi-layer perceptrons, batch normalization is more popular. Or maybe this can be a module so that users can use whatever they like (such as norm_layer=torch.nn.BatchNorm1d)

This is indeed more popular in CV instead of RL and NLP. I previously use layer_norm more often (because they are all sequential task).

tianshou/utils/net/common.py
youkaichao: > But this method is only used in the init.

Torchvision can be a reference. Their ResNet has a ``make_layer`` function. Even though it is just used during init, who knows if a user would like to access it ┓( ´∀` )┏

Good user experience makes a good community and attracts devoted users :)

tianshou/utils/net/common.py
youkaichao: > I previously use layer_norm more often (because they are all sequential task).

I think RL tasks are sequential but because they are markovian, the decision itself is not sequential (only depend on the last state). What's more, I'm not proposing to use BN rather than LN, but proposing to modify the ``layer_norm `` argument so that this can support various nomalization layers (and you can still use layer norm as the default).

Trinkle23897: Get it

tianshou/utils/net/common.py
```diff
@@ -1,43 +1,88 @@
...
 
-    def __init__(self, layer_num, state_shape, action_shape=0, device='cpu',
-                 softmax=False, concat=False, hidden_layer_size=128):
+    def __init__(self, layer_num: int, state_shape: tuple,
+                 action_shape: Optional[tuple] = 0,
+                 device: Union[str, torch.device] = 'cpu',
+                 softmax: bool = False, concat: bool = False,
+                 hidden_layer_size: int = 128,
+                 dueling: Tuple[int, int] = None,
+                 norm_layer: nn.modules.Module = None):
```
youkaichao: Type annotation: Optional[nn.modules.Module]

youkaichao (approved):

Trinkle23897 (approved):

Issue #166, opened 2020-07-26
SubprocVectorEnv is not released when collector is closed


ChenDRAG:
```
env_subproc_init = SubprocVectorEnv(
        [lambda: SimpleEnv() for _ in range(8)])
for _ in range(2000):
        c = Collector(policy, env_subproc_ini, buffer)
        c.close()
```
A broken pipe error will be reported. Seems that SubprocVectorEnv is not actually released and cannot be reused once assigned to a collector. I don't know if this can be counted as a bug.
`BrokenPipeError: [Errno 32] Broken pipe
`

ChenDRAG:
I think we should at least give users a hint once the closed subprocvec is used again. Or move function in init() to reset().

Trinkle23897:
`collector.close` is going to be removed in #179 so this issue will close.

PR #174, opened 2020-08-01
ShmemVectorEnv Implementation

ChenDRAG:
Following OpenAI baseline, I implement ShmemVectorEnv which is an optimized version of SubprocVectorEnv that uses shared variables to communicate observations. ShmemVectorEnv has exactly the same API as SubprocVectorEnv, but when `obs.shape` is too large, the usage of shared variables can save us a lot of time.

This pr include:

* Basic Implementation of ShmemVectorEnv
* update in test_env.py to test ShmemVectorEnv
* some improvement in of test_env.py for generalization
* some fix in test/base/env.py and test_env.py

I borrow from openai baseline to implement ShmemVectorEnv Class. Please refer to https://github.com/openai/baselines/blob/master/baselines/common/vec_env/shmem_vec_env.py

### DIFF
--- a/test/base/test_buffer.py
+++ b/test/base/test_buffer.py
@@ -68,10 +68,9 @@ def test_stack(size=5, bufsize=9, stack_num=4):
             obs = env.reset(1)
     indice = np.arange(len(buf))
     assert np.allclose(buf.get(indice, 'obs'), np.expand_dims(
-        np.array([
-            [1, 1, 1, 2], [1, 1, 2, 3], [1, 2, 3, 4],
-            [1, 1, 1, 1], [1, 1, 1, 2], [1, 1, 2, 3],
-            [3, 3, 3, 3], [3, 3, 3, 4], [1, 1, 1, 1]]), axis=2))
+        [[1, 1, 1, 2], [1, 1, 2, 3], [1, 2, 3, 4],
+         [1, 1, 1, 1], [1, 1, 1, 2], [1, 1, 2, 3],
+         [3, 3, 3, 3], [3, 3, 3, 4], [1, 1, 1, 1]], axis=-1))
     print(buf)
     _, indice = buf2.sample(0)
     assert indice == [2]
--- a/test/base/test_buffer.py
+++ b/test/base/test_buffer.py
@@ -67,10 +67,11 @@ def test_stack(size=5, bufsize=9, stack_num=4):
         if done:
             obs = env.reset(1)
     indice = np.arange(len(buf))
-    assert np.allclose(buf.get(indice, 'obs'), np.array([
-        [1, 1, 1, 2], [1, 1, 2, 3], [1, 2, 3, 4],
-        [1, 1, 1, 1], [1, 1, 1, 2], [1, 1, 2, 3],
-        [3, 3, 3, 3], [3, 3, 3, 4], [1, 1, 1, 1]]))
+    assert np.allclose(buf.get(indice, 'obs'), np.expand_dims(
+        np.array([
+            [1, 1, 1, 2], [1, 1, 2, 3], [1, 2, 3, 4],
+            [1, 1, 1, 1], [1, 1, 1, 2], [1, 1, 2, 3],
+            [3, 3, 3, 3], [3, 3, 3, 4], [1, 1, 1, 1]]), axis=2))
     print(buf)
     _, indice = buf2.sample(0)
     assert indice == [2]
--- a/test/base/test_env.py
+++ b/test/base/test_env.py
@@ -18,7 +18,7 @@ def recurse_comp(a, b):
                 return np.array(
                     [recurse_comp(m, n) for m, n in zip(a, b)]).all()
             else:
-                return (a == b).all()
+                return np.allclose(a, b)
         elif isinstance(a, (list, tuple)):
             return np.array(
                 [recurse_comp(m, n) for m, n in zip(a, b)]).all()
--- a/test/base/test_env.py
+++ b/test/base/test_env.py
@@ -3,7 +3,7 @@ import numpy as np
 from gym.spaces.discrete import Discrete
 from tianshou.data import Batch
 from tianshou.env import VectorEnv, SubprocVectorEnv, \
-    RayVectorEnv, AsyncVectorEnv
+    RayVectorEnv, AsyncVectorEnv, ShmemVectorEnv
 
 if __name__ == '__main__':
     from env import MyTestEnv
@@ -62,6 +62,7 @@ def test_vecenv(size=10, num=8, sleep=0.001):
     venv = [
         VectorEnv(env_fns),
         SubprocVectorEnv(env_fns),
+        ShmemVectorEnv(env_fns)
     ]
     if verbose:
         venv.append(RayVectorEnv(env_fns))
@@ -77,9 +78,11 @@ def test_vecenv(size=10, num=8, sleep=0.001):
                 if sum(C):
                     A = v.reset(np.where(C)[0])
                 o.append([A, B, C, D])
-            for i in zip(*o):
-                for j in range(1, len(i) - 1):
-                    assert (i[0] == i[j]).all()
+            for index, infos in enumerate(zip(*o)):
+                if index == 3:
+                    continue
+                for info in infos:
+                    assert (infos[0] == info).all()
     else:
         t = [0, 0, 0]
         for i, e in enumerate(venv):
--- a/test/base/test_env.py
+++ b/test/base/test_env.py
@@ -84,7 +84,7 @@ def test_vecenv(size=10, num=8, sleep=0.001):
                 for info in infos:
                     assert (infos[0] == info).all()
     else:
-        t = [0, 0, 0]
+        t = [0]*len(venv)
         for i, e in enumerate(venv):
             t[i] = time.time()
             e.reset()
@@ -93,9 +93,8 @@ def test_vecenv(size=10, num=8, sleep=0.001):
                 if sum(done) > 0:
                     e.reset(np.where(done)[0])
             t[i] = time.time() - t[i]
-        print(f'VectorEnv: {t[0]:.6f}s')
-        print(f'SubprocVectorEnv: {t[1]:.6f}s')
-        print(f'RayVectorEnv: {t[2]:.6f}s')
+        for i, v in enumerate(venv):
+            print(str(type(v))+f': {t[i]:.6f}s')
     for v in venv:
         assert v.size == list(range(size, size + num))
         assert v.env_num == num
--- a/test/base/test_env.py
+++ b/test/base/test_env.py
@@ -55,8 +55,9 @@ def test_async_env(num=8, sleep=0.1):
 
 def test_vecenv(size=10, num=8, sleep=0.001):
     verbose = __name__ == '__main__'
+    verbose = False
     env_fns = [
-        lambda i=i: MyTestEnv(size=i, sleep=sleep)
+        lambda i=i: MyTestEnv(size=i, sleep=sleep, recurse_state=True)
         for i in range(size, size + num)
     ]
     venv = [
@@ -67,7 +68,7 @@ def test_vecenv(size=10, num=8, sleep=0.001):
     if verbose:
         venv.append(RayVectorEnv(env_fns))
     for v in venv:
-        v.seed()
+        v.seed(0)
     action_list = [1] * 5 + [0] * 10 + [1] * 20
     if not verbose:
         o = [v.reset() for v in venv]
@@ -82,6 +83,7 @@ def test_vecenv(size=10, num=8, sleep=0.001):
                 if index == 3:  # do not check info here
                     continue
                 for info in infos:
+                    print(index)
                     assert (infos[0] == info).all()
     else:
         t = [0] * len(venv)
--- a/test/base/test_env.py
+++ b/test/base/test_env.py
@@ -84,7 +84,7 @@ def test_vecenv(size=10, num=8, sleep=0.001):
                 for info in infos:
                     assert (infos[0] == info).all()
     else:
-        t = [0]*len(venv)
+        t = [0] * len(venv)
         for i, e in enumerate(venv):
             t[i] = time.time()
             e.reset()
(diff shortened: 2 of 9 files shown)


test/base/test_env.py
```diff
@@ -77,11 +78,13 @@ def test_vecenv(size=10, num=8, sleep=0.001):
...
-                for j in range(1, len(i) - 1):
-                    assert (i[0] == i[j]).all()
+            for index, infos in enumerate(zip(*o)):
+                if index == 3:
+                    continue
+                for info in infos:
+                    assert (infos[0] == info).all()
     else:
-        t = [0, 0, 0]
+        t = [0]*len(venv)
```
Trinkle23897: add space around *

test/base/test_env.py
```diff
@@ -90,9 +93,8 @@ def test_vecenv(size=10, num=8, sleep=0.001):
                 if sum(done) > 0:
                     e.reset(np.where(done)[0])
             t[i] = time.time() - t[i]
-        print(f'VectorEnv: {t[0]:.6f}s')
-        print(f'SubprocVectorEnv: {t[1]:.6f}s')
-        print(f'RayVectorEnv: {t[2]:.6f}s')
+        for i, v in enumerate(venv):
+            print(str(type(v))+f': {t[i]:.6f}s')
```
Trinkle23897: same

tianshou/env/vecenv/shmemenv.py
```diff
@@ -0,0 +1,159 @@
...
+class ShmemVectorEnv(SubprocVectorEnv):
+    """Optimized version of SubprocVectorEnv that uses shared
+    variables to communicate observations.
+
+    .. seealso::
+
+        Please refer to :class:`~tianshou.env.BaseVectorEnv` for more detailed
+        explanation.
+    """
+    """I borrowed heavily from openai baseline to implement ShmemVectorEnv Class.
```
Trinkle23897: No need to again `"""`, merge them together.

ChenDRAG: fixed

test/base/test_env.py
ChenDRAG: fixed

test/base/test_env.py
ChenDRAG: fixed

tianshou/env/vecenv/shmemenv.py
```diff
@@ -60,23 +58,23 @@ def _encode_obs(maybe_dict):
...
-    common/vec_env/shmem_vec_env.py' for more info if you are interested.
+    Please refer to "https://github.com/openai/baselines/blob/master/baselines/
+    common/vec_env/shmem_vec_env.py" for more info if you are interested.
+
                     Huayu Chen(chenhuay17@mails.tsinghua.edu.cn)  2020/08/01
     """
 
     def __init__(self, env_fns: List[Callable[[], gym.Env]]) -> None:
         BaseVectorEnv.__init__(self, env_fns)
-        self.closed = False
```
ChenDRAG: I don't think this should be deleted since SubprocVectorEnv is not initialized.

Trinkle23897: Oh my bad

tianshou/env/vecenv/shmemenv.py
```diff
@@ -0,0 +1,158 @@
...
+    """Control a single environment instance using IPC and shared memory."""
+    def _encode_obs(maybe_dict):
+        flatdict = maybe_dict if isinstance(maybe_dict, dict) else {
+            None: maybe_dict}
+        for k in keys:
+            dst = obs_bufs[k].get_obj()
+            dst_np = np.frombuffer(
+                dst, dtype=obs_dtypes[k]).reshape(obs_shapes[k])
+            np.copyto(dst_np, flatdict[k])
+        return None
```
youkaichao, line 37: return None is not necessary.

tianshou/env/vecenv/shmemenv.py
```diff
@@ -0,0 +1,158 @@
...
+    .. seealso::
+
+        Please refer to :class:`~tianshou.env.BaseVectorEnv` for more detailed
+        explanation.
+
+    I borrowed heavily from openai baseline to implement ShmemVectorEnv Class.
+    Please refer to "https://github.com/openai/baselines/blob/master/baselines/
+    common/vec_env/shmem_vec_env.py" for more info if you are interested.
+
+                    Huayu Chen(chenhuay17@mails.tsinghua.edu.cn)  2020/08/01
```
youkaichao: Not necessary. Git blame knows you did it.

tianshou/env/vecenv/shmemenv.py
```diff
@@ -0,0 +1,158 @@
...
+                         for i in range(len(obs_space.spaces))}
+        else:
+            subspaces = {None: obs_space}
+        self.obs_keys = []
+        self.obs_shapes = {}
+        self.obs_dtypes = {}
+        for key, box in subspaces.items():
+            self.obs_keys.append(key)
+            self.obs_shapes[key] = box.shape
+            self.obs_dtypes[key] = box.dtype
```
youkaichao: I think a better way is to use ``.observation_space.sample()`` to get a sample of observation. And it seems current implementation does not deal with recurse dict. Below is a code snippet you can use (decode looks the same). Then your implementation is different from open ai baseline :) 

```python
obs = env.observation_space.sample()
def encode(obs):
    if isinstance(obs, np.ndarray):
        shm = Array(xxx)
    if isinstance(obs, tuple):
        shm = (encode(i) for i in obs)
    if isinstance(obs, dict):
        shm = {k: encode(v) for k, v in obs.items()}
    return shm
shm = encode(obs)
```

tianshou/env/vecenv/shmemenv.py
ChenDRAG: got it, thanks.

tianshou/env/vecenv/shmemenv.py
ChenDRAG: Yes, but I intentionally do this for clarity. A function with no return can sometimes be confusing for people who are not that familiar with the language, especially when its "return" serves as the input of another function.
```python
        obs = np.stack(
            [self._decode_obs(self.parent_remote[i].recv(), i) for i in id])
```

tianshou/env/vecenv/shmemenv.py
ChenDRAG: Thanks a lot. Yes recurse dict is not supported, but so far I haven't seen any case in which the observation is a recurse dict. What's more, it seems that standard gym env only supports the observation to be dict of numpy  array or a single numpy array. So I think maybe it is unnecessary to change this part just to be different with openai.
sample() looks good.

Trinkle23897: I think it should be done, for consistency.

Trinkle23897:
Also need to modify the documentation somewhere.

tianshou/env/vecenv/shmemenv.py
ChenDRAG: Yeah, okay. No harm anyway. Working on it. Thanks for suggestions.

youkaichao: > but so far I haven't seen any case in which the observation is a recurse dict

Here is a case of recurse dict from the [MineRL](https://minerl.io/docs/index.html) environment: 
```python
Dict(
    inventory: Dict(
        coal: Box(low=0, high=2304, shape=()), 
        cobblestone: Box(low=0, high=2304, shape=())
    ), 
    pov: Box(low=0, high=255, shape=(64, 64, 3))
)
```

Trinkle23897:
How about #166?

tianshou/env/vecenv/shmemenv.py
duburcqa: > Thanks a lot. Yes recurse dict is not supported, but so far I haven't seen any case in which the observation is a recurse dict.

In my project, mine are too. And it is pretty common. I just you have only seen some toy examples. In real project it is quite usually.

duburcqa:
It would be nice to refactor this module, to avoid many code duplications. But it could be done in another PR.

Trinkle23897:
> It would be nice to refactor this module, to avoid many code duplications. But it could be done in another PR.

Yes, I think so. Another PR will reconstruct these vecenv codes to support wait_num.

ChenDRAG:
recurse dict is supported now. Ready for review.

> How about #166?
I think this should be done in another PR, 'cause it is not related to Shmemenv?

tianshou/env/__init__.py
```diff
@@ -11,5 +12,6 @@
     'SubprocVectorEnv',
     'AsyncVectorEnv',
     'RayVectorEnv',
-    'MultiAgentEnv',
+    'ShmemVectorEnv',
+    'MultiAgentEnv'
```
youkaichao: Keep the comma, please.

tianshou/env/vecenv/shmemenv.py
```diff
@@ -0,0 +1,175 @@
...
+             np.uint8: ctypes.c_uint8,
+             np.uint16: ctypes.c_uint16,
+             np.uint32: ctypes.c_uint32,
+             np.uint64: ctypes.c_uint64,
+             np.int8: ctypes.c_int8,
+             np.int16: ctypes.c_int16,
+             np.int32: ctypes.c_int32,
+             np.int64: ctypes.c_int64,
+             np.float32: ctypes.c_float,
+             np.float64: ctypes.c_double}
```
youkaichao: How about np.int, np.bool_, and np.float?

tianshou/env/vecenv/shmemenv.py
```diff
@@ -0,0 +1,175 @@
...
+    def _setup_buf(self, space):
+        if isinstance(space, gym.spaces.Dict):
+            assert isinstance(space.spaces, OrderedDict)
+            buffer = {k: self._setup_buf(v) for k, v in space.spaces.items()}
+        elif isinstance(space, gym.spaces.Tuple):
+            assert isinstance(space.spaces, tuple)
+            buffer = tuple([self._setup_buf(t) for t in space.spaces])
+        else:
+            buffer = ShArray(space.dtype, space.shape)
+        return buffer
```
youkaichao: This can be a static method.

tianshou/env/vecenv/shmemenv.py
```diff
@@ -0,0 +1,175 @@
...
+        def decode_obs(buffer):
+            if isinstance(buffer, ShArray):
+                return buffer.get()
+            elif isinstance(buffer, tuple):
+                return tuple([decode_obs(b) for b in buffer])
+            elif isinstance(buffer, dict):
+                return {k: decode_obs(v) for k, v in buffer.items()}
+            else:
+                raise NotImplementedError
+        assert isNone is None
```
youkaichao: I'm puzzled. What does this mean? Why use an argument that is always None?

ChenDRAG: Like 'return None' in encoder, to help people understand that this is a 'fake encoder decoder system', which only buffer the data but not really handle them.

youkaichao: > Like 'return None' in encoder, to help people understand that this is a 'fake encoder decoder system', which only buffer the data but not really handle them.

That's crazy, don't do that. This is an object, ``self.decode(i)`` means to decode something inside ``self``, people all know that.

tianshou/env/vecenv/shmemenv.py
ChenDRAG: np.int == int and is not supported by gym(np.float the same)

tianshou/env/vecenv/shmemenv.py
youkaichao: Okay, I know it. This parameter is filled by ``recv()`` which returns None. Okay, then.

ChenDRAG: I don't think this is inappropriate, typically decoder receives an object, decodes it return this object. In the current scenario, decoders don't actually have to receive anything. (If it only receives i, this should really be called something like get_buffer_from, not 'decoder').
assert None is a little bit redundant， I can remove that.

test/base/test_buffer.py
```diff
@@ -67,10 +67,10 @@ def test_stack(size=5, bufsize=9, stack_num=4):
...
             obs = env.reset(1)
     indice = np.arange(len(buf))
-    assert np.allclose(buf.get(indice, 'obs'), np.array([
-        [1, 1, 1, 2], [1, 1, 2, 3], [1, 2, 3, 4],
-        [1, 1, 1, 1], [1, 1, 1, 2], [1, 1, 2, 3],
-        [3, 3, 3, 3], [3, 3, 3, 4], [1, 1, 1, 1]]))
+    assert np.allclose(buf.get(indice, 'obs'), np.expand_dims(
+        [[1, 1, 1, 2], [1, 1, 2, 3], [1, 2, 3, 4],
+         [1, 1, 1, 1], [1, 1, 1, 2], [1, 1, 2, 3],
+         [3, 3, 3, 3], [3, 3, 3, 4], [1, 1, 1, 1]], axis=-1))
```
youkaichao, line 73: Is there a better way to do this? I see many ``expand_dims ``. Same below.

test/base/test_env.py
```diff
@@ -3,14 +3,32 @@
...
 
 
+def recurse_comp(a, b):
+    try:
+        if isinstance(a, np.ndarray):
+            if a.dtype == np.object:
+                return np.array(
+                    [recurse_comp(m, n) for m, n in zip(a, b)]).all()
+            else:
+                return (a == b).all()
```
youkaichao: Use ``np.allclose``

test/base/test_buffer.py
ChenDRAG: This is what @Trinkle23897 suggested,  any suggestions?

Trinkle23897: Since using an int as a state is not valid in a standard gym.Env, the state should be a np.ndarray including only 1 element: `np.array([1])`.

youkaichao (approved):
Nice PR for memory efficiency and speed with large observations 👍

Trinkle23897 (approved):

PR #178, opened 2020-08-04
fix #166, assure close in vec env

### LINKED ISSUE #166: SubprocVectorEnv is not released when collector is closed

ChenDRAG:
fix #166 , SubprocVectorEnv and its subclasses cannot be called once closed.

### DIFF
--- a/tianshou/env/vecenv/subproc.py
+++ b/tianshou/env/vecenv/subproc.py
@@ -122,4 +122,4 @@ class SubprocVectorEnv(BaseVectorEnv):
 
     def _assert_is_closed(self):
         assert not self.closed, \
-             f"Methods of {type(self)} should not be called after closed."
+             f"Methods of {self.__class__.__name__} should not be called after closed."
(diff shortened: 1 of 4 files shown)


tianshou/env/vecenv/subproc.py
```diff
@@ -113,3 +119,7 @@ def close(self) -> List[Any]:
         for p in self.processes:
             p.join()
         return result
+
+    def _assert_is_closed(self):
+        assert not self.closed, \
+             f"Methods of {type(self)} should not be called after closed."
```
Trinkle23897: How about `{self.__class__.__name__}`?

Trinkle23897:
resolved in #179

PR #179, opened 2020-08-04
code refactor for venv

### LINKED ISSUE #103: Async Sampling

### LINKED ISSUE #166: SubprocVectorEnv is not released when collector is closed

youkaichao:
- Refacor code to remove duplicate code

- Enable async simulation for all vector envs

- Remove `collector.close` and rename `VectorEnv` to `DummyVectorEnv`

The abstraction of vector env changed.

Prior to this pr, each vector env is almost independent.

After this pr, each env is wrapped into a worker, and vector envs differ with their worker type. In fact, users can just use ``BaseVectorEnv`` with different workers, I keep ``SubprocVectorEnv``, ``ShmemVectorEnv`` for backward compatibility.


test/throughput/test_collector_profile.py
```diff
@@ -86,7 +86,7 @@ def data():
 def test_init(data):
     for _ in range(5000):
         c = Collector(data["policy"], data["env"], data["buffer"])
-        c.close()
+    c.close()
```
youkaichao: These collectors share the same vec env, so they can be closed only once.

tianshou/env/venvs.py
```diff
@@ -0,0 +1,344 @@
...
+        return f(self, *args, **kwargs)
+    return wrapper
+
+
+def generate_id():
+    generate_id.i += 1
+    return generate_id.i
+
+
+generate_id.i = 0
```
Trinkle23897: why this

docs/tutorials/cheatsheet.rst
```diff
@@ -31,7 +31,7 @@ See :ref:`customized_trainer`.
 Parallel Sampling
 -----------------
 
-Use :class:`~tianshou.env.VectorEnv`, :class:`~tianshou.env.SubprocVectorEnv` or :class:`~tianshou.env.ShmemVectorEnv`.
+Use :class:`~tianshou.env.DummyVectorEnv`, :class:`~tianshou.env.SubprocVectorEnv`, :class:`~tianshou.env.ShmemVectorEnv`, or :class:`~tianshou.env.RayVectorEnv`.
```
Trinkle23897: should mention the async feature here

tianshou/env/worker/subproc.py
```diff
@@ -0,0 +1,190 @@
...
+                return {k: decode_obs(v) for k, v in buffer.items()}
+            else:
+                raise NotImplementedError
+        return decode_obs(self.buffer)
+
+    def render(self, **kwargs) -> None:
+        self.parent_remote.send(['render', kwargs])
+        return self.parent_remote.recv()
+
+    def close(self) -> Any:
```
ChenDRAG: Why remove self.closed? Shouldn't we assert whether the processes are closed before calling reset, step, etc, as discussed in #178, #166? Otherwise calling those methods while closed will raise BrokenPipe Error.

ChenDRAG:
This refactor is quite neat and surely will work, but I don't quite get the initial motivation of substituting 'independent' envs with workers. Because it seems that the organization of current workers is quite similar with original envs, and there seems to be no less code. Could you help me understand?

>  I keep SubprocVectorEnv, ShmemVectorEnv for backward compatibility.


A potential problem: Do we still need to keep this compatibility in the future once applying this refactor? Cause if not, I think we also need to change calling methods of Basevec in every other file to show the 'worker' system is officially supported. Otherwise, we should keep backward compatibility each time we implement a vecenv, which I'm not sure whether it is a little redundant. 

Also, could you please profile the collector code to compare with the latest dev commit, I go to see GitHub action result but seems that the server is not stable and cannot provide reasonable results.

tianshou/env/worker/subproc.py
youkaichao: I think people do not have to explicitly close environments. If they want to re-use them, just reset those environments; otherwise, just leave them there and garbage collection (``__del__``) will do the trick (makes sure the processes will be terminated.).

youkaichao:
> but I don't quite get the initial motivation of substituting 'independent' envs with workers.

This pr makes ``wait_num`` (async simulation ) an internal ability of vector environments. Prior to this pr, async simulation only works for subproc environments. What's more, you may notice that duplicate code is reduced as much as possible (e.g. subproc and shmem) with the abstraction of worker.

> Do we still need to keep this compatibility in the future once applying this refactor?

Compatibility should be considered. Tianshou has many users now, we cannot break anything suddenly. All users of Tianshou uses vector environments, I think. So compatibility is important.

> Also, could you please profile the collector code to compare with the latest dev commit

You are welcome to do this. I mainly changed the internal implementation. I think efficiency does not change much.

docs/tutorials/cheatsheet.rst
```diff
@@ -48,6 +48,8 @@ where ``env_fns`` is a list of callable env hooker. The above code can be writte
     env_fns = [lambda x=i: MyTestEnv(size=x) for i in [2, 3, 4, 5]]
     venv = SubprocVectorEnv(env_fns)
 
+All of these VectorEnv class have async mode (related to issue #103), where we can give it an extra parameter ``wait_num``. If we have 4 envs and set ``wait_num = 3``, in each of the step in VectorEnv it only returns 3 results of these 4 envs. This mode eases the case where each step cost vary different time， e.g. 90% step cost 1s, but 10% cost 10s.
```
youkaichao: "All of these VectorEnv class" --> "all subclasses of ``VectorEnv ``"

tianshou/env/venvs.py
```diff
@@ -98,7 +76,11 @@ def __init__(self,
         self.waiting_id = []
         # all environments are ready in the beginning
         self.ready_id = list(range(self.env_num))
-        self.unique_id = generate_id()
+        self.is_closed = False
+
+    def _assert_is_closed(self):
```
youkaichao: Func name is ``_assert_is_closed ``, but the assertion is ``assert not self.is_closed``

youkaichao:
```suggestion
    def _assert_is_not_closed(self):
```

docs/tutorials/cheatsheet.rst
Trinkle23897:
```suggestion
All subclasses of ``VectorEnv`` have an async mode (related to issue #103), where we can give it an extra parameter ``wait_num``. If we have 4 envs and set ``wait_num = 3``, each of the step in VectorEnv only returns 3 results of these 4 envs. This mode eases the case where each step cost varies at different timescale， e.g. 90% step cost 1s, but 10% cost 10s.
```

duburcqa:
Just for you to now, I'm reviewing right now :smiley_cat: That's kind of a big PR!

youkaichao:
> Just for you to now, I'm reviewing right now 😺 That's kind of a big PR!

Take it easy😁 the diff is +856 −895 lines. Mostly this is about code refactor and moving some code around.

duburcqa:
> Mostly this is about code refactor and moving some code around.

Indeed, it was not as big as I thought x)

Nice job, I think it is really making Tianshou move forward. The legacy implementation had become really difficult to maintain and to extend.

tianshou/env/venvs.py
```diff
@@ -0,0 +1,319 @@
...
+        is disabled; else, ``1 <= wait_num <= env_num``.
+    """
+
+    def __init__(self,
+                 env_fns: List[Callable[[], gym.Env]],
+                 worker_fn: Callable[[Callable[[], gym.Env]], EnvWorker],
+                 wait_num: Optional[int] = None,
+                 ) -> None:
+        self._env_fns = env_fns
+        self.workers = [worker_fn(fn) for fn in env_fns]
```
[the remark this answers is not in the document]
Trinkle23897: `worker_fn` can be a lambda expression, you can see at the definition of `SubprocVectorEnv` and `ShmemVectorEnv`.

tianshou/env/worker/base.py
```diff
@@ -0,0 +1,71 @@
+import gym
+import numpy as np
+from abc import ABC, abstractmethod
+from typing import List, Tuple, Optional, Callable, Any
+
+
+class EnvWorker(ABC, gym.Env):
```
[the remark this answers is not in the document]
Trinkle23897: Not a pool. A worker here only interacts with one environment and the `VectorEnv` class is the pool for handling many workers.

tianshou/env/venvs.py
```diff
@@ -0,0 +1,319 @@
...
+                self._assert_id(id)
+                assert len(action) == len(id)
+                for i, (act, env_id) in enumerate(zip(action, id)):
+                    self.workers[env_id].send_action(act)
+                    self.waiting_conn.append(self.workers[env_id])
+                    self.waiting_id.append(env_id)
+                self.ready_id = [x for x in self.ready_id if x not in id]
+            result = []
+            while len(self.waiting_conn) > 0 and len(result) < self.wait_num:
+                ready_conns = self.worker_class.wait(self.waiting_conn)
```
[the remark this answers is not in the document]
Trinkle23897: `waiting_conn` is a list, not a class.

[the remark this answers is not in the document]
Trinkle23897: But `self.waiting_conn` is a list of Worker which cannot directly apply the `wait` method.

magicly:
very nice work, thanks @youkaichao 

I find the `wait` method doesn't have a `timeout` parameter,  so collector must wait for the exactly `wait_num` env finished before next `step`.  However, in some envs, different `step` time may vary so much that it's very hard to set a proper `wait_num`. And the network forward time for a batch seems to be nearly constant,  so I think it may be easier to set a proper `timeout` parameter. 

`wait` method of both `subproc` and `ray` have support for `timeout` parameter, so I think it needs just a little more work to match the semantics of `ray.wait`.

https://docs.ray.io/en/latest/package-ref.html?highlight=wait#ray.wait

> If timeout is set, the function returns either when the requested number of IDs are ready or when the timeout is reached, whichever occurs first. If it is not set, the function simply waits until that number of objects is ready and returns that exact number of object IDs.

tianshou/env/venvs.py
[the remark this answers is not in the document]
youkaichao: The ``wait`` interface is to wait a list of pending objects and to return a list of objects who are ready. This is async simulation, so we cannot let each individual connection to wait, we have to wait them all, and collect those ready ones.

tianshou/env/venvs.py
```diff
@@ -0,0 +1,319 @@
...
+
+    def __init__(self,
+                 env_fns: List[Callable[[], gym.Env]],
+                 worker_fn: Callable[[Callable[[], gym.Env]], EnvWorker],
+                 wait_num: Optional[int] = None,
+                 ) -> None:
+        self._env_fns = env_fns
+        self.workers = [worker_fn(fn) for fn in env_fns]
+        self.env_num = len(env_fns)
+        self.worker_class = self.workers[0].__class__
```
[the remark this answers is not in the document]
youkaichao: An alternative solution is to use a separate ``wait`` function and check the type of arguments to determine how to wait. But I'm afraid the type check overhead is large, because ``wait`` is a fundamental operator during async simulation (it will be called many, many times).

youkaichao:
> However, in some envs, different step time may vary so much that it's very hard to set a proper wait_num

@magicly  Yeah, timeout is a default functionality for the ``wait`` interface. I do not implement it because I think there will always be an optimal ``wait_num``. I do not understand why it's very hard to set a proper wait_num. Can you give an example?

tianshou/env/venvs.py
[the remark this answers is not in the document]
duburcqa: Clearly. But it is not possible to invoke `self.workers[i].wait()` directly ?

tianshou/env/worker/base.py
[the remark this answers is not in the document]
duburcqa: Yes I noticed that after writing my comment and I thought it was better to get it as is, just for you to know I was confused in the first place x) Meaning that some comments are missing in the code, the implementation is not self explanatory, or I'm just sleepy or slow minded :)

tianshou/env/venvs.py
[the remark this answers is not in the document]
duburcqa: OK I see the issue. `wait` should be a method of of `VectorEnv` itself rather than the worker(s) from a design point of view. Howerver, iits implementation is worker-specific,a nd because of parallelism, it is not impossible to simply loop of each workers using a standard `for-loop`. I have to think about it, but I don't like the idea to ask to one worker to wait for its friends. It should be the pool itself to handle such high level behavior requiring awareness of the co-workers. I'm thinking how to handle it properly.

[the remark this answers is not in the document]
youkaichao: > I don't like the idea to ask to one worker to wait for its friends

It is not one worker waiting for its friends. It is a static function of ``self.worker_class`` to wait a list of workers with the same type. Let the pool handle it is ok, but I'm concerned about the performance. See [my comment](https://github.com/thu-ml/tianshou/pull/179#discussion_r468045854)

tianshou/env/worker/base.py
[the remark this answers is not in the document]
youkaichao: Yes, it is somewhat confusing at first glance. Currently there is a comment saying this is ``An abstract worker for an environment.``, maybe we should emphasis that this is a worker for **an** environment.

tianshou/env/venvs.py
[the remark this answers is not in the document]
duburcqa: >  It is a static function of self.worker_class to wait a list of workers with the same type.

Yes I know, but it still feels like it is the workers that are handling it at the pool level. But I get your concern and indeed it must be avoided. Maybe using a Metaclass pattern could do the trick. I'm thinking.

tianshou/env/worker/base.py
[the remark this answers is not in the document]
duburcqa: No at worker level it is fine (or at least, fine enough). I rather got confused to VecEnv. The information that is missing is that VecEnv is not a usual pool, in a sense that it does not really handle parallelism by itself. Being asynchronous or not is to the resort of the sole worker and at its discretion, with no explicit mechanism for the pool to know this information in advance (hence the unusual `wait` mechanism). And this is what is actually confusing, since at first glance vecenv looks like a vector of pool more than a pool itself (and with the current implement, it totally could be) ! This point should be explained very clearly in the documentation.

magicly:
> I do not understand why it's very hard to set a proper wait_num. Can you give an example?

@youkaichao I think the reason why we need `async` mode is  different `step` may cost much different time, so the `fast step` must wait for `slow step`.  It is `fast/slow step` not `fast/slow env`.

So e.g. there are 10 envs, and at start every step cost 1ms, so the best `wait_num` is 10. Then after some time, 5 envs step time is 2ms, and the other 5 envs is 5ms, so I think the best `wait_num` is 5.  maybe after some time, 10 envs step time is so much different, e.g. 5ms, 10ms, 15ms, ..., 50ms, then the best `wait_num` is 1. 

So I think a timeout parameter is a good choice for such `dynamic` env.

youkaichao:
> So I think a timeout parameter is a good choice for such dynamic env.

Good point! Previously I only thought about static environments (the distribution of time cost of each step stays the same). It is true that timeout should be supported for dynamic environments. In fact, action distribution depends on policy, which may change the time cost distribution of step functions. 

So yes, ``timeout`` will be supported. Stay tuned!

tianshou/env/venvs.py
```diff
@@ -0,0 +1,334 @@
...
+                ready_conns = self.worker_class.wait(
+                    self.waiting_conn, self.timeout)
+                for conn in ready_conns:
+                    waiting_index = self.waiting_conn.index(conn)
+                    self.waiting_conn.pop(waiting_index)
+                    env_id = self.waiting_id.pop(waiting_index)
+                    obs, rew, done, info = conn.get_result()
+                    info["env_id"] = env_id
+                    result.append((obs, rew, done, info))
+                    self.ready_id.append(env_id)
```
magicly:
```suggestion
            ready_conns = self.worker_class.wait(
                self.waiting_conn, self.wait_num, self.timeout)
            for conn in ready_conns:
                waiting_index = self.waiting_conn.index(conn)
                self.waiting_conn.pop(waiting_index)
                env_id = self.waiting_id.pop(waiting_index)
                obs, rew, done, info = conn.get_result()
                info["env_id"] = env_id
                result.append((obs, rew, done, info))
                self.ready_id.append(env_id)
```

tianshou/env/worker/base.py
```diff
@@ -0,0 +1,69 @@
...
+        so typically users only call ``step`` function. But they can be called
+        separately in async simulation, i.e. someone calls ``send_action``
+        first, and calls ``get_result`` later.
+        """
+        self.send_action(action)
+        return self.get_result()
+
+    @staticmethod
+    def wait(workers: List['EnvWorker'],
+             timeout: Optional[float] = None) -> List['EnvWorker']:
```
magicly, line 43:
```suggestion
             wait_num: int,
             timeout: Optional[float] = None) -> List['EnvWorker']:
```

tianshou/env/worker/ray.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def __getattr__(self, key: str):
+        return ray.get(self.env.__getattr__.remote(key))
+
+    def reset(self) -> Any:
+        return ray.get(self.env.reset.remote())
+
+    @staticmethod
+    def wait(workers: List['RayEnvWorker'],
+             timeout: Optional[float] = None) -> List['RayEnvWorker']:
```
magicly, line 29:
```suggestion
             wait_num: int,
             timeout: Optional[float] = None) -> List['RayEnvWorker']:
```

tianshou/env/worker/ray.py
```diff
@@ -0,0 +1,54 @@
...
+        ready_envs = []
+        while not ready_envs:
+            ready_envs, _ = ray.wait([x.env for x in workers],
+                                     num_returns=len(workers), timeout=timeout)
```
magicly:
```suggestion
                                     num_returns=wait_num, timeout=timeout)
```

tianshou/env/worker/subproc.py
```diff
@@ -0,0 +1,190 @@
+import gym
+import ctypes
+import numpy as np
```
magicly, line 4:
```suggestion
import numpy as np
import time
```

tianshou/env/worker/subproc.py
```diff
@@ -0,0 +1,190 @@
...
+        if self.share_memory:
+            obs = self._decode_obs(obs)
+        return obs
+
+    @staticmethod
+    def wait(workers: List['SubprocEnvWorker'],
+             timeout: Optional[float] = None) -> List['SubprocEnvWorker']:
+        conns, ready_conns = [x.parent_remote for x in workers], []
+        while not ready_conns:
+            ready_conns = connection.wait(conns, timeout=timeout)
```
magicly:
```suggestion
             wait_num: int,
             timeout: Optional[float] = None) -> List['SubprocEnvWorker']:
        conns = [x.parent_remote for x in workers]
        ready_conns = []
        remain_conns = conns
        t1 = time.time()
        while len(remain_conns) > 0 and len(ready_conns) < wait_num:
            if timeout:
                remain_time = timeout - (time.time() - t1)
                if remain_time <= 0:
                    break
            else:
                remain_time = timeout
            remain_conns = [conn for conn in remain_conns if conn not in ready_conns]
            new_ready_conns = connection.wait(remain_conns, timeout=remain_time)
            ready_conns.extend(new_ready_conns)
```

tianshou/env/worker/dummy.py
```diff
@@ -0,0 +1,40 @@
...
+        if hasattr(self.env, key):
+            return getattr(self.env, key)
+        return None
+
+    def reset(self) -> Any:
+        return self.env.reset()
+
+    @staticmethod
+    def wait(workers: List['DummyEnvWorker'],
+             timeout: Optional[float] = None) -> List['DummyEnvWorker']:
```
magicly, line 26:
```suggestion
             wait_num: int,
             timeout: Optional[float] = None) -> List['DummyEnvWorker']:
```

magicly (commented):
to make `EnvWorker.wait` method match the semantics with `ray.wait` method:

https://docs.ray.io/en/latest/package-ref.html?highlight=wait#ray.wait

> If timeout is set, the function returns either when the requested number of IDs are ready or when the timeout is reached, whichever occurs first. If it is not set, the function simply waits until that number of objects is ready and returns that exact number of object refs.

tianshou/env/venvs.py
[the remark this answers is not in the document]
youkaichao: Any progress?

youkaichao:
I think the ``wait`` should honestly return an empty list if there is not a single finished step. The disposal of empty return is responsible to ``venv``.

test/base/test_env.py
```diff
@@ -124,5 +147,6 @@ def test_vecenv(size=10, num=8, sleep=0.001):
 
 
 if __name__ == '__main__':
-    test_vecenv()
-    test_async_env()
+    # test_vecenv()
```
Trinkle23897: TODO: revert

test/base/test_env.py
```diff
@@ -11,6 +11,14 @@
     from test.base.env import MyTestEnv
 
 
+def has_ray():
+    try:
+        import ray
+        return hasattr(ray, 'init')
```
youkaichao:
```suggestion
        # bypass pep8 check
        return hasattr(ray, 'init')
```

docs/tutorials/dqn.rst
```diff
@@ -30,11 +30,11 @@ It is available if you want the original ``gym.Env``:
     train_envs = gym.make('CartPole-v0')
     test_envs = gym.make('CartPole-v0')
 
-Tianshou supports parallel sampling for all algorithms. It provides four types of vectorized environment wrapper: :class:`~tianshou.env.VectorEnv`, :class:`~tianshou.env.SubprocVectorEnv`, :class:`~tianshou.env.ShmemVectorEnv`, and :class:`~tianshou.env.RayVectorEnv`. It can be used as follows: 
+Tianshou supports parallel sampling for all algorithms. It provides four types of vectorized environment wrapper: :class:`~tianshou.env.DummyVectorEnv`, :class:`~tianshou.env.SubprocVectorEnv`, :class:`~tianshou.env.ShmemVectorEnv`, and :class:`~tianshou.env.RayVectorEnv`. It can be used as follows:
```
youkaichao: I think a pointer to parallel sampling is enough. No need to elaborate on this in DQN.

duburcqa:
> It is true that timeout should be supported for dynamic environments. In fact, action distribution depends on policy, which may change the time cost distribution of step functions.

I would add that in m'y opinion it is usually more common to deal with such dynamic environment than a static one. For example in my case I have to deal with dynamic environments, never static. It comes from the way the state is integrated over time. Some state requires to refine the computation to guarantee the accuracy, while for others it is not necessary.

duburcqa:
Is it ready for review ? I didn't spent much time thinking to a better way to handle methods such as `wait` to avoid relying on static methods of the workers themself. I'm suite confident that using metaclass (in a way similar to metaprograming in c++) could do the trick but it is tricky to design and lack for clarity even for devs. So I don't know what to think about this.

Trinkle23897:
> Is it ready for review?

Yes, of course!

And I have another concern: can we reduce the async overhead as much as possible? I tested with wait_num == env_num for both sync and async mode (use test_drqn.py), and the result shows that the speed of async is 80% of sync. This could be the future work.

test/base/test_env.py
```diff
@@ -29,79 +37,110 @@ def recurse_comp(a, b):
...
+            [0, 2, 3],
+            [0, 1],
+        ]
+        ids = np.arange(num)
+        for res in expect_result:
+            t = time.time()
+            _, _, _, info = v.step([1] * len(ids), ids)
+            ids = Batch(info).env_id
+            print(ids, time.time() - t)
+            if cls != RayVectorEnv:
```
Trinkle23897: related to https://github.com/ray-project/ray/issues/10134

youkaichao:
> I would add that in m'y opinion it is usually more common to deal with such dynamic environment than a static one.

Agree. Only toy environments are static. Almost all realistic environments are dynamic.

duburcqa:
I'll have a look later today

tianshou/env/worker/subproc.py
```diff
@@ -0,0 +1,204 @@
...
+        self.process = Process(target=_worker, args=args, daemon=True)
+        self.process.start()
+        self.child_remote.close()
+
+    def __getattr__(self, key: str):
+        self.parent_remote.send(['getattr', key])
+        return self.parent_remote.recv()
+
+    @staticmethod
+    def _setup_buf(space):
```
duburcqa: What it is point of this method ? It is meaningful to make it accessible to the user or it can only be used internally ? If the later, then imho it would not be able a nested method of `__init__`.

tianshou/env/venvs.py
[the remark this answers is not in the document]
duburcqa: I reply a few days ago, but it seems my comment was never published for some reason. Basically I said that it would be possible using something similar to template meta-programming in C++, but it would make the code significantly more complicated to understand and I think it is not a good idea.

duburcqa (dismissed):

duburcqa:
As I said, I should be clear from the documentation that the vector env are not usual pool, a sense that it does not really handle parallelism directly. Being asynchronous or not is to the resort of the sole worker and at its discretion, with no explicit mechanism for the pool to know this information in advance (hence the unusual wait mechanism).

youkaichao:
> Being asynchronous or not is to the resort of the sole worker and at its discretion, with no explicit mechanism for the pool to know this information in advance

I don't agree. The workers have ``send_action`` / ``get_results`` interfaces so that they support asynchronous simulation. It is up to the user / vec env whether to use asynchronous or not. I don't think "the unusual wait mechanism" exists.

Trinkle23897 (approved):

duburcqa:
> The workers have send_action / get_results interfaces so that they support asynchronous simulation. It is up to the user / vec env whether to use asynchronous or not. 

No it is not. It requires the user to use a worker  that support async behavior. It is not asynchronous by itself. 

> I don't think "the unusual wait mechanism" exists.

Are you kidding me ? Show me a single well-known library using the same "calling static method from workers" strategy to implement `wait` method.

youkaichao:
> Are you kidding me ? Show me a single well-known library using the same "calling static method from workers" strategy to implement wait method.

I admit the current implementation requires the hack of calling static method from workers.

To remove the hack, each worker should implement a ``ready`` interface, which returns ``True`` if it is ready to return results. But I don't know how to implement it for ray, and I think it would be difficult to implement. Therefore, I rely on the readily implemented ``connection.wait`` and ``ray.wait``.

The ``ready`` interface enables waiting by while-loop. An efficient wait should rely on something like semaphore, i.e. sleep on a set of semaphores. But I doubt something general like sleeping on a set of semaphores exists. Most likely, sleeping on some signal is worker-dependent.

duburcqa:
I agree it is tricky to implement and I'm fine with the current implementation. It could be refactor later if necessary, or if we have time for that.

PR #180, opened 2020-08-06
add policy.update to enable post process and remove collector.sample

youkaichao:
In reply to [this comment](https://github.com/thu-ml/tianshou/pull/159#discussion_r464512834).

PR #182, opened 2020-08-12
Pickle compatible for replay buffer and improve buffer.get

### LINKED ISSUE #84: save/load data in replay buffer

Trinkle23897:
fix #84 and make buffer more efficient

### DIFF
--- a/test/base/test_buffer.py
+++ b/test/base/test_buffer.py
@@ -1,3 +1,4 @@
+import torch
 import pytest
 import numpy as np
 from timeit import timeit
@@ -219,13 +220,16 @@ def test_pickle():
     vbuf = ReplayBuffer(size, stack_num=2)
     lbuf = ListReplayBuffer()
     pbuf = PrioritizedReplayBuffer(size, 0.6, 0.4)
+    device = 'cuda' if torch.cuda.is_available() else 'cpu'
+    rew = torch.tensor([1.]).to(device)
+    print(rew)
     for i in range(4):
-        vbuf.add(obs=Batch(index=np.array([i])), act=0, rew=1, done=0)
+        vbuf.add(obs=Batch(index=np.array([i])), act=0, rew=rew, done=0)
     for i in range(3):
-        lbuf.add(obs=Batch(index=np.array([i])), act=1, rew=1, done=0)
+        lbuf.add(obs=Batch(index=np.array([i])), act=1, rew=rew, done=0)
     for i in range(5):
         pbuf.add(obs=Batch(index=np.array([i])),
-                 act=2, rew=1, done=0, weight=np.random.rand())
+                 act=2, rew=rew, done=0, weight=np.random.rand())
     # save
     vbuf.save('/tmp/vbuf.pkl')
     lbuf.save('/tmp/lbuf.pkl')
--- a/test/base/test_buffer.py
+++ b/test/base/test_buffer.py
@@ -2,8 +2,8 @@ import pytest
 import numpy as np
 from timeit import timeit
 
-from tianshou.data import Batch, PrioritizedReplayBuffer, \
-    ReplayBuffer, SegmentTree
+from tianshou.data import Batch, SegmentTree, \
+    ReplayBuffer, ListReplayBuffer, PrioritizedReplayBuffer
 
 if __name__ == '__main__':
     from env import MyTestEnv
@@ -214,10 +214,45 @@ def test_segtree():
         print('tree', timeit(sample_tree, setup=sample_tree, number=1000))
 
 
+def test_pickle():
+    size = 100
+    vbuf = ReplayBuffer(size, stack_num=2)
+    lbuf = ListReplayBuffer()
+    pbuf = PrioritizedReplayBuffer(size, 0.6, 0.4)
+    for i in range(4):
+        vbuf.add(obs=Batch(index=np.array([i])), act=0, rew=1, done=0)
+    for i in range(3):
+        lbuf.add(obs=Batch(index=np.array([i])), act=1, rew=1, done=0)
+    for i in range(5):
+        pbuf.add(obs=Batch(index=np.array([i])),
+                 act=2, rew=1, done=0, weight=np.random.rand())
+    # save
+    vbuf.save('/tmp/vbuf.pkl')
+    lbuf.save('/tmp/lbuf.pkl')
+    pbuf.save('/tmp/pbuf.pkl')
+    # normal load
+    _vbuf = ReplayBuffer.load('/tmp/vbuf.pkl')
+    _lbuf = ListReplayBuffer.load('/tmp/lbuf.pkl')
+    _pbuf = PrioritizedReplayBuffer.load('/tmp/pbuf.pkl')
+    assert len(_vbuf) == len(vbuf) and np.allclose(_vbuf.act, vbuf.act)
+    assert len(_lbuf) == len(lbuf) and np.allclose(_lbuf.act, lbuf.act)
+    assert len(_pbuf) == len(pbuf) and np.allclose(_pbuf.act, pbuf.act)
+    # make sure the meta var is identical
+    assert _vbuf.stack_num == vbuf.stack_num
+    assert np.allclose(_pbuf.weight[np.arange(len(_pbuf))],
+                       pbuf.weight[np.arange(len(pbuf))])
+    # load data by inconsistent class will raise an error
+    with pytest.raises(AssertionError):
+        ReplayBuffer.load('/tmp/lbuf.pkl')
+    with pytest.raises(AssertionError):
+        PrioritizedReplayBuffer.load('/tmp/vbuf.pkl')
+
+
 if __name__ == '__main__':
     test_replaybuffer()
     test_ignore_obs_next()
     test_stack()
+    test_pickle()
     test_segtree()
     test_priortized_replaybuffer()
     test_priortized_replaybuffer(233333, 200000)
--- a/test/base/test_buffer.py
+++ b/test/base/test_buffer.py
@@ -1,4 +1,5 @@
 import torch
+import pickle
 import pytest
 import numpy as np
 from timeit import timeit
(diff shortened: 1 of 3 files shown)


test/base/test_buffer.py
```diff
@@ -214,10 +216,43 @@ def sample_tree():
...
+    for i in range(4):
+        vbuf.add(obs=Batch(index=np.array([i])), act=0, rew=rew, done=0)
+    for i in range(3):
+        lbuf.add(obs=Batch(index=np.array([i])), act=1, rew=rew, done=0)
+    for i in range(5):
+        pbuf.add(obs=Batch(index=np.array([i])),
+                 act=2, rew=rew, done=0, weight=np.random.rand())
+    # save
+    pickle.dump(vbuf, open('/tmp/vbuf.pkl', 'wb'))
+    pickle.dump(lbuf, open('/tmp/lbuf.pkl', 'wb'))
```
youkaichao:
```suggestion
    ls = pickle.dumps(lbuf)
```

test/base/test_buffer.py
```diff
@@ -214,10 +216,43 @@ def sample_tree():
...
+    pickle.dump(pbuf, open('/tmp/pbuf.pkl', 'wb'))
+    # load
+    _vbuf = pickle.load(open('/tmp/vbuf.pkl', 'rb'))
+    _lbuf = pickle.load(open('/tmp/lbuf.pkl', 'rb'))
```
youkaichao:
```suggestion
    _lbuf = pickle.loads(ls)
```

youkaichao: Test both ``dump`` and ``dumps``

test/base/test_buffer.py
```diff
@@ -214,10 +216,39 @@ def sample_tree():
...
 
 
+def test_pickle():
+    size = 100
+    vbuf = ReplayBuffer(size, stack_num=2)
+    lbuf = ListReplayBuffer()
+    pbuf = PrioritizedReplayBuffer(size, 0.6, 0.4)
+    device = 'cuda' if torch.cuda.is_available() else 'cpu'
+    rew = torch.tensor([1.]).to(device)
+    print(rew)
```
youkaichao: Why print it?
```suggestion
```

youkaichao (dismissed):

tianshou/data/buffer.py
```diff
@@ -54,6 +55,11 @@ class ReplayBuffer:
         >>> batch_data, indice = buf.sample(batch_size=4)
         >>> batch_data.obs == buf[indice].obs
         array([ True,  True,  True,  True])
+        >>> len(buf)
+        13
+        >>> buf = pickle.load(open('buf.pkl', 'rb'))  # load from "buf.pkl"
```
youkaichao, line 60:
```suggestion
        >>> len(buf)
        13
        >>> buf = pickle.load(open('buf.pkl', 'rb'))  # load from "buf.pkl"
```

tianshou/data/buffer.py
```diff
@@ -141,6 +147,14 @@ def __getattr__(self, key: str) -> Union['Batch', Any]:
         """Return self.key"""
         return self._meta.__dict__[key]
 
+    def __setstate__(self, state):
+        """unpickling interface"""
+        self.__dict__.update(state)
```
youkaichao, line 152:
```suggestion
    def __setstate__(self, state):
        """unpickling interface"""
        self.__dict__.update(state)
```

tianshou/data/buffer.py
```diff
@@ -141,6 +147,14 @@ def __getattr__(self, key: str) -> Union['Batch', Any]:
         """Return self.key"""
         return self._meta.__dict__[key]
 
+    def __setstate__(self, state):
```
duburcqa, line 154: Are you sure it is not working out-of-the-box ?

Trinkle23897: What do you mean? I did so because I think it is not the best to re-initialize the buffer, because there are so many metavar to be restored (unlike Batch).

duburcqa: usually, it is not necessary to define this can of interface for Pickle to work, unless you need to do some fancy stuff in it. And from what I'm saying, what you are doing looks like the default behaviour, but I may be wrong.

duburcqa: > I did so because I think it is not the best to re-initialize the buffer, because there are so many metavar to be restored (unlike Batch).

Yes and your are right. But this is what pickle should be doing by default. It should not call the constructor (`__init__`), not even `__new__` if I remember well.

Trinkle23897: > usually, it is not necessary to define this can of interface for Pickle to work, unless you need to do some fancy stuff in it. And from what I'm saying, what you are doing looks like the default behavior, but I may be wrong.

Yes, you are right. I fix the existing `buffer.__getattr__` to exclude `self._meta["__getstate__"]` happen and it works without `__getstate__` and `__setstate__`.

tianshou/data/buffer.py
```diff
@@ -145,16 +145,10 @@ def __repr__(self) -> str:
 
     def __getattr__(self, key: str) -> Union['Batch', Any]:
         """Return self.key"""
+        if key.startswith('__'):  # since we do not use key begin with "__"
```
duburcqa: You really have to do this or if is juste in case ?

Trinkle23897: I'm not sure if there is a better way. Any suggestions?

tianshou/data/buffer.py
duburcqa: Ok so it should be mentioned in the doc that user keys must not start with `__` (which means hidden in Python anyway so it makes total sense)

tianshou/data/buffer.py
duburcqa: No I'm fine with it. See m'y other comment. Batch has both internal ans user defined keys, it is not surprising to restrict the names that can be defined by the user.

Trinkle23897: > Ok so it should be mentioned in the doc that user keys must not start with `__` (which means hidden in Python anyway so it makes total sense)

But this is buffer, and users cannot define key by themselves (we reserve 7 keys "obs", ...)

youkaichao:
> Are you sure it is not working out-of-the-box ?

Do you guys have any tutorial on what exactly do ``pickle.dump`` and ``pickle.load`` do? I mean, in what order, call which function. I googled but found nothing satisfying. @Trinkle23897 @duburcqa

tianshou/data/buffer.py
duburcqa: Hum Indeed m'y bad. So maybe there is a better way. But it isn't too bad.

duburcqa:
Hum strange, I found the online Python documentation about Pickle quite well written.

tianshou/data/buffer.py
Trinkle23897: I changed to `key.startswith('__') and key.endswith('__')`, seems better.

Trinkle23897: I tried with 
```python
try:
    return self._meta[key]
except KeyError:
    raise AttributeError
```
but it doesn't work.

youkaichao:
> Hum strange, I found the online Python documentation about Pickle quite well written.

Oh yes, I didn't notice [this section](https://docs.python.org/3/library/pickle.html#pickling-class-instances). After reading it, everything is crystal clear.

Trinkle23897:
Okay, so far there are 2 solutions to resolve this issue. The main issue is: `pickle.loads` will not initialize the `self._meta` first. When calling `setstate = getattr(inst, "__setstate__", None)` in `pickle.loads`, it will recursively call `buffer.__getattr__` since `self._meta` always fails to fetch.

The first is to write `__getstate__` and `__setstate__` because it will not call `__getattr__` which will cause RecursionError;

The second is to add a `if` before `self._meta[key]`:
```python
def __getattr__(self, key):
    if '_meta' not in self.__dict__:
        raise AttributeError
    try:
        return self._meta[key]
    except KeyError:
        return AttributeError
```
 But this will cause another overhead, though it doesn't need to rewrite the set/getattr. What's your opinion @duburcqa @youkaichao

youkaichao:
> The first is to write __getstate__ and __setstate__

I vote for this, the pythonic way to do it. Pickling buffer does not work out of box because the customized behavior of ``buffer.__getattr__``.

youkaichao:
I think this is actually a stupid behavior of python. Why the hell does it invoke ``__getattr__`` during pickle and unpickle? Just provide a default ``__setstate__`` and ``__getstate__`` makes the day much easier 👎

duburcqa:
> I vote for this, the pythonic way to do it. 

Why would it be more pythonic ?

I also think it is more logical to override the pickle interface, or at least part of it, since it more directly solve the problème your are facing.

youkaichao:
By "pythonic", I mean, do what python wants us to do, since python leaves these interfaces to achieve pickle.

duburcqa:
> By "pythonic", I mean, do what python wants us to do

I don't agree. Those interfaces have been designed to allow implementing much more complex behavior than one liners with default behavior except for a single attribute. So it is not clear to me if it is more appropriate to adapt the code to make the default work, or define custom interface, making pickle behavior less predictable for the user without reading the code.

Modifying the behavior of standard methods is almost always at last resort for me.

youkaichao:
Okay, I withdraw the word "pythonic".  Just because pickling buffer does not work out of box(``buffer.__getattr__`` is customized) so we need to use implement ``__setstate__`` and ``__getstate__``.

youkaichao:
Btw, take a look at #180 when you are available 😄 @duburcqa

tianshou/data/buffer.py
```diff
@@ -137,9 +143,18 @@ def __repr__(self) -> str:
...
         """Return self.key"""
-        return self._meta.__dict__[key]
+        try:
+            return self._meta[key]
+        except KeyError:
+            raise AttributeError
+
+    def __setstate__(self, state):
+        """Unpickling interface. We need it because pickling buffer does not
+        work out-of-the-box (buffer.__getattr__ is customized).
```
youkaichao:
```suggestion
        work out-of-the-box (``buffer.__getattr__`` is customized).
```

youkaichao (dismissed):

youkaichao (dismissed):

tianshou/policy/base.py
```diff
@@ -231,7 +231,8 @@ def post_process_fn(self, batch: Batch,
         usage is to update the sampling weight in prioritized experience
         replay. Check out :ref:`policy_concept` for more information.
         """
-        if isinstance(buffer, PrioritizedReplayBuffer):
+        if isinstance(buffer, PrioritizedReplayBuffer) \
+                and hasattr(batch, 'weight'):
```
youkaichao, line 235: That's strange. Does any PrioritizedReplayBuffer have no ``weight ``?

duburcqa (dismissed):

tianshou/policy/base.py
duburcqa: Good point. I would add that as a requirement.

youkaichao:
Nice to see you again! @duburcqa  We are discussing about releasing version 0.3.0 after merging #182 and #179 :) It seems like a celebration for Tianshou having 2k stars (soon)!

youkaichao (approved):

duburcqa:
Maybe it would be safe to wait one week or so before releasing it, just in case there is still a few bugs around. But I think the features are there.

PR #187, opened 2020-08-19
DQN Atari examples

yingchengyang:
This PR aims to provide the script of Atari DQN setting:
1. A speedrun of PongNoFrameskip-v4 (finished, about half an hour in i7-8750 + GTX1060 with 1M environment steps)
2. A general script for all atari game

Since we use multiple env for simulation, the result is slightly different from the original paper, but consider to be acceptable.

PR #189, opened 2020-08-22
optimize training procedure and improve code coverage

### LINKED ISSUE #96: Batch & Buffer profiling

### LINKED ISSUE #184: Memory error when sampling from collector
    + [x] exception-raising bug
    + [ ] RL algorithm bug
    + [ ] new feature request
  ```python
  import tianshou, torch, sys
  0.2.5 1.6.0 3.7.7
  [GCC 7.3.0] linux
  ```
The collector raised the memory error when I training under the self-defined environment with PPO algorithm. The full error message is as below:  
`MemoryError: Unable to allocate 8.05 GiB for an array with shape (20000, 54006) and data type float64`  
The self-defined environment uses List state observation and List discrete action. Is there any necessary to change the observation type to dict to fit tianshou and use a smaller buffer size?   
I tried to change the sample size of [on-policy](https://github.com/thu-ml/tianshou/blob/master/tianshou/trainer/onpolicy.py#L108) but it will raise the dimension matching problem(of course). Any answer is welcome for the stupid problem.

### LINKED ISSUE #185: Variable batch-size during on-policy training
    + [ ] exception-raising bug
    + [ ] RL algorithm bug
    + [x] new feature request
  ```python
  import tianshou, torch, sys
  print(tianshou.__version__, torch.__version__, sys.version)
  0.2.5, 1.5.0, 3.6.10
  ```

On-policy training now splits the replaybuffer into chunks. For 10 transitions with batch-size 4, tianshou yields three batches with batch-size 4, 4, 2, causing the variable batch-size problem.

Generally it is ok. But sometimes it causes problems, especially if we are using BatchNorm during training, which raises an error if the batch-size is 1.

I think there are three possible solutions:

- Discard the one remaining batch. For the example above, just yield two batches with batch-size 4, 4

- Pad the last batch. For the example above, yield three batches with batch-size 4, 4, 4

- Sample batches with the same batch-size. For the example above, sample three batches with batch-size 4 from 10 transitions.

Trinkle23897:
This is a cherry-pick of #187 
1. add policy.eval() in all test scripts' "watch performance"
2. remove dict return support for collector preprocess_fn
3. add `__contains__` and `pop` in batch: `key in batch`, `batch.pop(key, deft)`
4. exact n_episode for a list of n_episode limitation and save fake data in cache_buffer when self.buffer is None (#184)
5. fix tensorboard logging: h-axis stands for env step instead of gradient step; add test results into tensorboard
5. add test_returns (both GAE and nstep)
6. change the type-checking order in batch.py and converter.py in order to meet the most often case first
7. fix shape inconsistency for torch.Tensor in replay buffer
8. remove `**kwargs` in ReplayBuffer
8. remove default value in batch.split() and add merge_last argument (#185)
8. improve nstep efficiency
9. add max_batchsize in onpolicy algorithms
10. potential bugfix for subproc.wait
11. fix RecurrentActorProb
12. improve the code-coverage (from 90% to 95%) and remove the dead code
13. fix some incorrect type annotation

The above improvement also increases the training FPS: on my computer, the previous version is only ~1800 FPS and after that, it can reach ~2050 (faster than v0.2.4.post1).


youkaichao (commented):
It is surprising that changing the order of some conditions brings a speedup of over 10%. Good job!

tianshou/data/batch.py
```diff
@@ -712,18 +712,27 @@ def shape(self) -> List[int]:
...
+        assert 1 <= size <= length
         if shuffle:
             indices = np.random.permutation(length)
         else:
             indices = np.arange(length)
-        for idx in np.arange(0, length, size):
-            yield self[indices[idx:(idx + size)]]
+        count = length // size + (length % size > 0 and not merge_last)
+        if merge_last and length % size == 0:
+            merge_last = False
```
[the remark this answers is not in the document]
Trinkle23897: it cannot work

Trinkle23897:
Should be okay now and please have a check when you have time. @youkaichao @duburcqa

duburcqa:
Ok just give me a couple of jours, but based on what I already read it is quite nice ! Good job :)

tianshou/data/batch.py
```diff
@@ -728,10 +729,9 @@ def split(self, size: int, shuffle: bool = True,
             indices = np.random.permutation(length)
         else:
             indices = np.arange(length)
-        count = length // size + \
-            (length % size > 0 and not merge_last or length < size)
-        if merge_last and (length % size == 0 or count == 0):
-            merge_last = False
+        merge_last = merge_last and length % size > 0 and length >= size
+        count = length // size + (1 - merge_last) * \
+            (length % size > 0 or length < size)
```
youkaichao: That's strange. I think ``length < size`` implies ``length % size > 0``.

tianshou/policy/modelfree/ppo.py
```diff
@@ -34,6 +34,8 @@ class PPOPolicy(PGPolicy):
         defaults to ``True``.
     :param bool reward_normalization: normalize the returns to Normal(0, 1),
         defaults to ``True``.
+    :param int max_batchsize: the maximum number of batchsize when computing
+        GAE, defaults to 2000.
```
youkaichao: 2000 -> 256

tianshou/data/batch.py
Trinkle23897: but removing length%size=0 cannot pass the unit test

Trinkle23897:
ping @duburcqa

duburcqa (dismissed):

tianshou/data/collector.py
```diff
@@ -202,27 +200,42 @@ def collect(self,
             * ``rew`` the mean reward over collected episodes.
             * ``len`` the mean length over collected episodes.
         """
-        assert (n_step and not n_episode) or (not n_step and n_episode), \
-            "One and only one collection number specification is permitted!"
+        assert (n_step is not None and n_episode is None and n_step > 0) or (
+            n_step is None and n_episode is not None and np.sum(n_episode) > 0
+        ), "One and only one collection number specification is permitted!"
```
duburcqa: This message is not clear to me.

tianshou/data/collector.py
```diff
@@ -245,26 +258,27 @@ def collect(self,
             # convert None to Batch(), since None is reserved for 0-init
             if state is None:
                 state = Batch()
-            self.data.update(state=state, policy=result.get('policy', Batch()))
+            # since result is a Batch, it can bypass the type check here
+            self.data.__dict__['state'] = state
```
duburcqa: I'm not a huge fan of bypassing stuff. It makes things harder to maintain. Indeed, it is quite usual to change the internal behaviour without changing its external API. So the performance improvement is not significant, I would rather stick to the official API. But in this case it is not a serious issue.

Trinkle23897: Hmm, I'll compare those two versions

tianshou/data/collector.py
```diff
@@ -321,6 +349,10 @@ def collect(self,
                         (episode_count >= n_episode).all():
                     break
 
+        # finished envs are ready, and can be used for the next collection
+        self._ready_env_ids = np.asarray(
```
duburcqa: `np.array` should be faster, and more appropriate IMHO.

tianshou/data/collector.py
```diff
@@ -279,37 +293,51 @@ def collect(self,
...
-            if self.is_async:
+            self.data.__dict__['obs'] = obs_next
+            if is_async:
                 # set data back
+                whole_data = deepcopy(whole_data)  # avoid reference in ListBuf
                 _batch_set_item(whole_data, self._ready_env_ids,
                                 self.data, self.env_num)
                 # let self.data be the data in all environments again
                 self.data = whole_data
+            self._ready_env_ids = np.asarray(
```
duburcqa: `np.array` should be faster, and more appropriate IMHO.

tianshou/data/utils/converter.py
```diff
@@ -10,13 +10,17 @@ def to_numpy(x: Union[
     Batch, dict, list, tuple, np.ndarray, torch.Tensor]) -> Union[
         Batch, dict, list, tuple, np.ndarray, torch.Tensor]:
     """Return an object without torch.Tensor."""
-    if isinstance(x, torch.Tensor):
+    if isinstance(x, torch.Tensor):  # most often case
         x = x.detach().cpu().numpy()
+    elif isinstance(x, np.ndarray):  # second often case
+        pass
+    elif isinstance(x, (np.number, np.bool_, Number)) or x is None:
```
duburcqa: It is a bit ugly to pass None to `np.asanyarray` knowing it is None in advance.

tianshou/env/venvs.py
```diff
@@ -218,10 +222,10 @@ def seed(self,
             which a reproducer pass to "seed".
         """
         self._assert_is_not_closed()
-        if np.isscalar(seed):
-            seed = [seed + _ for _ in range(self.env_num)]
-        elif seed is None:
+        if seed is None:
             seed = [seed] * self.env_num
+        elif not isinstance(seed, list):
+            seed = [seed + _ for _ in range(self.env_num)]
```
duburcqa: `_` is not supposed to be used as an actual variable by convention. Please use something else.

tianshou/data/utils/converter.py
Trinkle23897: But if x is None, the previous version of code also returns `np.asanyarray(None)`. I'm not sure whether returning `None` directly is okay.

tianshou/policy/base.py
```diff
@@ -213,16 +213,14 @@ def compute_nstep_return(
             returns[done[now] > 0] = 0
             returns = (rew[now] - mean) / std + gamma * returns
         terminal = (indice + n_step - 1) % buf_len
-        target_q = target_q_fn(buffer, terminal).flatten()  # shape: [bsz, ]
+        target_q_ = target_q_fn(buffer, terminal).flatten()  # shape: (bsz, )
```
duburcqa: I would prefer a more explicit variable name but it is not a big deal.

tianshou/policy/modelfree/ppo.py
```diff
@@ -34,6 +34,8 @@ class PPOPolicy(PGPolicy):
         defaults to ``True``.
     :param bool reward_normalization: normalize the returns to Normal(0, 1),
         defaults to ``True``.
+    :param int max_batchsize: the maximum number of batchsize when computing
```
duburcqa: I guess `maximum number of batchsize` is not what you mean. It should be `the maximum size of the batches` no ?

tianshou/trainer/offpolicy.py
```diff
@@ -104,13 +104,13 @@ def offpolicy_trainer(
...
                         result['n/st'] // collect_per_step, t.total - t.n)):
-                    global_step += 1
+                    global_step += collect_per_step
                     losses = policy.update(batch_size, train_collector.buffer)
                     for k in result.keys():
                         data[k] = f'{result[k]:.2f}'
                         if writer and global_step % log_interval == 0:
-                            writer.add_scalar(
-                                k, result[k], global_step=global_step)
+                            writer.add_scalar('train/' + k, result[k],
```
duburcqa, line 112: Nice addition! Hierarchical data logging is very useful when a lot of data is monitored.

tianshou/policy/modelfree/ppo.py
Trinkle23897: Exactly, so how about "batch_maxsize" or "max_size_batch" or some others?

tianshou/data/utils/converter.py
duburcqa: I don't know, but I find it more intelligible and easier to handle than `array(None, dtype=object)`.

tianshou/policy/modelfree/ppo.py
duburcqa: `max_batchsize` is fine to me ! Just the description `the maximum number of batchsize` that is not :smile:

docs/tutorials/cheatsheet.rst
```diff
@@ -96,7 +96,7 @@ This is related to `Issue 42 <https://github.com/thu-ml/tianshou/issues/42>`_.
 
 If you want to get log stat from data stream / pre-process batch-image / modify the reward with given env info, use ``preproces_fn`` in :class:`~tianshou.data.Collector`. This is a hook which will be called before the data adding into the buffer.
 
-This function receives typically 7 keys, as listed in :class:`~tianshou.data.Batch`, and returns the modified part within a dict or a Batch. For example, you can write your hook as:
+This function receives typically up to 7 keys, as listed in :class:`~tianshou.data.Batch`, and returns the modified part within a :class:`~tianshou.data.Batch`. If it receives only ``obs``, it is applied in the env reset; otherwise, it is applied in the normal env step. For example, you can write your hook as:
```
duburcqa: This part is not clear. I would rather all the keys instead of just saying `typically up to 7 keys`. And I would remove `typically`. Than I would replace ` If it receives only ``obs``, it is applied in the env reset; otherwise, it is applied in the normal env step.` by `only ``obs`` is defined at env reset, while every key are specified for normal steps.`

setup.py
```diff
@@ -10,7 +10,9 @@
 
 # Get the version string
 with open(path.join(here, 'tianshou', '__init__.py')) as f:
```
duburcqa: It is not the recommended way to handle the version number. You should rather `import tianshou` in `setup.py` itself, and assign `version` to `tianshou.__version__`.

Trinkle23897: Nice catch!

Trinkle23897: Oops, it cannot work: https://github.com/thu-ml/tianshou/pull/189/checks?check_run_id=1031970772

Trinkle23897: Should I modify `tianshou/__init__.py` to only a single line:
```python
__version__ = '0.2.6'
```
?
But this will conflict with what we said in README:
```
import tianshou

If no error occurs, you have successfully installed.
```

tianshou/data/batch.py
```diff
@@ -115,17 +116,22 @@ def _create_value(inst: Any, size: int, stack=True) -> Union[
...
 def _parse_value(v: Any):
-    if isinstance(v, dict):
-        v = Batch(v)
-    elif isinstance(v, (Batch, torch.Tensor)):
-        pass
+    if isinstance(v, Batch):  # most often case
+        return v
+    elif _is_number(v):  # second often case
+        return np.asanyarray(v)
+    elif v is None or isinstance(v, np.ndarray) and \
```
duburcqa: I would add bracket to emphasize the logic.

duburcqa:
I'm done with the review ! Nice PR !

tianshou/env/venvs.py
```diff
@@ -120,12 +120,11 @@ def _wrap_id(
             self, id: Optional[Union[int, List[int]]] = None) -> List[int]:
         if id is None:
             id = list(range(self.env_num))
-        elif np.isscalar(id):
+        elif not isinstance(id, list):
```
youkaichao: Same.

tianshou/data/collector.py
youkaichao: Is it true when ``state`` is large (e.g. in RNN), too?

tianshou/data/collector.py
youkaichao: Good point. I often used ``np.array, np.asarray, np.asanyarray`` interchangeably. I will make it clear the next time:

- If the type is clear (e.g. we want to make an array from a list of int), use ``np.array``.

- If the input has something to do with ndarray (e.g. a list of ndarray), use ``np.asarray``.

youkaichao (dismissed):

setup.py
duburcqa: Ok I see the issue. Have a look to the good practice to doing what you are looking for on the official python documentation. There are several approches  including uding a "VERSION" file. It would be more standard than a ".txt" file even though it is basically the same thing.

Issue #169, opened 2020-07-27
Shared preprocess_net for actor and critic network


wuchlei:
- [ ] I have marked all applicable categories:
    + [ ] exception-raising bug
    + [ ] RL algorithm bug
    + [ ] documentation request (i.e. "X is missing from the documentation.")
    + [ ] new feature request
- [x] I have visited the [source website](https://github.com/thu-ml/tianshou/)
- [x] I have searched through the [issue tracker](https://github.com/thu-ml/tianshou/issues) for duplicates
- [x] I have mentioned version numbers, operating system and environment, where applicable:
  ```python
  import tianshou, torch, sys
  print(tianshou.__version__, torch.__version__, sys.version, sys.platform)
  ```
```
import tianshou, torch, sys
print(tianshou.__version__, torch.__version__, sys.version, sys.platform)
0.2.5 1.3.1 3.7.5 (default, Apr 19 2020, 20:18:17) 
[GCC 9.2.1 20191008] linux
```
Thanks for the good work!

So I've just migrated my envrionment from the repo [pytorch-a2c-ppo-acktr-gail](https://github.com/ikostrikov/pytorch-a2c-ppo-acktr-gail) to Tianshou. My code uses a2c for agent decision.
What I found out is that, even after tuning the hyper-parameters, Tianshou is still almost two-times slower than that simple implementation. The reason I can think of is that Tianshou uses separated preprocessing network for actor and critic network, which will embed the same state twice. However, the other repo only embeds state once.
Are there any other reasons you can think of? Thanks.

The fps of two repos are attached below.

[pytorch-a2c-ppo-acktr-gail](https://github.com/ikostrikov/pytorch-a2c-ppo-acktr-gail) 
![批注 2020-07-28 001139](https://user-images.githubusercontent.com/3911539/88576416-ff3b8d80-d077-11ea-856e-63f4c0e35752.png)

Tianshou
![批注 2020-07-28 020406](https://user-images.githubusercontent.com/3911539/88576419-01055100-d078-11ea-822b-bd1f394c481b.png)

Trinkle23897:
Thanks! The issue you mentioned is indeed a reason, but you can change this network to the desired style, something like:

```python
actor.forward(s) -> (logits, preprocess_state)
critic.forward(preprocess_state) -> V(s)
```

[...]

PR #193, opened 2020-08-27
Numba acceleration

Trinkle23897:
Training FPS improvement (base commit is 94bfb32):
test_pdqn: 1660 (without numba) -> 1930
discrete/test_ppo: 5100 -> 5170

since nstep has little impact on overall performance, the unit test result is:
GAE: 4.1s -> 0.057s
nstep: 0.3s -> 0.15s (little improvement)

Others:
- fix a bug in ttt set_eps
- keep only sumtree in segment tree implementation
- dirty fix for asyncVenv check_id test


tianshou/data/utils/segtree.py
```diff
@@ -1,30 +1,32 @@
...
     index = np.ones(value.shape, dtype=np.int64)
     while index[0] < bound:
         index *= 2
-        direct = sums[index] < value
-        value -= sums[index] * direct
+        lsons = sums[index]
+        direct = lsons < value
+        value -= lsons * direct
         index += direct
-    # for _, s in enumerate(value):
```
duburcqa, line 18: Depending on how critical the performance is for this method, I would suggest to keep using the numpy vectorized version with numba, even though it is slightly less efficient. It is way more readable.

duburcqa:
Could you add more information about the relevance of improving the efficiency of 'segtree' ? How much time is spent in it ?

Trinkle23897:
> Could you add more information about the relevance of improving the efficiency of 'segtree' ? How much time is spent in it ?

Sure, posted on the top.

Trinkle23897:
@duburcqa ready for review, any comment is appreciated!

tianshou/data/utils/segtree.py
```diff
@@ -41,27 +15,15 @@ class SegmentTree:
...
-            f'Unknown operation {operation}.'
-        if operation == 'sum':
-            self._op, self._init_value = np.add, 0.
-        elif operation == 'min':
-            self._op, self._init_value = np.minimum, np.inf
-        else:
-            self._op, self._init_value = np.maximum, -np.inf
-        # assert isinstance(self._op, np.ufunc)
-        self._value = np.full([bound * 2], self._init_value)
+        self._value = np.zeros([bound * 2])
```
duburcqa, line 26: It is great to have simplified this. Since your are reimplementing something existing, it is way better to specialize it as much as possible to fit your need while being as fast as possible.

tianshou/data/utils/segtree.py
```diff
@@ -75,59 +37,95 @@ def __setitem__(self, index: Union[int, np.ndarray],
...
+    for-loop: 0.2914 -> 0.019 (but not so stable)
+    """
+    index = np.ones(value.shape, dtype=np.int64)
+    while index[0] < bound:
+        index *= 2
+        lsons = sums[index]
+        direct = lsons < value
+        value -= lsons * direct
+        index += direct
+    # for id, s in enumerate(value):
```
duburcqa: I would remove the commented code, but it is up to you.

tianshou/policy/base.py
```diff
@@ -50,23 +52,19 @@ class BasePolicy(ABC, nn.Module):
         policy.load_state_dict(torch.load('policy.pth'))
     """
 
-    def __init__(self, **kwargs) -> None:
+    def __init__(self,
+                 observation_space: gym.Space = None,
+                 action_space: gym.Space = None
+                 ) -> None:
         super().__init__()
-        self.observation_space = kwargs.get('observation_space')
```
duburcqa, line 55: I find it more clear without `**kwargs`.

tianshou/policy/base.py
```diff
@@ -201,51 +226,72 @@ def compute_nstep_return(
...
+        now = (indice + n) % buf_len
+        gammas[done[now] > 0] = n
+        returns[done[now] > 0] = 0.
+        returns = (rew[now] - mean) / std + gamma * returns
+    target_q[gammas != n_step] = 0
+    target_q = target_q * (gamma ** gammas) + returns
+    return target_q
+
+
+def _compile():
```
duburcqa: Why do you want to precompile the function ? I don't see the point. It goes against the JIT (Just-in-Time compilation) principle. I would use instead AOT (Ahead-of-Time compilation) if you really think it is better to pre-compile whose methods.

Here is the [official documentation](https://numba.pydata.org/numba-doc/dev/user/pycc.html), including examples. Note that it is possible to define the shape of the input arrays without passing examples for both JIT and AOT.

duburcqa (commented):
Done with review!

tianshou/policy/base.py
Trinkle23897: I post the reason in the docstring. Since I want to compare the training time of several versions, if we do not compile the main function pattern ahead of time, the first-time compilation time will add onto the training time in the collector, and thus it will show that Numba slows down the whole process (from 2050 fps to 1800 fps).

Trinkle23897: I don't think AOT is a good choice since we cannot determine some arguments' type ahead of time, for example, `v_s_` in GAE.

duburcqa: Then I would rather move _compile at the beginning of the brenchmark, rather than here. I thing it makes more sense, because it is really just a benchmarking issue.

duburcqa: > I don't think AOT is a good choice since we cannot determine some arguments' type ahead of time, for example, v_s_ in GAE.

I don't now, it is strange. Compilation support variable size numpy array, just the number of dimensions must be known in advance. Knowing that you can also precompile several signatures if necessary.

Trinkle23897 (commented):
I think `_compile` should under `tianshou.utils` and include also SegTree. I'll change it.

duburcqa:
Nice ! Good to go now :)

test/base/test_env.py
```diff
@@ -90,7 +90,9 @@ def test_async_check_id(size=100, num=4, sleep=.2, timeout=.7):
     test_cls = [SubprocVectorEnv, ShmemVectorEnv]
     if has_ray():
         test_cls += [RayVectorEnv]
+    cnt = 0
     for cls in test_cls:
+        flag = 1
```
youkaichao: Give it a meaningful name: what flag it is.

test/base/test_returns.py
```diff
@@ -58,15 +58,16 @@ def test_episodic_returns(size=2560):
...
-        t = time.time()
-        for _ in range(cnt):
-            compute_episodic_return_base(batch, gamma=.1)
-        print(f'vanilla: {(time.time() - t) / cnt}')
-        t = time.time()
-        for _ in range(cnt):
-            fn(batch, None, gamma=.1, gae_lambda=1)
-        print(f'policy: {(time.time() - t) / cnt}')
+        print('GAE vanilla', timeit(vanilla, setup=vanilla, number=cnt))
+        print('GAE optim  ', timeit(optimized, setup=optimized, number=cnt))
```
youkaichao, line 70: assert some speedup here?

test/continuous/test_td3.py
```diff
@@ -120,4 +120,6 @@ def stop_fn(x):
 
 
 if __name__ == '__main__':
+    from tianshou.utils import pre_compile
+    pre_compile()  # exclude compilation time to get the correct train_speed
```
youkaichao: That's a bit strange for newcomers. Maybe ``pre_compile `` can be called during tianshou.__init__ .

Trinkle23897: I think so, but this conflicts with @duburcqa https://github.com/thu-ml/tianshou/pull/193#discussion_r480140514

test/base/test_returns.py
Trinkle23897: Numba optimized version is not the same thing as AsyncVenv, so I do not prefer to assert time here.

test/continuous/test_td3.py
youkaichao: > Then I would rather move _compile at the beginning of the brenchmark, rather than here. I thing it makes more sense, because it is really just a benchmarking issue.

There are many benchmarks in tianshou, but some of the benchmarks (in ``test`` folder) are also examples for beginners. Therefore, I think it is not a good solution to add pre-compile for every benchmark. Pre-compile in ``tianshou.__init__.py`` should be enough.

@duburcqa

duburcqa: I get your point but it is an anti-pattern. JIT is for jit. If precompile is actually able to precompile all you need, then I would use, it works just the same as jit and it is actually intenses to do what you are looking for.

duburcqa: Well AOT is not exactly as portable as JIT in practice because of some limitations. You can put it in init if you want but I really don't like the idea. I would rather keep it as it is right now. Maybe a more comprehensive name for beginner would be enough, like 'init_profiler'. It is not exactly what it does but it is what it is designed for. From this name plus a comment it would be clear it is there on the CI.

Trinkle23897: But beginners may build up their own code and benchmark and say, "Hey, your provided result is not compatible with my own test", something like #169 and #196. That's a sad thing.
Also, maybe more functions will be added to the future version. Perhaps they need to be compiled too.

duburcqa: I don't think it is a sad thing. People have never been able to properly benchmark a code, you can see this in all but one in a million posts related to stuff this on stackoverflow. It is a petty to lower our standard to match people programming knowledge, especially from people not contributing but only complaining. But I get your point too and I understand your point of view. At the end it can be changed in the future pretty easily, it is not an code design issue so it is not a big deal. As I see you seem to both agree on this point, maybe you should go for it.

Trinkle23897:
I have no more commits on this pr.

duburcqa (approved):

youkaichao (approved):

PR #200, opened 2020-09-01
type check in unit test

### LINKED ISSUE #195: Type check in the unit test
Currently there are no type-checking in the unit test. I personally recommend to use [mypy](https://github.com/python/mypy/) instead of [pytype](https://github.com/google/pytype) since pytype is too slow (? need to be checked).

However, there are still 200+ errors with mypy linting.

A possible configure for mypy.ini:

```
# Global options:

[mypy]
warn_unused_configs = True
warn_redundant_casts = True
show_error_codes = True
check_untyped_defs = True
show_error_context = True
ignore_missing_imports = True
files = tianshou/**/*.py

# Per-module options:
# To classify a given module as Level 1, 2 or 3 it must be added both in files (variable above) and in the lists below.
# Level 1 - modules checked on the strictest settings.
;[mypy-]
;warn_return_any = True
;disallow_any_expr = True
;disallow_any_explicit = True
;disallow_any_generics = True
;disallow_subclassing_any = True
;disallow_untyped_calls = True
;disallow_untyped_defs = True
;disallow_incomplete_defs = True
;check_untyped_defs = True
;disallow_untyped_decorators = True
;warn_unused_ignores = True
;follow_imports = normal
;warn_unreachable = True
;strict_equality = True

# Level 2 - module that pass reasonably strict settings.
#           No untyped functions allowed. Imports must be typed or explicitly ignored.
;[mypy-]
;warn_return_any = True
;disallow_untyped_defs = True
;disallow_incomplete_defs = True
;follow_imports = normal

# Level 3 - modules that pass mypy default settings (only those in `files` global setting and not in previous levels)
#           Function/variables are annotated to avoid mypy erros, but annotations are not complete.

# Level 4 - modules that do not pass mypy check: they are excluded from "files" setting in global section

# External packages that lack annotations
[mypy-tianshou.data.batch]
ignore_errors = True
```

Trinkle23897:
Fix #195: Add mypy test in `.github/workflows/docs_and_lint.yml`.

Also remove the out-of-the-date api

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -31,6 +31,7 @@ Here is Tianshou's other features:
 * Support :ref:`customize_training`
 * Support n-step returns estimation :meth:`~tianshou.policy.BasePolicy.compute_nstep_return` and prioritized experience replay :class:`~tianshou.data.PrioritizedReplayBuffer` for all Q-learning based algorithms; GAE, nstep and PER are very fast thanks to numba jit function and vectorized numpy operation
 * Support :doc:`/tutorials/tictactoe`
+* Comprehensive `unit tests <https://github.com/thu-ml/tianshou/actions>`_, including functional checking, RL pipeline checking, documentation checking, PEP8 code-style checking, and type checking
 
 中文文档位于 `https://tianshou.readthedocs.io/zh/latest/ <https://tianshou.readthedocs.io/zh/latest/>`_
 
--- a/tianshou/__init__.py
+++ b/tianshou/__init__.py
@@ -1,7 +1,7 @@
 from tianshou import data, env, utils, policy, trainer, exploration
 
 
-__version__ = "0.3.0a0"
+__version__ = "0.3.0rc0"
 
 __all__ = [
     "env",
--- a/tianshou/__init__.py
+++ b/tianshou/__init__.py
@@ -1,7 +1,7 @@
 from tianshou import data, env, utils, policy, trainer, exploration
 
 
-__version__ = "0.2.7"
+__version__ = "0.3.0a0"
 
 __all__ = [
     "env",
--- a/tianshou/data/batch.py
+++ b/tianshou/data/batch.py
@@ -6,7 +6,7 @@ from copy import deepcopy
 from numbers import Number
 from collections.abc import Collection
 from typing import Any, List, Dict, Union, Iterator, Optional, Iterable, \
-    Sequence, KeysView, ValuesView, ItemsView
+    Sequence
 
 # Disable pickle warning related to torch, since it has been removed
 # on torch master branch. See Pull Request #39003 for details:
@@ -144,7 +144,7 @@ def _parse_value(v: Any) -> Optional[Union["Batch", np.ndarray, torch.Tensor]]:
         if not isinstance(v, np.ndarray) and isinstance(v, Collection) and \
                 len(v) > 0 and all(isinstance(e, torch.Tensor) for e in v):
             try:
-                return torch.stack(v)
+                return torch.stack(v)  # type: ignore
             except RuntimeError as e:
                 raise TypeError("Batch does not support non-stackable iterable"
                                 " of torch.Tensor as unique value yet.") from e
@@ -191,11 +191,19 @@ class Batch:
             elif _is_batch_set(batch_dict):
                 self.stack_(batch_dict)
         if len(kwargs) > 0:
-            self.__init__(kwargs, copy=copy)
+            self.__init__(kwargs, copy=copy)  # type: ignore
 
     def __setattr__(self, key: str, value: Any) -> None:
         """Set self.key = value."""
         self.__dict__[key] = _parse_value(value)
+
+    def __getattr__(self, key: str) -> Any:
+        """Return self.key. The "Any" return type is needed for mypy."""
+        return getattr(self.__dict__, key)
+
+    def __contains__(self, key: str) -> bool:
+        """Return key in self."""
+        return key in self.__dict__
 
     def __getstate__(self) -> Dict[str, Any]:
         """Pickling interface.
@@ -215,11 +223,11 @@ class Batch:
         At this point, self is an empty Batch instance that has not been
         initialized, so it can safely be initialized by the pickle state.
         """
-        self.__init__(**state)
+        self.__init__(**state)  # type: ignore
 
     def __getitem__(
         self, index: Union[str, slice, int, np.integer, np.ndarray, List[int]]
-    ) -> Union["Batch", np.ndarray, torch.Tensor]:
+    ) -> Any:
         """Return self[index]."""
         if isinstance(index, str):
             return self.__dict__[index]
@@ -245,7 +253,7 @@ class Batch:
         if isinstance(index, str):
             self.__dict__[index] = value
             return
-        if isinstance(value, (np.ndarray, torch.Tensor)):
+        if not isinstance(value, Batch):
             raise ValueError("Batch does not supported tensor assignment. "
                              "Use a compatible Batch or dict instead.")
         if not set(value.keys()).issubset(self.__dict__.keys()):
@@ -330,30 +338,6 @@ class Batch:
             s = self.__class__.__name__ + "()"
         return s
 
-    def __contains__(self, key: str) -> bool:
-        """Return key in self."""
-        return key in self.__dict__
-
-    def keys(self) -> KeysView[str]:
-        """Return self.keys()."""
-        return self.__dict__.keys()
-
-    def values(self) -> ValuesView[Any]:
-        """Return self.values()."""
-        return self.__dict__.values()
-
-    def items(self) -> ItemsView[str, Any]:
-        """Return self.items()."""
-        return self.__dict__.items()
-
-    def get(self, k: str, d: Optional[Any] = None) -> Any:
-        """Return self[k] if k in self else d. d defaults to None."""
-        return self.__dict__.get(k, d)
-
-    def pop(self, k: str, d: Optional[Any] = None) -> Any:
-        """Return & remove self[k] if k in self else d. d defaults to None."""
-        return self.__dict__.pop(k, d)
-
     def to_numpy(self) -> None:
         """Change all torch.Tensor to numpy.ndarray in-place."""
         for k, v in self.items():
@@ -375,7 +359,6 @@ class Batch:
             if isinstance(v, torch.Tensor):
                 if dtype is not None and v.dtype != dtype or \
                         v.device.type != device.type or \
-                        device.index is not None and \
                         device.index != v.device.index:
                     if dtype is not None:
                         v = v.type(dtype)
@@ -517,7 +500,7 @@ class Batch:
             return
         batches = [x if isinstance(x, Batch) else Batch(x) for x in batches]
         if not self.is_empty():
-            batches = [self] + list(batches)
+            batches = [self] + batches
         # collect non-empty keys
         keys_map = [
             set(k for k, v in batch.items()
@@ -672,8 +655,8 @@ class Batch:
         for v in self.__dict__.values():
             if isinstance(v, Batch) and v.is_empty(recurse=True):
                 continue
-            elif hasattr(v, "__len__") and (not isinstance(
-                v, (np.ndarray, torch.Tensor)) or v.ndim > 0
+            elif hasattr(v, "__len__") and (
+                isinstance(v, Batch) or v.ndim > 0
             ):
                 r.append(len(v))
             else:
--- a/tianshou/data/buffer.py
+++ b/tianshou/data/buffer.py
@@ -1,7 +1,7 @@
 import torch
 import numpy as np
 from numbers import Number
-from typing import Any, Dict, Tuple, Union, Optional
+from typing import Any, Dict, List, Tuple, Union, Optional
 
 from tianshou.data import Batch, SegmentTree, to_numpy
 from tianshou.data.batch import _create_value
@@ -138,7 +138,7 @@ class ReplayBuffer:
         self._indices = np.arange(size)
         self.stack_num = stack_num
         self._avail = sample_avail and stack_num > 1
-        self._avail_index = []
+        self._avail_index: List[int] = []
         self._save_s_ = not ignore_obs_next
         self._last_obs = save_only_last_obs
         self._index = 0
@@ -175,12 +175,12 @@ class ReplayBuffer:
         except KeyError:
             self._meta.__dict__[name] = _create_value(inst, self._maxsize)
             value = self._meta.__dict__[name]
-        if isinstance(inst, (torch.Tensor, np.ndarray)) \
-                and inst.shape != value.shape[1:]:
-            raise ValueError(
-                "Cannot add data to a buffer with different shape, with key "
-                f"{name}, expect {value.shape[1:]}, given {inst.shape}."
-            )
+        if isinstance(inst, (torch.Tensor, np.ndarray)):
+            if inst.shape != value.shape[1:]:
+                raise ValueError(
+                    "Cannot add data to a buffer with different shape with key"
+                    f" {name}, expect {value.shape[1:]}, given {inst.shape}."
+                )
         try:
             value[self._index] = inst
         except KeyError:
@@ -205,7 +205,7 @@ class ReplayBuffer:
         stack_num_orig = buffer.stack_num
         buffer.stack_num = 1
         while True:
-            self.add(**buffer[i])
+            self.add(**buffer[i])  # type: ignore
             i = (i + 1) % len(buffer)
             if i == begin:
                 break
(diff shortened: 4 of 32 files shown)


tianshou/data/batch.py
```diff
@@ -177,12 +176,21 @@ def __init__(self,
...
                 self.stack_(batch_dict)
         if len(kwargs) > 0:
-            self.__init__(kwargs, copy=copy)
+            self.__init__(kwargs, copy=copy)  # type: ignore
 
     def __setattr__(self, key: str, value: Any) -> None:
         """self.key = value"""
         self.__dict__[key] = _parse_value(value)
 
+    def __getattr__(self, key: str) -> Any:
```
Trinkle23897, line 200: Override `__getattr__` can directly inherit existing functions of dict, thus we do not need to rewrite keys/values/items/...

tianshou/policy/multiagent/mapolicy.py
```diff
@@ -36,7 +36,8 @@ def process_fn(self, batch: Batch, buffer: ReplayBuffer,
         # reward can be empty Batch (after initial reset) or nparray.
         has_rew = isinstance(buffer.rew, np.ndarray)
         if has_rew:  # save the original reward in save_rew
-            save_rew, buffer.rew = buffer.rew, Batch()
+            # since buffer.__setattr__ is not override, here we use _meta
+            save_rew, buffer._meta.rew = buffer.rew, Batch()
```
Trinkle23897: Here is a bug fix. `buffer.rew = Batch()` will not change the reward saved previously.

Trinkle23897:
UPDATE: resolved

There are many errors like 

```python
tianshou/policy/imitation/base.py:45: error: Return type "Dict[str, float]" of "learn" incompatible with return type "Dict[str, Union[float, List[float]]]" in supertype "BasePolicy"  [override]
```

Possibly a bug in mypy:

```python
from typing import Union

class A:
    def test(self) -> Union[float, int]:
        pass

class B(A):
    def test(self) -> int:
        pass
```

This works in mypy check. However,

```python
from typing import Dict, Union

class A:
    def test(self) -> Dict[str, Union[float, int]]:
        pass

class B(A):
    def test(self) -> Dict[str, int]:
        pass
```

This cannot pass the check: 
```bash
$ mypy try.py
try.py: note: In member "test" of class "B":
try.py:8: error: Return type "Dict[str, int]" of "test" incompatible with return type "Dict[str, Union[float, int]]" in supertype "A"  [override]
Found 1 error in 1 file (checked 1 source file)
```

However, [here](https://mypy.readthedocs.io/en/latest/common_issues.html?highlight=override#incompatible-overrides) says narrow type in subclass is acceptable.

https://github.com/python/mypy/issues/9418

tianshou/data/batch.py
```diff
@@ -194,12 +203,12 @@ def __getstate__(self) -> dict:
...
+    def __setstate__(self, state: Dict[str, Any]) -> None:
         """Unpickling interface. At this point, self is an empty Batch instance
         that has not been initialized, so it can safely be initialized by the
         pickle state.
         """
-        self.__init__(**state)
+        self.__init__(**state)  # type: ignore
 
     def __getitem__(self, index: Union[
             str, slice, int, np.integer, np.ndarray, List[int]]) -> 'Batch':
```
Trinkle23897: It returns not only Batch (should be `Union['Batch', np.ndarray, torch.Tensor]`, same as `_parse_value`), but changing it to real return type will cause other typing errors.

tianshou/data/batch.py
duburcqa: I'm strongly against overloading standard methods not to change anything but for type checking. It is extremely confusing to me. It should be possible to better configure the type checker, otherwise better dropping it since it is probably not good. Those methods in particular are very critical and I feel very uncomfortable to overload them for nothing.

tianshou/data/batch.py
```diff
@@ -177,12 +180,21 @@ def __init__(self,
...
     def __setattr__(self, key: str, value: Any) -> None:
         """self.key = value"""
         self.__dict__[key] = _parse_value(value)
 
+    def __getattr__(self, key: str) -> Any:
+        """Return self.key. It is the out-of-the-box usage, but the "Any"
+        return type is needed for mypy type checking."""
+        return getattr(self.__dict__, key)
+
+    def __contains__(self, key: str) -> bool:
```
duburcqa, line 204: Same

tianshou/data/batch.py
```diff
@@ -446,37 +432,38 @@ def __cat(self,
...
             # keep it.
             lens = [0 if x.is_empty(recurse=True) else len(x)
-                    for x in batches]
+                    for x in batch_list]
         except TypeError as e:
             raise ValueError(
-                f'Batch.cat_ meets an exception. Maybe because there is any '
-                f'scalar in {batches} but Batch.cat_ does not support the '
-                f'concatenation of scalar.') from e
+                f"Batch.cat_ meets an exception. Maybe because there is any "
```
duburcqa: I would use f-string only when needed

tianshou/data/batch.py
```diff
@@ -652,8 +643,8 @@ def __len__(self) -> int:
         for v in self.__dict__.values():
             if isinstance(v, Batch) and v.is_empty(recurse=True):
                 continue
-            elif hasattr(v, '__len__') and (not isinstance(
-                    v, (np.ndarray, torch.Tensor)) or v.ndim > 0):
+            elif hasattr(v, '__len__') and (
```
duburcqa: Are you sure it is equivalent / better ?

tianshou/data/buffer.py
```diff
@@ -123,17 +124,20 @@ class ReplayBuffer:
...
+        sample_avail: bool = False
+    ) -> None:
         super().__init__()
         self._maxsize = size
         self._indices = np.arange(size)
-        self._stack = None
         self.stack_num = stack_num
         self._avail = sample_avail and stack_num > 1
-        self._avail_index = []
+        self._avail_index: List[int] = []
```
duburcqa, line 141: Are we really using variable typing in Tianshou ?

tianshou/data/collector.py
```diff
@@ -109,12 +110,14 @@ def __init__(self,
...
-    def _default_rew_metric(x):
+    def _default_rew_metric(
+        x: Union[Number, np.number]
+    ) -> Union[Number, np.number]:
         # this internal function is designed for single-agent RL
         # for multi-agent RL, a reward_metric must be provided
         assert np.asanyarray(x).size == 1, \
-            'Please specify the reward_metric ' \
-            'since the reward is not a scalar.'
+            "Please specify the reward_metric " \
```
duburcqa: I would prefer never use `\` but rather `(...)`:
```
assert np.asanyarray(x).size == 1, (
    "Please specify the reward_metric "
    "since the reward is not a scalar.")
```

tianshou/exploration/random.py
```diff
@@ -52,32 +50,39 @@ class OUNoise(BaseNoise):
...
+    def __call__(
+        self, size: Sequence[int], mu: Optional[float] = None
+    ) -> np.ndarray:
         """Generate new noise. Return a ``numpy.ndarray`` which size is equal
         to ``size``.
         """
-        if self._x is None or self._x.shape != size:
-            self._x = 0
+        if (
+            self._x is None
```
duburcqa: Strange writting.

tianshou/policy/modelfree/ddpg.py
```diff
@@ -131,16 +133,15 @@ def forward(self, batch: Batch,
             Please refer to :meth:`~tianshou.policy.BasePolicy.forward` for
             more detailed explanation.
         """
-        model = getattr(self, model)
         obs = getattr(batch, input)
-        actions, h = model(obs, state=state, info=batch.info)
+        actions, h = getattr(self, model)(obs, state=state, info=batch.info)
```
duburcqa: why changing this ? I don't remember the rational behind `getattr(self, model)` vs `self.model`

tianshou/policy/modelfree/sac.py
```diff
@@ -75,20 +75,20 @@ def __init__(self,
         self.critic2_old.eval()
         self.critic2_optim = critic2_optim
 
-        self._automatic_alpha_tuning = not isinstance(alpha, float)
-        if self._automatic_alpha_tuning:
-            self._target_entropy = alpha[0]
+        self._automatic_alpha_tuning = False
```
duburcqa: usually add `is_`, `has_` prefix to boolean variable. For instance `_is_alpha_tuning_automatic`

tianshou/policy/multiagent/mapolicy.py
duburcqa: good point.

tianshou/utils/moving_average.py
```diff
@@ -26,36 +27,38 @@ class MovAvg(object):
...
         only one element, a python scalar, or a list of python scalar.
         """
         if isinstance(x, torch.Tensor):
             x = to_numpy(x.flatten())
         if isinstance(x, list) or isinstance(x, np.ndarray):
-            for _ in x:
-                if _ not in self.banned:
-                    self.cache.append(_)
-        elif x not in self.banned:
+            for i in x:
```
duburcqa: way better.

tianshou/utils/moving_average.py
```diff
@@ -26,36 +27,38 @@ class MovAvg(object):
...
+                if i not in self.banned:
+                    self.cache.append(i)
+        elif np.isscalar(x) and x not in self.banned:
```
duburcqa: `if isinstance(x, list) or isinstance(x, np.ndarray)` is not already doing the job ?

duburcqa (dismissed):

docs/index.rst
```diff
@@ -24,12 +24,13 @@ Welcome to Tianshou!
 Here is Tianshou's other features:
 
 * Elegant framework, using only ~2000 lines of code
-* Support parallel environment sampling for all algorithms: :ref:`parallel_sampling`
+* Support parallel environment simulation (synchronous or asynchronous) for all algorithms: :ref:`parallel_sampling`
```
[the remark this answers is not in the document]
Trinkle23897: It is to be consistent with README

tianshou/data/batch.py
Trinkle23897: This `__contains__` is the previous code, I only change the order.

tianshou/data/batch.py
Trinkle23897: No, it isn't only for type checking. If we use something like `batch.pop`, it will use the pop function in `dict` without rewrite it.

Also, during my interaction with mypy, it cannot configure to remove this attribute error. Only a `getattr -> Any` can make it shut up. Otherwise, we need to add `type: ignore` everywhere. See https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html

```python
# If you want dynamic attributes on your class, have it override "__setattr__"
# or "__getattr__" in a stub or in your source code.
#
# "__setattr__" allows for dynamic assignment to names
# "__getattr__" allows for dynamic access to names
class A:
    # This will allow assignment to any A.x, if x is the same type as "value"
    # (use "value: Any" to allow arbitrary types)
    def __setattr__(self, name: str, value: int) -> None: ...

    # This will allow access to any A.x, if x is compatible with the return type
    def __getattr__(self, name: str) -> int: ...

a.foo = 42  # Works
a.bar = 'Ex-parrot'  # Fails type checking
```

This is the example described in their cheatsheet.

tianshou/data/batch.py
Trinkle23897: Since the type of `v` is (Batch, np.ndarray, torch.Tensor) (same as _parse_value), if v is not (np.ndarray and torch.Tensor), we can assume that v is a Batch.
By the way, if I use the original code-snippets, mypy will raise an error: 
```
tianshou/data/batch.py: note: In member "__len__" of class "Batch":
tianshou/data/batch.py:646: error: Left operand of 'or' is always false  [unreachable]
                elif hasattr(v, '__len__') and (not isinstance(
                                                ^
```

tianshou/data/buffer.py
Trinkle23897: This is the purpose of this PR. Remove this will cause mypy type-check error:
```
tianshou/data/buffer.py: note: In member "__init__" of class "ReplayBuffer":
tianshou/data/buffer.py:140: error: Need type annotation for '_avail_index' (hint: "_avail_index: List[<type>] = ...")  [var-annotated]
            self._avail_index = []
            ^
```

tianshou/data/collector.py
Trinkle23897: Agree. This is what `black` preferred.

tianshou/exploration/random.py
Trinkle23897: https://github.com/psf/black/blob/master/docs/the_black_code_style.md#how-black-wraps-lines

tianshou/policy/modelfree/ddpg.py
Trinkle23897: Hmm, this is another mypy error. 

First, `getattr(self, model)` is not equal to self.model (self.model is `getattr(self, 'model')`; since we want to dynamically determine which model to be used.

~~Second, **mypy doesn't accept a variable to be assigned with a different type**  currently~~, https://mypy.readthedocs.io/en/stable/common_issues.html#redefinitions-with-incompatible-types. In this scenario, `model` is a string previously (as indicated in the argument list), however, `model` is re-assigned to be a `torch.nn.Module` in `model = getattr(self, model)`. This will raise an error in mypy:
```
tianshou/policy/modelfree/dqn.py: note: In member "forward" of class "DQNPolicy":
tianshou/policy/modelfree/dqn.py:135: error: "str" not callable  [operator]
            q, h = model(obs_, state=state, info=batch.info)
                   ^
Found 1 error in 1 file (checked 47 source files)
```

update: I'm wrong. `allow_redefinition = True`

tianshou/utils/moving_average.py
Trinkle23897: Yes, you are right. I was previously in a mypy error and patch this fix, but now I find that removing this can also work well!

tianshou/data/batch.py
Trinkle23897: I tested and in fact, whether to overload getattr does not hurt the performance. The bottleneck of Batch is type-checking (lots of isinstance)

duburcqa: It is not about performance but clarity. Overloading a method for nothing is just very confusing. Type checking is supposed to make things more clear, not the contrary.

duburcqa: But if you are overloading `__getattr__` on purpose, for instance to benefit from some standard methods, I don't think it is very good either. Ideally it would be better to find a way to inherit from dict directly. But for now I guess it is good enough.

tianshou/exploration/random.py
duburcqa: so, what's the point ?

tianshou/policy/modelfree/ddpg.py
duburcqa: I don't like the idea to make everything more complex and difficult to understand just because the type checker is shit. It is a good idea in principle but currently It feels like it is just encouraging but practice.

duburcqa:
My overall comment is that typing checking should make things more robust and clearer. It is clearly not the case at the current point, so I do not recommand satisfying all its recommendations at this point. Maybe it will get better in the future but it does not look viable for now. Just my opinion though.

tianshou/exploration/random.py
Trinkle23897: I use black to format only a little parts of the code, here is the formatter’s result...

Trinkle23897:
> My overall comment is that typing checking should make things more robust and clearer. It is clearly not the case at the current point, so I do not recommand satisfying all its recommendations at this point. Maybe it will get better in the future but it does not look viable for now. Just my opinion though.

Agree most part. But I think it’s a tradeoff, especially in Python such a dynamic programming language. If we use other type checker, possibly the same things will happen again. Also, mypy is continuing improving, so I don’t pin a specific version and once the advanced features are added into mypy (like variable type re-definition), we can modify our code with minimal changes.
I think the most important part of type checker is to find the potential bugs, thus we need it. But we cannot notice the potential bugs under many type errors.

tianshou/policy/modelfree/ddpg.py
Trinkle23897: I’ll change this to a better way

tianshou/__init__.py
```diff
@@ -1,7 +1,7 @@
 from tianshou import data, env, utils, policy, trainer, exploration
 
 
-__version__ = "0.2.7"
+__version__ = "0.3.0a0"
```
duburcqa: `rc` instead of `a` is more usual in my opinion.

duburcqa (dismissed):

tianshou/__init__.py
Trinkle23897:
```suggestion
__version__ = "0.3.0rc0"
```

duburcqa (approved):

PR #213, opened 2020-09-12
code format and update function signatures

Trinkle23897:
Cherry-pick from #200 

- update the function signature
- format code-style
- move _compile into separate functions
- fix a bug in to_torch and to_numpy (Batch)
- remove None in action_range

In short, the code-format only contains function-signature style and `'` -> `"`. (pick up from [black](https://github.com/psf/black))


tianshou/data/utils/converter.py
```diff
@@ -6,67 +6,71 @@
...
+        return x
     elif isinstance(x, (np.number, np.bool_, Number)):
-        x = np.asanyarray(x)
+        return np.asanyarray(x)
     elif x is None:
-        x = np.array(None, dtype=np.object)
+        return np.array(None, dtype=np.object)
     elif isinstance(x, Batch):
         x.to_numpy()
+        return x
```
Trinkle23897, line 26: I think this inplace operation is a bug.

tianshou/data/utils/converter.py
```diff
@@ -6,67 +6,71 @@
...
+        return x.to(device)
     elif isinstance(x, (np.number, np.bool_, Number)):
-        x = to_torch(np.asanyarray(x), dtype, device)
+        return to_torch(np.asanyarray(x), dtype, device)
     elif isinstance(x, dict):
-        for k, v in x.items():
-            x[k] = to_torch(v, dtype, device)
+        return {k: to_torch(v, dtype, device) for k, v in x.items()}
     elif isinstance(x, Batch):
         x.to_torch(dtype, device)
```
Trinkle23897, line 62: same

tianshou/policy/modelfree/ddpg.py
```diff
@@ -37,20 +37,21 @@ class DDPGPolicy(BasePolicy):
...
+    def __init__(
+        self,
+        actor: Optional[torch.nn.Module],
+        actor_optim: Optional[torch.optim.Optimizer],
+        critic: Optional[torch.nn.Module],
+        critic_optim: Optional[torch.optim.Optimizer],
+        tau: float = 0.005,
+        gamma: float = 0.99,
+        exploration_noise: Optional[BaseNoise] = GaussianNoise(sigma=0.1),
+        action_range: Optional[Tuple[float, float]] = None,
```
Trinkle23897: should not be None

duburcqa: Why not ? None could be replaced by [-np.inf, np.inf] automatically. I think it makes sense.

duburcqa (dismissed):
Not much to say about this  PR in practice it changes nothing, but I can see that I don't like much the new formatting style x)

tianshou/data/utils/converter.py
duburcqa: I don't know, it is hard to tell. But in some case, no copy is required and in-place is more efficient. But at least it should be documented.

PR #217, opened 2020-09-13
change PER update interface in BasePolicy

### LINKED ISSUE #215: PER weight does not support customized replay buffer
    + [x] exception-raising bug
    + [ ] RL algorithm bug
    + [ ] new feature request

I'm running TD3 with the newest version of tianshou, and when calling td3.learn, there is an exception raised due to td1 is a tensor but weight is a numpy.ndarray, the pytorch automatically turn td1 tensor into numpy.ndarray. But I'm running on gpu, so here comes the exception.
Here is the code that raised exception:

```python
def learn(self, batch: Batch, **kwargs) -> Dict[str, float]:
    weight = batch.pop('weight', 1.)
    # critic 1
    current_q1 = self.critic1(batch.obs, batch.act).flatten()
    target_q = batch.returns.flatten()
    td1 = current_q1 - target_q
    critic1_loss = (td1.pow(2) * weight).mean() ------exception
```

```
Epoch #1:   0%|          | 0/1 [00:21<?, ?it/s]
  File "D:/PycharmProjects/Stable-BaselineTrading/Tianshou/TD3.py", line 85, in <module>
    save_fn=lambda p: torch.save(p.state_dict(), save_dir))
  File "D:\PycharmProjects\Stable-BaselineTrading\Tianshou\Trainer\offpolicy.py", line 110, in offpolicy_trainer
    losses = policy.update(batch_size, train_collector.buffer)
  File "C:\Users\zkx74\Anaconda3\envs\RL\lib\site-packages\tianshou\policy\base.py", line 147, in update
    result = self.learn(batch, *args, **kwargs)
  File "C:\Users\zkx74\Anaconda3\envs\RL\lib\site-packages\tianshou\policy\modelfree\td3.py", line 123, in learn
    critic1_loss = (td1.pow(2) * weight).mean()
  File "C:\Users\zkx74\Anaconda3\envs\RL\lib\site-packages\torch\tensor.py", line 480, in __array__
    return self.numpy()
```

By the way, it's very odd that I can run this code normally on centos gpu server without this error.

Here is my env version:
```
numpy 1.19.0
torch 1.6.0
cuda 10.1
tianshou 0.2.7
```

Trinkle23897:
fix #215

### DIFF
--- a/tianshou/policy/base.py
+++ b/tianshou/policy/base.py
@@ -6,8 +6,7 @@ from numba import njit
 from abc import ABC, abstractmethod
 from typing import Any, List, Union, Mapping, Optional, Callable
 
-from tianshou.data import Batch, ReplayBuffer, PrioritizedReplayBuffer, \
-    to_torch_as, to_numpy
+from tianshou.data import Batch, ReplayBuffer, to_torch_as, to_numpy
 
 
 class BasePolicy(ABC, nn.Module):
@@ -138,9 +137,7 @@ class BasePolicy(ABC, nn.Module):
         Typical usage is to update the sampling weight in prioritized
         experience replay. Used in :meth:`update`.
         """
-        if isinstance(buffer, PrioritizedReplayBuffer) and hasattr(
-            batch, "weight"
-        ):
+        if hasattr(buffer, "update_weight") and hasattr(batch, "weight"):
             buffer.update_weight(indice, batch.weight)
 
     def update(
@@ -253,8 +250,7 @@ class BasePolicy(ABC, nn.Module):
                                  gamma, n_step, len(buffer), mean, std)
 
         batch.returns = to_torch_as(target_q, target_q_torch)
-        # prio buffer update
-        if isinstance(buffer, PrioritizedReplayBuffer):
+        if hasattr(batch, "weight"):  # prio buffer update
             batch.weight = to_torch_as(batch.weight, target_q_torch)
         return batch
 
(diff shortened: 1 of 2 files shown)


tianshou/policy/base.py
```diff
@@ -138,9 +137,7 @@ def post_process_fn(
         Typical usage is to update the sampling weight in prioritized
         experience replay. Used in :meth:`update`.
         """
-        if isinstance(buffer, PrioritizedReplayBuffer) and hasattr(
-            batch, "weight"
-        ):
+        if hasattr(buffer, "update_weight") and hasattr(batch, "weight"):
```
duburcqa, line 140: It would be better to provide a `ReplayBufferBase` with `update_weight` doing nothing by default, instead of relying on attribute checking.

Trinkle23897: Good idea!

youkaichao: I do not agree. ``ReplayBuffer`` does not fit into the concept of ``weight``. This is specific to prioritized replaybuffer.

youkaichao (dismissed):

tianshou/policy/base.py
duburcqa: If so, provide a `PrioritizedReplayBufferBase` then. It is the same Logic: never check for attribute existence, but for inheritence. (Obviously there are some counter-examples but that's not the point here)

Trinkle23897: So where should this base class be placed? Also creating PERBase alone seems strange for me.

youkaichao: I think ``PrioritizedReplayBufferBase`` is not necessary when we have just one per buffer.

youkaichao (approved):

PR #224, opened 2020-09-21
clarify updating state

### LINKED ISSUE #203: Why setting policy.train() before collect samples in offline_trainer?
    + [ ] exception-raising bug
    + [ ] RL algorithm bug
    + [ ] new feature request
  ```python
  import tianshou, torch, sys
  ```
I have asked a Batchnorm question earlier, after reading pytorch doc, I think it's more reasonable to set policy.eval() rather than policy.train() before collect samples.
```python
for epoch in range(1, 1 + max_epoch):
        # train
        policy.train()
        if train_fn:
            train_fn(epoch)
        with tqdm.tqdm(total=step_per_epoch, desc=f'Epoch #{epoch}',
                       **tqdm_config) as t:
            while t.n < t.total:
                result = train_collector.collect(n_step=collect_per_step)
                data = {}
                if test_in_train and stop_fn and stop_fn(result['rew']):
                    test_result = test_episode(
                        policy, test_collector, test_fn,
                        epoch, episode_per_test)
```
Also, I think it should be set to eval mode before test, incase someone using BatchNorm layer or dropout.

### LINKED ISSUE #211: Potential bug caused by calling policy.eval() before collecting experience
    + [ ] exception-raising bug
    + [x] RL algorithm bug
    + [ ] new feature request
  ```python
  import tianshou, torch, sys
  ```

In [this commit](https://github.com/thu-ml/tianshou/commit/8bb8ecba6ea28fdba8f6130755b0ddf493073c84) I found policy.eval() was called before collecting experience in both on/off policy trainer, but some algorithms, such as the DDPGPolicy Tianshou provided, use the attribute `training` to control exploration and exploitation. Wouldn't this change cause serious bug by preventing these algorithms from exploration? Also, I'm not sure if there were any other place that Tianshou uses `training` to distinguish training and testing.

piece of source code of DDPGPolicy:
```python
def forward(self, batch: Batch,
                state: Optional[Union[dict, Batch, np.ndarray]] = None,
                model: str = 'actor',
                input: str = 'obs',
                explorating: bool = True,
                **kwargs) -> Batch:
        """Compute action over the given batch data.
        :return: A :class:`~tianshou.data.Batch` which has 2 keys:
            * ``act`` the action.
            * ``state`` the hidden state.
        .. seealso::
            Please refer to :meth:`~tianshou.policy.BasePolicy.forward` for
            more detailed explanation.
        """
        model = getattr(self, model)
        obs = getattr(batch, input)
        actions, h = model(obs, state=state, info=batch.info)
        actions += self._action_bias
        if self.training and explorating:
            actions += to_torch_as(self._noise(actions.shape), actions)
        actions = actions.clamp(self._range[0], self._range[1])
        return Batch(act=actions, state=h)
```

rocknamx8:
    + [ ] exception-raising fix
    + [x] new feature

  ```python
  import tianshou, torch, sys
  ```
Add an indicator(i.e. `self.learning`) of learning will be convenient for distinguishing state of policy.
Meanwhile, the state of `self.training` will be undisputed in the training stage.
Related issue: #211 

Others:
- fix a bug in DDQN: target_q could not be sampled from np.random.rand
- fix a bug in DQN atari net: it should add a ReLU before the last layer
- fix a bug in collector timing

### DIFF
--- a/docs/tutorials/concepts.rst
+++ b/docs/tutorials/concepts.rst
@@ -77,27 +77,30 @@ A policy class typically has the following parts:
 
 .. _policy_state:
 
-Three states for policy
-^^^^^^^^^^^^^^^^^^^^^^^
+States for policy
+^^^^^^^^^^^^^^^^^
 
-During the training process, the policy has three state, namely collecting state, learning state, and testing state:
-Collecting state is defined as interacting with environments and collecting training data into the buffer;
-we define the learning state as performing a model update (such as ``policy.learn()``) during training process;
-and the testing state is obvious: evaluate the performance of the current policy during training process.
+During the training process, the policy has two main states: training state and testing state. The training state can be further divided into the collecting state and updating state.
 
-In order to distinguish the collecting state, learning state and testing state, you can check the policy state by ``policy.training`` and ``policy.learning``. The state setting is as follows:
+The meaning of training and testing state is obvious: the agent interacts with environment, collects training data and performs update, that's training state; the testing state is to evaluate the performance of the current policy during training process.
 
-+------------------+-----------------+-----------------+
-| State for policy | policy.training | policy.learning |
-+==================+=================+=================+
-| Collecting state | True            | False           |
-+------------------+-----------------+-----------------+
-| Learning state   | True            | True            |
-+------------------+-----------------+-----------------+
-| Testing state    | False           | False           |
-+------------------+-----------------+-----------------+
+As for the collecting state, it is defined as interacting with environments and collecting training data into the buffer;
+we define the updating state as performing a model update by :meth:`~tianshou.policy.BasePolicy.update` during training process.
 
-``policy.learning`` is helpful to distinguish the different exploration state, for example, in DQN we don't have to use epsilon-greedy in a pure network update, so ``policy.learning`` is helpful for setting epsilon in this case.
+
+In order to distinguish these states, you can check the policy state by ``policy.training`` and ``policy.updating``. The state setting is as follows:
+
++-----------------------------------+-----------------+-----------------+
+|          State for policy         | policy.training | policy.updating |
++================+==================+=================+=================+
+|                | Collecting state |       True      |      False      |
+| Training state +------------------+-----------------+-----------------+
+|                |  Updating state  |       True      |      True       |
++----------------+------------------+-----------------+-----------------+
+|           Testing state           |       False     |      False      |
++-----------------------------------+-----------------+-----------------+
+
+``policy.updating`` is helpful to distinguish the different exploration state, for example, in DQN we don't have to use epsilon-greedy in a pure network update, so ``policy.updating`` is helpful for setting epsilon in this case.
 
 
 policy.forward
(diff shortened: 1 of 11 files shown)


docs/tutorials/concepts.rst
```diff
@@ -75,6 +75,28 @@ A policy class typically has the following parts:
 * :meth:`~tianshou.policy.BasePolicy.update`: the main interface for training. This function samples data from buffer, pre-process data (such as computing n-step return), learn with the data, and finally post-process the data (such as updating prioritized replay buffer); in short, ``process_fn -> learn -> post_process_fn``.
 
 
+.. _policy_state:
+
+Three states for policy
+^^^^^^^^^^^^^^^^^^^^^^^
+
+TODO: explain what is training state, learning state and testing state.
```
Trinkle23897: Also with a sentence about why we need policy.learning

rocknamx8: In some cases, `policy.forward()` needs to execute according to different states.

rocknamx8 (commented):
the action in _target_q of dqn may still controversial.

tianshou/policy/base.py
```diff
@@ -60,6 +60,7 @@ def __init__(
         self.observation_space = observation_space
         self.action_space = action_space
         self.agent_id = 0
+        self.learning = False
```
duburcqa: I don't like this new "state". IMHO, `training` state should be enough (with the former definition, corresponding to what you are calling `learning` in this PR). Even though this state actually exists, you always go from learning to testing. the new training state is therefore irrelevant in practice. For me, this clarification is just confusing me, by adding a new artificial state that is not used in practice. Of course you can correct me if I'm wrong.

Trinkle23897: But if "learning" is not added, we need to add eps=0 in DQN update everywhere and exploration=False in DDPG/TD3/SAC everywhere, which possibly makes more bugs. Also with some layer like BatchNorm, it needs to distinguish the different state like [data collection in training (batchnorm.eval()), nn update (batchnorm.train()), data collection in testing (batchnorm.eval())], this is a real demand.

Trinkle23897: I'll change the "training state" to "collecting state" to avoid confusion.

duburcqa: Why would you collect samples without training and exploration ? For testing ?

Trinkle23897: Yes

duburcqa: So why is `training` not enough ? Is there any case where you do not want to collect data either ?

Trinkle23897: I change the `learning` to `updating` and I think this time it is clearer:
```
During the training process, the policy has two main states: training state and testing state. 
The training state can be further divided into the collecting state and updating state.
```
2020-09-22 09-43-17 的屏幕截图

This `collecting state` doesn't exist in traditional ML pipeline, so I think it is better to clarify it.

duburcqa: It's perfect! Look way more clear to me, good job !

duburcqa (approved):

PR #229, opened 2020-09-25
change API of train_fn and test_fn

Trinkle23897:
train_fn(epoch) -> train_fn(epoch, env_step)
test_fn(epoch) -> test_fn(epoch, env_step)


youkaichao:
What is the rationale of changing the signature? Maybe ``env_step `` means the number of steps for environments. Then I think ``env_step`` should be passed to ``test_fn ``, too.

BTW, I find ``env_step`` not so meaningful. I prefer something like ``num_steps``.

youkaichao:
I agree with the idea of this PR. Just remember to change all occurrence. This PR changes the signature and thus breaks the compatibility, so it is easy to check whether you have changed all occurrence: just re-run all the examples.

Trinkle23897:
> I agree with the idea of this PR. Just remember to change all occurrences. This PR changes the signature and thus breaks the compatibility, so it is easy to check whether you have changed all occurrences: just re-run all the examples.

Yes, I've checked and all examples are runnable.

docs/tutorials/dqn.rst
```diff
@@ -123,9 +123,9 @@ Tianshou provides :class:`~tianshou.trainer.onpolicy_trainer` and :class:`~tians
         policy, train_collector, test_collector,
         max_epoch=10, step_per_epoch=1000, collect_per_step=10,
         episode_per_test=100, batch_size=64,
-        train_fn=lambda e: policy.set_eps(0.1),
-        test_fn=lambda e: policy.set_eps(0.05),
-        stop_fn=lambda x: x >= env.spec.reward_threshold,
+        train_fn=lambda epoch, num_env_step: policy.set_eps(0.1),
```
duburcqa: IMHO `env_step_idx` would be clearer

examples/atari/runnable/pong_a2c.py
```diff
@@ -76,11 +76,11 @@ def test_a2c(args=get_args()):
         preprocess_fn=preprocess_fn)
     test_collector = Collector(policy, test_envs, preprocess_fn=preprocess_fn)
     # log
-    writer = SummaryWriter(args.logdir + '/' + 'a2c')
+    writer = SummaryWriter(args.logdir + '/a2c')
```
duburcqa: `os.path.join(args.logdir, 'a2c')`

examples/atari/runnable/pong_ppo.py
```diff
@@ -80,11 +80,11 @@ def test_ppo(args=get_args()):
         preprocess_fn=preprocess_fn)
     test_collector = Collector(policy, test_envs, preprocess_fn=preprocess_fn)
     # log
-    writer = SummaryWriter(args.logdir + '/' + 'ppo')
+    writer = SummaryWriter(args.logdir + '/ppo')
```
duburcqa: same.

duburcqa (dismissed):

docs/tutorials/dqn.rst
duburcqa: usually, `num` is for total/cumulative sum. It is a constant that is not supposed to change. That's why I think `idx` is more appropriate.

youkaichao: How about simply use "step", just like "epoch".

Trinkle23897: My original proposal is "env_step" because "step" may be ambiguous. Since we have two kinds of step in trainer: env step and gradient step.

youkaichao: "env_step" is okay then. Be sure to explain it in the doc.

youkaichao (approved):

PR #230, opened 2020-09-25
add singleton pattern version of summary_writter

rocknamx8:
    + [ ] exception-raising fix
    + [x] new feature

  ```python
  import tianshou, torch, sys
  ```
add a singleton pattern version of SummaryWriter may convenient for logging in different modules.

### DIFF
--- a/test/base/test_utils.py
+++ b/test/base/test_utils.py
@@ -2,11 +2,11 @@ import torch
 import numpy as np
 
 from tianshou.utils import MovAvg
-from tianshou.exploration import GaussianNoise, OUNoise
+from tianshou.utils import SummaryWriter
 from tianshou.utils.net.common import Net
 from tianshou.utils.net.discrete import DQN
+from tianshou.exploration import GaussianNoise, OUNoise
 from tianshou.utils.net.continuous import RecurrentActorProb, RecurrentCritic
-from tianshou.utils import SummaryWriter
 
 
 def test_noise():
--- a/test/base/test_utils.py
+++ b/test/base/test_utils.py
@@ -67,10 +67,8 @@ def test_summary_writer():
     # get first instance by key of `default` or your own key
     writer1 = SummaryWriter.get_instance(
         key="first", log_dir="log/test_sw/first")
-    writer2 = SummaryWriter.get_instance(
-        key="default", log_dir="log/test_sw/first")
-    writer3 = SummaryWriter.get_instance()
-    assert writer1 is writer2 is writer3
+    writer2 = SummaryWriter.get_instance()
+    assert writer1 is writer2
     # create new instance by specify a new key
     writer4 = SummaryWriter.get_instance(
         key="second", log_dir="log/test_sw/second")
--- a/test/base/test_utils.py
+++ b/test/base/test_utils.py
@@ -22,14 +22,14 @@ def test_moving_average():
     stat = MovAvg(10)
     assert np.allclose(stat.get(), 0)
     assert np.allclose(stat.mean(), 0)
-    assert np.allclose(stat.std()**2, 0)
+    assert np.allclose(stat.std() ** 2, 0)
     stat.add(torch.tensor([1]))
     stat.add(np.array([2]))
     stat.add([3, 4])
     stat.add(5.)
     assert np.allclose(stat.get(), 3)
     assert np.allclose(stat.mean(), 3)
-    assert np.allclose(stat.std()**2, 2)
+    assert np.allclose(stat.std() ** 2, 2)
 
 
 def test_net():
@@ -65,15 +65,15 @@ def test_net():
 
 def test_summary_writer():
     # get first instance by key of `default` or your own key
-    writer1 = SummaryWriter.get_instance(key="first",
-                                         log_dir="log/test_sw/first")
-    writer2 = SummaryWriter.get_instance(key="default",
-                                         log_dir="log/test_sw/first")
+    writer1 = SummaryWriter.get_instance(
+        key="first", log_dir="log/test_sw/first")
+    writer2 = SummaryWriter.get_instance(
+        key="default", log_dir="log/test_sw/first")
     writer3 = SummaryWriter.get_instance()
     assert writer1 is writer2 is writer3
     # create new instance by specify a new key
-    writer4 = SummaryWriter.get_instance(key="second",
-                                         log_dir="log/test_sw/second")
+    writer4 = SummaryWriter.get_instance(
+        key="second", log_dir="log/test_sw/second")
     writer5 = SummaryWriter.get_instance(key="second")
     assert writer1 is not writer4
     assert writer4 is writer5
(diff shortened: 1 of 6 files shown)


tianshou/utils/log_tools.py
```diff
@@ -0,0 +1,27 @@
...
+    """
+    A more convenient Summary Writer, you can get the same instance of summary
+    writer every where after you have create one.
+
+    """
+    _mutex_lock = threading.Lock()
+    _default_key = None
+
+    @classmethod
+    def get_instance(cls, key: str = 'default', *args, **kwargs):
```
youkaichao: Maybe you can specify the ``args, kwargs`` (like ``log_dir``). They make coding much easier.

test/base/test_utils.py
```diff
@@ -21,14 +22,14 @@ def test_moving_average():
     stat = MovAvg(10)
     assert np.allclose(stat.get(), 0)
     assert np.allclose(stat.mean(), 0)
-    assert np.allclose(stat.std() ** 2, 0)
+    assert np.allclose(stat.std()**2, 0)
```
Trinkle23897, line 25:
```suggestion
    assert np.allclose(stat.std() ** 2, 0)
```

test/base/test_utils.py
```diff
@@ -21,14 +22,14 @@ def test_moving_average():
...
     stat.add(torch.tensor([1]))
     stat.add(np.array([2]))
     stat.add([3, 4])
     stat.add(5.)
     assert np.allclose(stat.get(), 3)
     assert np.allclose(stat.mean(), 3)
-    assert np.allclose(stat.std() ** 2, 2)
+    assert np.allclose(stat.std()**2, 2)
```
Trinkle23897, line 32:
```suggestion
    assert np.allclose(stat.std() ** 2, 2)
```

test/base/test_utils.py
```diff
@@ -62,7 +63,24 @@ def test_net():
...
+    writer1 = SummaryWriter.get_instance(key="first",
+                                         log_dir="log/test_sw/first")
+    writer2 = SummaryWriter.get_instance(key="default",
+                                         log_dir="log/test_sw/first")
+    writer3 = SummaryWriter.get_instance()
+    assert writer1 is writer2 is writer3
+    # create new instance by specify a new key
+    writer4 = SummaryWriter.get_instance(key="second",
+                                         log_dir="log/test_sw/second")
+    writer5 = SummaryWriter.get_instance(key="second")
```
Trinkle23897, line 77:
```suggestion
    writer1 = SummaryWriter.get_instance(
        key="first", log_dir="log/test_sw/first")
    writer2 = SummaryWriter.get_instance(
        key="default", log_dir="log/test_sw/first")
    writer3 = SummaryWriter.get_instance()
    assert writer1 is writer2 is writer3
    # create new instance by specify a new key
    writer4 = SummaryWriter.get_instance(
        key="second", log_dir="log/test_sw/second")
    writer5 = SummaryWriter.get_instance(key="second")
```

tianshou/utils/__init__.py
```diff
@@ -1,7 +1,9 @@
 from tianshou.utils.config import tqdm_config
 from tianshou.utils.moving_average import MovAvg
+from tianshou.utils.log_tools import SummaryWriter
 
 __all__ = [
     "MovAvg",
     "tqdm_config",
+    "SummaryWriter"
```
Trinkle23897, line 8:
```suggestion
    "SummaryWriter",
```

tianshou/utils/log_tools.py
```diff
@@ -0,0 +1,27 @@
+from torch.utils import tensorboard
+import threading
```
Trinkle23897, line 2:
```suggestion
import threading
from torch.utils import tensorboard
```

tianshou/utils/log_tools.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class SummaryWriter(tensorboard.SummaryWriter):
+    """
+    A more convenient Summary Writer, you can get the same instance of summary
+    writer every where after you have create one.
+
```
Trinkle23897, line 9:
```suggestion
    """A more convenient Summary Writer.
    
    You can get the same instance of summary writer everywhere after you
    created one.
```

tianshou/utils/log_tools.py
```diff
@@ -0,0 +1,27 @@
...
+    _mutex_lock = threading.Lock()
+    _default_key = None
+
+    @classmethod
+    def get_instance(cls, key: str = 'default', *args, **kwargs):
+        with SummaryWriter._mutex_lock:
+            if not hasattr(SummaryWriter, "_instance"):
+                SummaryWriter._instance = {
+                    key: SummaryWriter(*args, **kwargs)
+                    }
```
Trinkle23897, line 20:
```suggestion
                }
```

Trinkle23897 (changes requested):

duburcqa:
For the record, `traitlets` is a powerful module maintained by `ipython` teams mainly used for type checking at import time, but it provide a base class for singleton pattern just as you reimplemented (you can get the global instance using `instance()` method). IT would make your implementation only 2 lines of code instead of 80 :)

Waiting for your point of view on this point before going further into reviewing this PR.

tianshou/utils/log_tools.py
```diff
@@ -0,0 +1,62 @@
...
+        >>> writer4 = SummaryWriter.get_instance(
+                key="second", log_dir="log/test_sw/second")
+        >>> writer5 = SummaryWriter.get_instance(key="second")
+        >>> writer1 is not writer4
+        True
+        >>> writer4 is writer5
+        True
+    """
+
+    _mutex_lock = threading.Lock()
```
duburcqa, line 27: The mutex would be different if used in conjunction with multiprocessing. The only way to circumvent this limitation is to add the lock in the constructor. I don't know if  `traitlet` is handling this properly but I guess so.

Trinkle23897:
> For the record, `traitlets` is a powerful module maintained by `ipython` teams mainly used for type checking at import time, but it provide a base class for singleton pattern just as you reimplemented (you can get the global instance using `instance()` method). IT would make your implementation only 2 lines of code instead of 80 :)
> 
> Waiting for your point of view on this point before going further into reviewing this PR.

I look at the traitlets code (https://github.com/ipython/traitlets/blob/294219773e453cd7d05b7906923197ceb8dd8731/traitlets/config/configurable.py#L508) and it seems that we need to inherit the original SummaryWriter instead of a function decorator.
A possible way is:
```python
from traitlets.config.configurable import SingletonConfigurable
class SummaryWriter(tensorboard.SummaryWriter, SingletonConfigurable):
    pass
```
But traitlets singleton class does not have a mutex lock and the key selection.

I simplify the code just now, but in my opinion it is more common to create a singleton decorator (or import an existing implementation), something like https://stackoverflow.com/a/7346105.

tianshou/utils/log_tools.py
Trinkle23897: > I don't know if traitlet is handling this properly but I guess so.

nope

danagi: I don't think there's any situation where singleton summarywriter need to be used in multi-process. If that happened, then different summarywriters were supposed to be used or only the main process should take responsibility for logging. The mutex here is mainly used to avoid creating multiple instances with same key in multi-thread.

danagi (approved):

Trinkle23897 (approved):

PR #235, opened 2020-10-04
change the step in trainer

Trinkle23897:
This PR separates the `global_step` into `env_step` and `gradient_step`. In the future, the data from the collecting state will be stored under `env_step`, and the data from the updating state will be stored under `gradient_step`.

Others:
- add `rew_std` and `best_result` into the monitor
- fix network unbounded in `test/continuous/test_sac_with_il.py` and `examples/box2d/bipedal_hardcore_sac.py`
- change the dependency of ray to 1.0.0 since ray-project/ray#10134 has been resolved

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -60,7 +60,7 @@ setup(
             "flake8",
             "pytest",
             "pytest-cov",
-            "ray>=0.8.0",
+            "ray>=1.0.0",
             "mypy",
             "pydocstyle",
             "doc8",
--- a/tianshou/data/collector.py
+++ b/tianshou/data/collector.py
@@ -214,7 +214,7 @@ class Collector(object):
         # number of episodes, these envs will be recorded in this list, and
         # they will not be stepped.
         finished_env_ids = []
-        reward_total = []
+        rewards = []
         whole_data = Batch()
         if isinstance(n_episode, list):
             assert len(n_episode) == self.get_env_num()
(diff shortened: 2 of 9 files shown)


setup.py
```diff
@@ -60,7 +60,7 @@ def get_version() -> str:
             "flake8",
             "pytest",
             "pytest-cov",
-            "ray>=0.8.0",
+            "ray>=1.0.0",
```
duburcqa, line 63: I think ray should be an optional requirement, because it it not necessary at all to use most of the features of Tianshou.

tianshou/data/collector.py
```diff
@@ -214,7 +214,7 @@ def collect(
         # number of episodes, these envs will be recorded in this list, and
         # they will not be stepped.
         finished_env_ids = []
-        reward_total = 0.0
+        reward_total = []
```
duburcqa: `rewards` would be a more appropriate name at this point

duburcqa (dismissed):

setup.py
Trinkle23897: Yes, this is optional now since the requirements are under "dev".

duburcqa: Indeed my bad, I missed it !

duburcqa (dismissed):

danagi (approved):

PR #244, opened 2020-11-08
Mujoco td3 benchmark

### LINKED ISSUE #209: mujoco examples' result mismatch
With all of ddpg, sac and td3, I get similar results like this:
image

But the official results like this:
image

ChenDRAG:
This is the first step to try to add benchmarks for mujoco games using tianshou platform. I try to let parameters be consistent with [Github: TD3](https://github.com/sfujim/TD3), and results are almost the same as TD3 baseline.

However, this is still quite far from being satisfying. We still lack detailed experiment results (TD3 algorithm is only evaluated with a fixed seed 0), and due to some reasons, our agents learn a bit slower than tianshou. Benchmarks for other algorithms like sac&ddpg still wait to be supported.


ChenDRAG:
partly solve #209

ChenDRAG:
@Trinkle23897 I think this pr should be closed now. #274 will do its work.

PR #246, opened 2020-11-09
SAC mujoco result

### LINKED ISSUE #209: mujoco examples' result mismatch

### LINKED ISSUE #233: Deterministic output for SAC policy
When I ran the example code, I found that even in the testing session, the SAC strategy is still a random policy. Shouldn't the mean value be output as a deterministic policy during the training?

Trinkle23897:


tianshou/policy/modelfree/sac.py
```diff
@@ -125,6 +130,9 @@ def forward(  # type: ignore
         if self._noise is not None and not self.updating:
             act += to_torch_as(self._noise(act.shape), act)
         act = act.clamp(self._range[0], self._range[1])
+        if self._deterministic_eval and not self.training:
+            act = torch.tanh(logits[0]) * \
+                self._action_scale + self._action_bias
```
danagi: Stochastic action is sampled before these lines causing the `log_prob` to be wrong and `act.clamp` to be ignored. Suggest coding like `x = logits[0] if self._deterministic_eval and not self.training else dist.rsample()` and updating the logic of adding noise instead.

danagi (changes requested):

tianshou/policy/modelfree/sac.py
```diff
@@ -130,9 +133,6 @@ def forward(  # type: ignore
         if self._noise is not None and not self.updating:
             act += to_torch_as(self._noise(act.shape), act)
```
danagi: action noise is supposed to be added at collecting state (`self.training`=`True` and `self.updating`=`False`) only but it is added at  testing state (`self.training`=`False` and `self.updating`=`False`) also now.

danagi (changes requested):

danagi (approved):

PR #261, opened 2020-12-09
Saving and loading replay buffer with HDF5

### LINKED ISSUE #260: Saving and loading replay buffer with HDF5
    + [ ] exception-raising bug
    + [ ] RL algorithm bug
    + [x] new feature request
  ```python
  import tianshou, torch, sys
  ```
yields
```bash
0.3.0 1.7.0 3.6.9 (default, Oct  8 2020, 12:12:24)
[GCC 8.4.0] linux
```

Hi, 
I have a suggestion regarding loading and saving the replay buffer. My understanding is that the currently implemented solution uses pickle. However, pickle has a large memory overhead which can be problematic for large replay buffers. In my case pickling a replay buffer of about 2,5 GB took more than 17 GB of memory. In some cases this overhead is too large.

I therefore implemented saving to and loading from HDF5 files. The idea is to represent a Batch by a group in the HDF5 file whereas tensors and numpy arrays are represented by datasets. This would add a dependency on h5py.

There is still an overhead when loading the replay buffer, but it is smaller. In the example above the peak memory usage was 7,3 GB. This could possibly be improved though as I am not an expert when it comes to HDF5.

If you are interested, I can make a pull request.

nicoguertler:
    + [ ] exception-raising fix
    + [x] new feature

As mentioned in #260, this pull request is about an implementation of saving and loading the replay buffer with HDF5.

I added a test of this feature but its probably not exhaustive. Feedback on whether saving and loading works for real projects is welcome.


setup.py
```diff
@@ -51,6 +51,7 @@ def get_version() -> str:
         "tensorboard",
         "torch>=1.4.0",
         "numba>=0.51.0",
+        "h5py"
```
Trinkle23897: Do we need any version constraints?

test/base/test_buffer.py
```diff
@@ -278,6 +280,48 @@ def test_pickle():
...
+    _vbuf.load_contents(path)
+
+    assertions(vbuf, _vbuf)
+
+    os.remove(path)
+
+
 if __name__ == '__main__':
     test_replaybuffer()
     test_ignore_obs_next()
```
Trinkle23897, line 353: You can also add test_hdf5 here. (But it doesn't matter in the ci)

Trinkle23897 (commented):
.swp file should not add to this PR.

test/base/test_buffer.py
```diff
@@ -278,6 +280,48 @@ def test_pickle():
...
+    device = 'cuda' if torch.cuda.is_available() else 'cpu'
+    rew = torch.tensor([1.]).to(device)
+    for i in range(4):
+        vbuf.add(
+                obs=Batch(index=np.array([i])),
+                act=i,
+                rew=rew,
+                done=0,
+                info={"number": i})
+
```
Trinkle23897:
```suggestion
        vbuf.add(
            obs=Batch(index=np.array([i])),
            act=i,
            rew=rew,
            done=0,
            info={"number": i}
        )

```

tianshou/data/buffer.py
```diff
@@ -359,6 +360,78 @@ def __getitem__(
             policy=self.get(index, "policy"),
         )
 
+    @classmethod
+    def _copy_to_hdf5(
+            cls, k: str, v: Union[np.ndarray, Batch], grp: h5py.Group
+            ) -> None:
```
Trinkle23897:
```suggestion
        cls, k: str, v: Union[np.ndarray, Batch], grp: h5py.Group
    ) -> None:
```

tianshou/data/buffer.py
```diff
@@ -359,6 +360,78 @@ def __getitem__(
...
+            grp.create_dataset(k, data=v.numpy())
+        elif isinstance(v, Batch):
+            subgrp = grp.create_group(k)
+            for bk, bv in v.__dict__.items():
+                cls._copy_to_hdf5(bk, bv, subgrp)
+
+    @classmethod
+    def _copy_from_hdf5(
+            cls, grp: h5py.Group, dst: Batch, device: Optional[str] = "numpy"
+            ) -> None:
```
Trinkle23897:
```suggestion
        cls, grp: h5py.Group, dst: Batch, device: Optional[str] = "numpy"
    ) -> None:
```

tianshou/data/buffer.py
```diff
@@ -359,6 +360,78 @@ def __getitem__(
...
+            for k, v in self.__dict__.items():
+                if k not in ["_meta", "_indices"]:
+                    f.attrs[k] = v
+
+            for k, v in self._meta.__dict__.items():
+                self._copy_to_hdf5(k, v, f)
+
+    def load_contents(
+            self, path: str, device: Optional[str] = "numpy"
+            ) -> None:
```
Trinkle23897:
```suggestion
        self, path: str, device: Optional[str] = "numpy"
    ) -> None:
```

tianshou/data/buffer.py
```diff
@@ -359,6 +360,78 @@ def __getitem__(
...
+            assert f.attrs["_maxsize"] == self._maxsize, \
+                    f"Data size in '{path}' deviates from buffer size."
+            for k in ["_size", "_index"]:
+                self.__dict__[k] = f.attrs[k]
+            self._copy_from_hdf5(f, self._meta, device=device)
+
+    @classmethod
+    def load(
+            cls, path: str, device: Optional[str] = "numpy"
+            ) -> "ReplayBuffer":
```
Trinkle23897:
```suggestion
        cls, path: str, device: Optional[str] = "numpy"
    ) -> "ReplayBuffer":
```

Trinkle23897 (commented):
These are formatting issues.

tianshou/data/buffer.py
```diff
@@ -359,6 +360,78 @@ def __getitem__(
...
+                        raise Exception("Cannot copy HDF5 dataset into object"
+                                        f"with type {type(dst.__dict__[k])}.")
+                else:
+                    if device == "numpy":
+                        dst.__dict__[k] = np.empty(v.shape, dtype=v.dtype)
+                        v.read_direct(dst.__dict__[k])
+                    else:
+                        dst.__dict__[k] = torch.tensor(v, device=device)
+
+    def save(self, path: str) -> None:
```
Trinkle23897: I personally suggest changing this function name to `save_hdf5` because we may add other types of save/load functionality in the future. Same as load.

test/base/test_buffer.py
nicoguertler: I have added the version requirement h5py>=3.1.0 because this is the version I have tested the code with. Possibly runs with older versions as well though. I also added the test_hdf5 and removed the swap files as you suggested.

tianshou/data/buffer.py
nicoguertler: Ok, good point. I have added _hdf5 to the function names.

Trinkle23897:
Seems it cannot work well with cuda tensor. (github action doesn't have cuda environment :(

```bash
$ python3 test/base/test_buffer.py                                               
  File "test/base/test_buffer.py", line 327, in <module>
    test_hdf5()
  File "test/base/test_buffer.py", line 300, in test_hdf5
    vbuf.save(path)
  File "/home/trinkle/github/tianshou-new/tianshou/data/buffer.py", line 409, in save
    self._copy_to_hdf5(k, v, f)
  File "/home/trinkle/github/tianshou-new/tianshou/data/buffer.py", line 370, in _copy_to_hdf5
    grp.create_dataset(k, data=v.numpy())
```

update: resolved

Trinkle23897:
Also, could you please add this feature description in the docstring of `ReplayBuffer`?

nicoguertler:
Thank you for your feedback! I can't test in a cuda environment right now but I think calling .cpu() first should fix the issue. Also thanks for adapting the test method. I have added loading and saving with HDF5 to the usage  code snippet. Let me now if you had something else in mind.

tianshou/data/buffer.py
```diff
@@ -359,6 +372,78 @@ def __getitem__(
...
+                        dst.__dict__[k] = np.empty(v.shape, dtype=v.dtype)
+                        v.read_direct(dst.__dict__[k])
+                    else:
+                        dst.__dict__[k] = torch.tensor(v, device=device)
+
+    def save_hdf5(self, path: str) -> None:
+        """Save replay buffer to HDF5 file."""
+        with h5py.File(path, "w") as f:
+            for k, v in self.__dict__.items():
+                if k not in ["_meta", "_indices"]:
```
Trinkle23897: I'm not comfortable with L417 and L430. `_meta` is fine but others are kind of tricky. It means we have to pay additional attention to these lines in future maintenance. Is there a better way of grp organization?

```
- xxx
- xxx
- meta
  - xxx
  - xxx
```

tianshou/data/buffer.py
```diff
@@ -359,6 +372,78 @@ def __getitem__(
             policy=self.get(index, "policy"),
         )
 
+    @classmethod
+    def _copy_to_hdf5(
```
Trinkle23897: ~I'm considering if it is appropriate to move this function (also `copy_from_hdf5`) to `batch.py`. What do you think?~

nicoguertler: Yes, sounds reasonable to move copy_from_hdf5 and copy_to_hdf5 to batch.py as they are not directly related to the replay buffer.

tianshou/data/buffer.py
nicoguertler: I can see your point. Maybe an alternative way would be to introduce a method _get_state that returns a dictionary with all attributes that should be saved to a file when saving the replay buffer? At some point a distinction between attributes like _indices that do not have to be saved and those that should be saved has to be made.

Trinkle23897: You can use `__getstate__` directly. It works.

Trinkle23897: Can we build this feature **only** on top of `__getstate__` and `__setstate__`?

tianshou/data/buffer.py
duburcqa: I think you should have a look to how pickle works, and use the same hidden methods interface. Basically, the hdf5 data to store should be the same than the pickle one, thus I think you should rather adapt the pickle interface to handle both hdf5 and usual pickle format. It is not always possible, but since in this case you are playing with classes meant to deal with data, it should be doable fairly easily. At the end, it would be beneficial to pickle, which is likely to be more efficient if optimised this way, but even more importantly easier to understand and maintain.

tianshou/data/buffer.py
nicoguertler: Sounds good, however, HDF5 isn't really intended to be interchangeable with pickle. So I am not sure whether it is possible to write saving and loading code that won't break no matter what changes to the attributes are done. I will try to come up with something that is more general (like you suggested for group organization) and get back to you.

duburcqa: Yes your are right. In the general case, hdf5 is not intended to be interchangeable with pickle, but in this particular case, you are dealing with a class designed to store data. In this particular setting, I think it should be possible to come up a common method to extract raw data from `batch` and restore them. Proabably it won't be possible to avoid having to implement a custom `from/to_hdf5` interface, but it should be reduced to bare minimum.

tianshou/data/buffer.py
nicoguertler: Ok, I agree that unifying the data exposed to pickle and saved via HDF5 is a good idea. I implemented `__getstate__` and `__setstate__` in `ReplayBuffer` to work with a dictionary in which a `Batch` is represented by a dictionary (via `Batch.__getstate__`). `save_hdf5` is now based on the output of `__getstate__`. 
The main assumption when converting between this dictionary and HDF5 is that all entries which aren't dictionaries, numpy arrays or pytorch tensors can be saved as HDF5 attributes. According to the h5py documentation that means they have to be scalars (numpy arrays are already taken care of explicitly) and smaller than 64k. 
Currently, saving and loading with HDF5 doesn't work for `ListReplayBuffer` and `PrioritizedReplayBuffer`. I can have a look at that when we have converged on something for ReplayBuffer.

duburcqa: > The main assumption when converting between this dictionary and HDF5 is that all entries which aren't dictionaries

This is not necessary. Andy hierarchy is natively supported by hdf5 standard.

> According to the h5py documentation that means they have to be scalars (numpy arrays are already taken care of explicitly) 

What do you mean ? I dont get it.

> and smaller than 64k.

If such limitation actually existe, it is not part of hdf5 standard. Can you point me out where you found this information ?

tianshou/data/buffer.py
```diff
@@ -394,6 +480,17 @@ def reset(self) -> None:
...
+        return self.__dict__
+
+    def __setstate__(self, state: Dict[str, Any]) -> None:
+        """Unpickling interface.
+
+        We need it because pickling buffer does not work out-of-the-box
+        ("buffer.__getattr__" is customized).
+        """
+        self.__dict__.update(state)
+
```
Trinkle23897: I don't think it is necessary since it is equal to the implementation in `ReplayBuffer`. It will directly inherit that function. Same as the below.

tianshou/data/buffer.py
```diff
@@ -394,6 +480,17 @@ def reset(self) -> None:
             if isinstance(self._meta.__dict__[k], list):
                 self._meta.__dict__[k] = []
 
+    def __getstate__(self) -> dict:
+        return self.__dict__
```
Trinkle23897: I don't think it is necessary because this is the original `__getstate__` function implementation. Maybe we need some extra tests to prove that?

duburcqa: Yes, but no. It derives from ReplayBuffer, so it inherits its implementation. This if it it necessary to modify `set`, the same goes for `get`. Even though I agree it shouldn't be necessary to définie `set` in the first place 👍

tianshou/data/buffer.py
nicoguertler: Hi, thank you for your reply. I was talking about attributes and not datasets. The information I am referring to can be found here: https://docs.h5py.org/en/stable/high/attr.html. It is also possible to store small numpy arrays as attributes but I explicitly save them in datasets. This is what I was trying to say.
Yes, it is not strictly necessary to make these assumptions. I can also implement something more general but this will increase the amount of code needed. If you would prefer a solution that can deal with fairly general dictionaries and lists I can implement that.

tianshou/data/buffer.py
nicoguertler: I put these methods in `ListReplayBuffer` and `PrioritizedReplayBuffer` to make them behave as before with pickle because I did not have a look at them yet. I tried to mention that above, sorry if I haven't made myself clear. Yes, in the final implementation it shouldn't be necessary to define these methods in derived classes.

tianshou/data/buffer.py
duburcqa: Do not make it overly general if not strictly necessary. I juste want to make sure your implementation is robust for the intended use, and flexible enough to be eventually extended if `ReplayBuffer` gets more complex in the future or new classes need to support from/to hdf5 interface.

duburcqa: Also, note that the 64k limits is only a recommandation and not part of the standard. I'm not aware of any shortcoming of using attributs to store any kind of data, of any size.

tianshou/data/utils/converter.py
```diff
@@ -80,3 +82,58 @@ def to_torch_as(
     """
     assert isinstance(y, torch.Tensor)
     return to_torch(x, dtype=y.dtype, device=y.device)
+
+
+def to_hdf5(x: Any, y: h5py.Group) -> None:
```
duburcqa: `Any` is not convertible to hdf5 format. You should be way more precise than that. Eventually:
`Hdf5ConvertibleType =  Dict[str, Union['Hdf5ConvertibleType', np.ndarray, torch.Tensor, ...]]`, or if you don't support nested dictionary: `Dict[str, Union[np.ndarray, torch.Tensor, ...]]`.

Typing is supported to be informative for the end-user AND  the developer, not a chore for the developer.

tianshou/data/utils/converter.py
```diff
@@ -80,3 +82,58 @@ def to_torch_as(
...
+        # resort to pickle for any other type of object
+        else:
+            int_data = np.fromstring(pickle.dumps(v), dtype="uint8")
+            y.create_dataset(k, data=int_data, dtype="uint8")
+            y[k].attrs["__data_type__"] = "pickle"
+
+
+def from_hdf5(
+    x: h5py.Group, device: Optional[str] = None
+) -> Any:
```
duburcqa: same comment.

tianshou/data/utils/converter.py
nicoguertler: Thank you for your feedback. I do support nested dictionaries. It is a bit difficult to say precisely what can be saved this way as the function falls back to using pickle. So it would actually be something like 
```python
Hdf5ConvertibleType = 
Dict[str, Union[int, float, Batch, np.ndarray, torch.Tensor, 'anything that can be pickled', 'Hdf5ConvertibleType']]
```
I am not sure how to best express that.

tianshou/data/buffer.py
nicoguertler: Thank you for pointing that out. I tried to address the robustness concerns by pickling objects that cannot be handled directly.

tianshou/data/utils/converter.py
duburcqa: Hum OK I see the issue. There is no type associated with 'pickle-able'. Indeed, any `object` is `pickle-able`, so you must use:

```python
Hdf5ConvertibleType = 
Dict[str, Union[int, float, Batch, np.ndarray, torch.Tensor, object, 'Hdf5ConvertibleType']]
```

Or, if you want to avoid redundancy (since everything derives from `object`):

```python
Hdf5ConvertibleType =  Dict[str, Any]
```

Unfortunately, you cannot do any better... Personally I would rather go for the first one, because it carries the intend, but it is a matter of taste. 

By the way, thank you for your contribution, it is very much appreciated.

tianshou/data/utils/converter.py
```diff
@@ -80,3 +82,62 @@ def to_torch_as(
...
+) -> Hdf5ConvertibleType:
+    """Restore object from HDF5 group."""
+    # handle datasets
+    if isinstance(x, h5py.Dataset):
+        if x.attrs["__data_type__"] == "numpy":
+            y = np.empty(x.shape, dtype=x.dtype)
+            x.read_direct(y)
+        elif x.attrs["__data_type__"] == "torch":
+            y = torch.tensor(x, device=device)
+        elif x.attrs["__data_type__"] == "pickle":
```
duburcqa: To me, "object" is clearer than "pickle". It makes it explicit that it is a fallback and it is an actual type. I don't know about you.

tianshou/data/utils/converter.py
```diff
@@ -80,3 +82,62 @@ def to_torch_as(
...
+            y = pickle.loads(x[()])
+    # handle groups representing a dict or a batch
+    else:
+        y = {k: v for k, v in x.attrs.items() if k != "__convert_to__"}
+        for k, v in x.items():
+            y[k] = from_hdf5(v, device)
+        if "__convert_to__" in x. attrs:
+            # if dictionary represents Batch have to convert to Batch
+            if x.attrs["__convert_to__"] == "Batch":
```
duburcqa: Why not just consider "Batch" as another `__data_type__`, instead of using a dedicated field `__convert_to__` ?

tianshou/data/utils/converter.py
nicoguertler: Thank you for your advice and for helping me with the code! I have implemented your suggestion.

tianshou/data/utils/converter.py
nicoguertler: Makes sense. I will change the string to "object".

tianshou/data/utils/converter.py
```diff
@@ -80,3 +82,64 @@ def to_torch_as(
...
+
+
+def from_hdf5(
+    x: h5py.Group, device: Optional[str] = None
+) -> Hdf5ConvertibleType:
+    """Restore object from HDF5 group."""
+    # handle datasets
+    if isinstance(x, h5py.Dataset):
+        if x.attrs["__data_type__"] == "numpy":
+            y = np.empty(x.shape, dtype=x.dtype)
```
duburcqa: I would suggest you to store the _"fully qualified type"_ in `__data_type__`. It would be much easier to maintain and more clear to me. Here is an example:
```python
from pydoc import locate  # Standard library helper method
object_type = locate("torch.Tensor")  # or "torch.tensor", the same goes for "numpy.array"/"numpy.ndarray" 
object = object_type(x, device=device)
```

This way you don't need any `elif`, just one `else` for pickle fallback.

It could be something very generic as: `y = locate(x.attrs["__data_type__"])(**x.kwargs)`. What do you think?

tianshou/data/utils/converter.py
nicoguertler: Ok, agreed. The code will be more consistent that way.

tianshou/data/utils/converter.py
nicoguertler: Sounds good. I agree, that the code can probably be simplified. ~~However, I probably still need a separate clause for torch.Tensor as I have to copy to cpu first and convert to numpy.~~ Ah, I can use to_numpy for that...
In the case of a numpy array I think using read_direct could reduce the memory overhead. I'll have to test whether it really makes a difference though.

duburcqa: > However, I probably still need a separate clause for torch.Tensor as I have to copy to cpu first and convert to numpy.

Indeed. So, what you are doing right now is probably the way to go. Trying to generalize something that it not generic is often more harmful than the contrary. So maybe you should keep it like that.

duburcqa: > In the case of a numpy array I think using read_direct could reduce the memory overhead. I'll have to test whether it really makes a difference though.

Do as you wish. To me speed and memory overhead/footprint is not a big deal in this case. Just keep it simple and clear, so that we can build on it more easily in the future.

tianshou/data/buffer.py
```diff
@@ -359,6 +379,22 @@ def __getitem__(
...
+
+    @classmethod
+    def load_hdf5(
+        cls, path: str, device: Optional[str] = None
+    ) -> "ReplayBuffer":
+        """Load replay buffer from HDF5 file."""
+        with h5py.File(path, "r") as f:
+            buf = cls.__new__(cls)
+            buf_state = from_hdf5(f, device=device)
+            buf.__setstate__(buf_state)
```
duburcqa: Why not `buf.__setstate__(from_hdf5(f, device=device))` as you did for `save_hdf5` but reversed ?

test/base/test_buffer.py
```diff
@@ -278,7 +280,79 @@ def test_pickle():
...
+        pbuf.add(
+            obs=Batch(index=np.array([i])),
+            act=i,
+            rew=rew,
+            done=0,
+            info={"number": {"n": i}},
+            weight=np.random.rand()
+        )
+    buffers = {
+            "v": vbuf,
```
duburcqa: I know it is just a unit test, but could you make those fields a little bit more explicit ?

duburcqa (dismissed):

test/base/test_buffer.py
nicoguertler: Ok, I will go over the test again once I have given your suggestions for simplifying the code a try.

tianshou/data/buffer.py
nicoguertler: Sorry, I don't understand. Isn't this what I have done?

tianshou/data/utils/converter.py
```diff
@@ -80,3 +82,64 @@ def to_torch_as(
...
+            y[k].attrs["__data_type__"] = "numpy"
+        elif isinstance(v, torch.Tensor):
+            y.create_dataset(k, data=v.cpu().numpy())
+            y[k].attrs["__data_type__"] = "torch"
+        # ints and floats are stored as attributes of groups
+        elif isinstance(v, (int, float)):
+            y.attrs[k] = v
+        # resort to pickle for any other type of object
+        else:
+            int_data = np.fromstring(pickle.dumps(v), dtype="uint8")
```
duburcqa: I guess this could fail ? If so, wrap it in a `try ... except ...` block, and raise `NotImplementedError`, describing that the object is not compatible with hdf5 import/export format. Try to make as clear as possible for the user to understand what he did wrong.

Since you decided to go for a generic converter, it is likely to be extended in the future, and raising `NotImplementError` is a very clear hint for the user that it could work by implementing the missing feature. Otherwise it would be a misuse of the method. Both cases should be distinguished.

tianshou/data/utils/converter.py
```diff
@@ -80,3 +82,64 @@ def to_torch_as(
...
+            y.create_dataset(k, data=int_data, dtype="uint8")
+            y[k].attrs["__data_type__"] = "object"
```
duburcqa: Now that I'm thinking twice about it, maybe the actual class type would be more informative. Why not to store `v.__class__` or `v.__name__`, then replace `elif x.attrs["__data_type__"] == "object":` by `else` ? Because otherwise there is no way to know in advance type of the object. But that's just a recommendation. Don't feel the need to follow me on this if you disagree.

tianshou/data/buffer.py
duburcqa: I'm just being extra picky. Here you wrote it in 2 lines of code, whereas you wrote it in 1 in `save_hdf5`. Choose one or the other and stick on it for both methods to be consistent.

nicoguertler: Ah, I see. I will put everything into one line, thanks.

tianshou/data/utils/converter.py
nicoguertler: Yes, makes sense. I have added a try-catch block with a `NotImplementedError`. Let me know if the message makes sense to you.

tianshou/data/utils/converter.py
nicoguertler: Ok, that actually simplifies the code as well. So I have used `v.__class__.__name__` instead of custom strings. This way it is indeed clearer what the HDF5 file contains.

tianshou/data/utils/converter.py
```diff
@@ -97,27 +97,31 @@ def to_hdf5(x: Hdf5ConvertibleType, y: h5py.Group) -> None:
...
         elif isinstance(v, (int, float)):
             y.attrs[k] = v
         # resort to pickle for any other type of object
         else:
-            int_data = np.fromstring(pickle.dumps(v), dtype="uint8")
+            try:
+                int_data = np.frombuffer(pickle.dumps(v), dtype="uint8")
+            except Exception as e:
+                raise NotImplementedError(
+                        "No coonversion to HDF5 for object of type "
```
duburcqa: "coonversion"

duburcqa:
It looks perfect to me ! Ready to be merged :) Very good job @nicoguertler

tianshou/data/utils/converter.py
nicoguertler: Ah, thanks for pointing that out!

nicoguertler:
Great! Thank you @duburcqa and @Trinkle23897 for your help and advice! There is still some room for improvement. For example, there is no custom conversion for the `SegmentTree` so for now it is just pickled. But it should be easy to add that later on.

duburcqa (dismissed):

duburcqa:
> There is still some room for improvement. For example, there is no custom conversion for the SegmentTree so for now it is just pickled. But it should be easy to add that later on.

Yes, but the current implementation can be extended very easily, which is great !

tianshou/data/utils/converter.py
```diff
@@ -80,3 +82,68 @@ def to_torch_as(
...
+            y.create_dataset(k, data=to_numpy(v))
+            y[k].attrs["__data_type__"] = v.__class__.__name__
+        elif isinstance(v, (int, float)):
+            # ints and floats are stored as attributes of groups
+            y.attrs[k] = v
+        else:  # resort to pickle for any other type of object
+            try:
+                int_data = np.frombuffer(pickle.dumps(v), dtype="uint8")
+            except Exception as e:
+                raise NotImplementedError(
```
Trinkle23897, line 143: Could you please add some tests to cover these two lines of code?

tianshou/data/utils/converter.py
```diff
@@ -80,3 +82,68 @@ def to_torch_as(
...
+        elif isinstance(v, (np.ndarray, torch.Tensor)):
+            # NumPy arrays and PyTorch tensors are written to datasets
+            y.create_dataset(k, data=to_numpy(v))
+            y[k].attrs["__data_type__"] = v.__class__.__name__
+        elif isinstance(v, (int, float)):
+            # ints and floats are stored as attributes of groups
+            y.attrs[k] = v
+        else:  # resort to pickle for any other type of object
+            try:
+                int_data = np.frombuffer(pickle.dumps(v), dtype="uint8")
```
Trinkle23897: I don't quite understand why we need "uint8" here.

Trinkle23897 (commented):
Nice implementation!

test/base/test_buffer.py
```diff
@@ -278,7 +280,60 @@ def test_pickle():
...
+    buffer_types = {k: b.__class__ for k, b in buffers.items()}
+    device = 'cuda' if torch.cuda.is_available() else 'cpu'
+    rew = torch.tensor([1.]).to(device)
+    for i in range(4):
+        kwargs = {
+            'obs': Batch(index=np.array([i])),
+            'act': i,
+            'rew': rew,
+            'done': 0,
+            'info': {"number": {"n": i}, 'extra': None},
```
Trinkle23897, line 301: `None` is very common in gym's `info`, however, the current implementation does not support it.

```
step0: info: {}
step1: info: {}
step2: info: {'timelimit.policy': 1}
```

tianshou/data/utils/converter.py
nicoguertler: Thank you for cleaning up the test. I have added tests for two exceptions. Let me know if this is what you had in mind.

tianshou/data/utils/converter.py
nicoguertler: I used `frombuffer` because HDF5's handling of strings excludes strings with null bytes. However, the data type wasn't inferred correctly somehow. If I don't specify a data type I get `ValueError: buffer size must be a multiple of element size`. I can see that "uint8" is a confusing choice though. I have replaced it with `np.byte`.

test/base/test_buffer.py
nicoguertler: Good point. I suggest that the the converter first tries to write numpy arrays to HDF5 directly and, if that fails, falls back to pickling the array instead. This added a bit of `try`-`except` code that maybe looks a bit confusing. Let me know what you think.

tianshou/data/utils/converter.py
```diff
@@ -107,21 +113,32 @@ def to_hdf5(x: Hdf5ConvertibleType, y: h5py.Group) -> None:
             subgrp.attrs["__data_type__"] = v.__class__.__name__
             to_hdf5(subgrp_data, subgrp)
         elif isinstance(v, (np.ndarray, torch.Tensor)):
```
duburcqa: If you keep this condition as is, I don't see why `y.create_dataset(k, data=to_numpy(v))` could fail.

tianshou/data/utils/converter.py
```diff
@@ -94,6 +94,12 @@ def to_torch_as(
 Hdf5ConvertibleType = Dict[str, Hdf5ConvertibleValues]  # type: ignore
 
 
+def to_hdf5_via_pickle(x: object, y: h5py.Group, key: str) -> None:
+    """Pickle, convert byte stream to numpy array and write to HDF5 dataset."""
+    data = np.frombuffer(pickle.dumps(x), dtype=np.byte)
```
duburcqa: I would rather make it a nested method, since it is never intended to be called manually, and it is very short.

tianshou/data/utils/converter.py
nicoguertler: One problem are arrays with `dtype=object`. For example, if `None` appears in the info from the gym environment, this happens as Trinkle23897 pointed out. Possibly there are also other cases which are not directly supported by HDF5 like structured arrays. To treat all of them explicitly might be difficult.

tianshou/data/utils/converter.py
nicoguertler: You are right it belongs into `to_hdf5`. I have turned it into an inner function.

tianshou/data/utils/converter.py
duburcqa: Ok I see. Why not then. Just add a comment for us to remember the initial motivation of catching exceptions here.

tianshou/data/utils/converter.py
```diff
@@ -80,3 +82,85 @@ def to_torch_as(
...
+                # NumPy arrays and PyTorch tensors are written to datasets
+                y.create_dataset(k, data=to_numpy(v))
+                y[k].attrs["__data_type__"] = v.__class__.__name__
+            except TypeError:
+                # if data type is not supported by HDF5 fall back to pickle
+                try:
+                    to_hdf5_via_pickle(v, y, k)
+                except Exception as e:
+                    raise NotImplementedError(
+                        f"Attempted to pickle {v.__class__.__name__} with key"
```
duburcqa: This exception is not very clear in this case.

tianshou/data/utils/converter.py
duburcqa: After rethinking about it, I don't think such behavior can happen with `torch.Tensor`. They are way more restrictive than Numpy array, so just catching the exception for numpy.ndarray should be fine and help to make the exception more clear by reducing the potential root causes.

tianshou/data/utils/converter.py
nicoguertler: Maybe a `NotImplementedError` isn't the right choice. Would you prefer a `RuntimeError` with a message like `"Fallback to pickle failed for {v.__class__.__name__} (with key '{k}') which cannot be written to a HDF5 dataset directly."`?

duburcqa: Yes `RuntimeError` is better. And instead of mentioning 'key', I would clarify that it is because the datatype of the array is not supported, as you explained it to be previously.

tianshou/data/utils/converter.py
nicoguertler: Ok, agreed. I have split the elif statement up into two cases for Tensors and ndarrays.

tianshou/data/utils/converter.py
nicoguertler: Makes sense. I have replaced the `NotImplementedError` with a `RuntimeError` and removed the part of the message mentioning the key.

tianshou/data/utils/converter.py
```diff
@@ -80,3 +82,90 @@ def to_torch_as(
...
+
+    for k, v in x.items():
+        if isinstance(v, (Batch, dict)):
+            # dicts and batches are both represented by groups
+            subgrp = y.create_group(k)
+            if isinstance(v, Batch):
+                subgrp_data = v.__getstate__()
+            else:
+                subgrp_data = v
+            subgrp.attrs["__data_type__"] = v.__class__.__name__
```
duburcqa, line 113: "ndarray" for consistency with `from_hdf5`

tianshou/data/utils/converter.py
```diff
@@ -80,3 +82,90 @@ def to_torch_as(
...
+            to_hdf5(subgrp_data, subgrp)
+        elif isinstance(v, torch.Tensor):
+            # PyTorch tensors are written to datasets
+            y.create_dataset(k, data=to_numpy(v))
+            y[k].attrs["__data_type__"] = v.__class__.__name__
```
duburcqa: "Tensor" for consistency with `from_hdf5`

tianshou/data/utils/converter.py
```diff
@@ -80,3 +82,90 @@ def to_torch_as(
...
+        elif isinstance(v, np.ndarray):
+            try:
+                # NumPy arrays are written to datasets
+                y.create_dataset(k, data=v)
+                y[k].attrs["__data_type__"] = v.__class__.__name__
```
duburcqa: "ndarray" for consistency with `from_hdf5`

tianshou/data/utils/converter.py
duburcqa: "Batch" for consistency with `from_hdf5`.

nicoguertler: Ok, only using `__class__.__name__` when its value is not already clear makes the code more readable. I have changed the corresponding lines.

duburcqa (approved):

duburcqa:
Looks good to me now @Trinkle23897

Trinkle23897 (approved):

PR #263, opened 2020-12-15
Add offline trainer and discrete BCQ algorithm

### LINKED ISSUE #188: The best practice using Tianshou for offline RL?
    + [ ] exception-raising bug
    + [ ] RL algorithm bug
    + [ ] new feature request
  ```python
  import tianshou, torch, sys
  ```
What might be a best practice using Tianshou for offline RL?
Save to the buffer and pickle it to storage and then use it as an offline replay buffer (no further data gathering) for algorithm training?

offline-RL

This issue has some relevant discussion but I am not talking about TPU. https://github.com/thu-ml/tianshou/issues/24#issuecomment-614466071

### LINKED ISSUE #248: Questions about the imitation learning
    + [ ] exception-raising bug
    + [ ] RL algorithm bug
    + [ ] new feature request
  ```python
  import tianshou, torch, sys
  ```
Now I am training a new network with no pre-trained weights, while I meet with the action asturation problem(the agent only take one action). Therefore, I prepare to train the network with supervised data(generate from another optimization algorithm). So how could I send the supervised data to tianshou, or I have to write another script for training. Thanks for any help!
By the way, I currently use the PPO with onpolicy trainer, and is there any examples to #188 if it could solve my problem ?

zhujl1991:
Discrete BCQ: https://arxiv.org/abs/1910.01708
Offline trainer discussion: https://github.com/thu-ml/tianshou/issues/248#issuecomment-744921908

Will implement a test_imitation.py in the next PR.


Trinkle23897:
> Will implement a test_imitation.py in the next PR.

You can add here.

tianshou/policy/modelfree/bcq.py
```diff
@@ -0,0 +1,138 @@
...
+                imt = (imt / imt.max(1, keepdim=True)[0] > self._tau).float()
+
+                # Use large negative number to mask actions from argmax
+                next_action = (imt * q + (1 - imt) * -1e8).argmax(1, keepdim=True)
+
+                q, _, _ = self._target_net(non_final_next_states)
+                q = q.gather(1, next_action).reshape(-1, 1)
+
+                next_state_values = torch.zeros(len(batch), device=self._device).float()
+                next_state_values[non_final_mask] = q.squeeze()
```
Trinkle23897: Current script cannot run on cuda device.
```
  File "test/discrete/test_bcq.py", line 100, in <module>
    test_bcq(get_args())
  File "test/discrete/test_bcq.py", line 94, in test_bcq
    args.test_frequency,
  File "/home/trinkle/github/tianshou-new/tianshou/trainer/offline.py", line 30, in offline_trainer
    loss = policy.update(batch_size, buffer)
  File "/home/trinkle/github/tianshou-new/tianshou/policy/base.py", line 171, in update
    result = self.learn(batch, **kwargs)
  File "/home/trinkle/github/tianshou-new/tianshou/policy/modelfree/bcq.py", line 123, in learn
    next_state_values[non_final_mask] = q.squeeze()
RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu!
```

test/discrete/test_bcq.py
```diff
@@ -0,0 +1,99 @@
...
+    buffer = ReplayBuffer(size=args.buffer_size)
+    for i in range(args.buffer_size):
+        buffer.add(
+            obs=torch.rand(state_shape),
+            act=random.randint(0, action_shape - 1),
+            rew=1,
+            done=False,
+            obs_next=torch.rand(state_shape),
+            info={},
+        )
```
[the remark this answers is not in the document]
zhujl1991: Yes, this is more of a demo because we don't have real training data. But actually, we have already tested the algorithm on an internal dataset.

tianshou/trainer/offline.py
```diff
@@ -0,0 +1,53 @@
...
+    best_reward = -1
+    best_policy = policy
+    total_iter = 0
+    iter_per_epoch = len(buffer) // batch_size
+
+    for epoch in range(1, 1 + epochs):
+        for iter in range(iter_per_epoch):
+            total_iter += 1
+            loss = policy.update(batch_size, buffer)
+            if total_iter % test_frequency == 0:
```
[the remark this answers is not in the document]
zhujl1991: The concern here is if the buffer is huge, we only need to use few epochs to train. So the test runs too few times.

tianshou/policy/modelfree/bcq.py
```diff
@@ -0,0 +1,138 @@
...
+            self.i3 = nn.Linear(imitation_model_hidden_dim, n_actions)
+
+        def forward(self, state):
+            q = F.relu(self.q1(state))
+            q = F.relu(self.q2(q))
+
+            i = F.relu(self.i1(state))
+            i = F.relu(self.i2(i))
+            i = F.relu(self.i3(i))
+            return self.q3(q), F.log_softmax(i, dim=1), i
```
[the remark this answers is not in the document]
zhujl1991: What do you mean by moving to `test/`? I think in other modules, taking dqn for example, it uses the pytorch model defined in `tianshou/utils/net/common.py`. Are you saying I should move the pytorch model definition here to `tianshou/utils/net/common.py` as well? The reason I define it here is that the pytorch model used in BCQ is different from others -- it actually includes two nets instead of one. So it's pretty much BCQ specific and not common at all.

tianshou/policy/modelfree/bcq.py
zhujl1991: Will have it tested on a GPU machine.

tianshou/policy/modelfree/bcq.py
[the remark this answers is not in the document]
Trinkle23897: I think you can take a look at the current code of TD3/SAC. They both use Q-learning and double critic network.

tianshou/trainer/offline.py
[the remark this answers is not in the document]
Trinkle23897: So why not use `stop_fn`?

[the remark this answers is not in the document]
zhujl1991: Can you elaborate here?

zhujl1991:
PR updated. But any idea how to fix this by mypy?

```
tianshou/policy/modelfree/bcq.py: note: In member "learn" of class "BCQPolicy":
tianshou/policy/modelfree/bcq.py:71: error: "Batch" has no attribute "__iter__"
(not iterable)  [attr-defined]
                    [s.obs.unsqueeze(0) for s in batch if not s.done], dim...
```

Trinkle23897:
> PR updated. But any idea how to fix this by mypy?
> 
> ```
> tianshou/policy/modelfree/bcq.py: note: In member "learn" of class "BCQPolicy":
> tianshou/policy/modelfree/bcq.py:71: error: "Batch" has no attribute "__iter__"
> (not iterable)  [attr-defined]
>                     [s.obs.unsqueeze(0) for s in batch if not s.done], dim...
> ```

I can fix it, don't need to worry about that.

zhujl1991:
> > PR updated. But any idea how to fix this by mypy?
> > ```
> > tianshou/policy/modelfree/bcq.py: note: In member "learn" of class "BCQPolicy":
> > tianshou/policy/modelfree/bcq.py:71: error: "Batch" has no attribute "__iter__"
> > (not iterable)  [attr-defined]
> >                     [s.obs.unsqueeze(0) for s in batch if not s.done], dim...
> > ```
> 
> I can fix it, don't need to worry about that.

OK. Do you have any feedback on this PR? I'll fix other mypy error except the one above.

zhujl1991:
```
$python test/discrete/test_bcq.py 
  File "test/discrete/test_bcq.py", line 1, in <module>
    from tianshou.policy import BCQPolicy
  File "/home/jialu.zhu/work/tianshou/tianshou/__init__.py", line 1, in <module>
    from tianshou import data, env, utils, policy, trainer, exploration
  File "/home/jialu.zhu/work/tianshou/tianshou/data/__init__.py", line 2, in <module>
    from tianshou.data.utils.converter import to_numpy, to_torch, to_torch_as
  File "/home/jialu.zhu/work/tianshou/tianshou/data/utils/converter.py", line 1, in <module>
    import h5py
ModuleNotFoundError: No module named 'h5py'
```

This error is not from my PR. Can you help here? @Trinkle23897

Trinkle23897:
> OK. Do you have any feedback on this PR? I'll fix other mypy error except the one above.

I'll have a look this afternoon.

> This error is not from my PR. Can you help here? @Trinkle23897

Just `pip install h5py` because in #261 it introduced buffer save/load by hdf5 data format. The `setup.py` has changed accordingly.

tianshou/policy/__init__.py
```diff
@@ -10,6 +10,7 @@
 from tianshou.policy.modelfree.td3 import TD3Policy
 from tianshou.policy.modelfree.sac import SACPolicy
 from tianshou.policy.modelfree.discrete_sac import DiscreteSACPolicy
+from tianshou.policy.modelfree.bcq import BCQPolicy
```
Trinkle23897: Imitation learning's code should be under `policy/imitation`

tianshou/policy/modelfree/bcq.py
```diff
@@ -0,0 +1,124 @@
...
+    ) -> Batch:
+        batch.to_torch()
+
+        state = batch.obs.to(self._device)
+        q, imt, _ = self._policy_net(state.float())
+        imt = imt.exp()
+        imt = (imt / imt.max(1, keepdim=True)[0] > self._tau).float()
+
+        # Use large negative number to mask actions from argmax
+        action = (imt * q + (1.0 - imt) * -1e8).argmax(1)
```
Trinkle23897: replace 1e8 with np.inf

tianshou/trainer/offline.py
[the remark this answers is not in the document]
Trinkle23897: In onpolicy_trainer and offpolicy_trainer, the `stop_fn` is a function that receives avg test reward and outputs whether or not it reaches the threshold.

test/discrete/test_bcq.py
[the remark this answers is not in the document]
Trinkle23897: Can we provide some data from the oracle and add the assertion of expected performance at the end of this test?

tianshou/policy/modelfree/bcq.py
```diff
@@ -0,0 +1,124 @@
...
+        tau: float,
+        target_update_frequency: int,
+        device: str,
+        gamma: float,
+        imitation_logits_penalty: float,
+        **kwargs: Any,
+    ) -> None:
+        super().__init__(**kwargs)
+        self._policy_net = model
+        self._policy_net.to(device)
```
Trinkle23897: I suggest that `to(device)` should not stay at the policy level. Either in the test code `net.to(args.device)` or `policy.to(args.device)`, or pass this `device` argument into the network (BCQN below) like the other examples in `utils/net`, so that you don't need to care about the device in `BCQPolicy`. That would make the code cleaner.

tianshou/policy/modelfree/bcq.py
```diff
@@ -0,0 +1,124 @@
...
+
+from tianshou.data import Batch
+from tianshou.policy import BasePolicy
+from copy import deepcopy
+
+
+class BCQPolicy(BasePolicy):
+    """Implementation discrete BCQ algorithm. Some code is from
+    https://github.com/sfujim/BCQ/tree/master/discrete_BCQ
+
```
Trinkle23897: What's the meaning of each argument?

tianshou/policy/modelfree/bcq.py
```diff
@@ -0,0 +1,124 @@
...
+            tuple(map(lambda s: not s, batch.done)),
+            device=self._device,
+            dtype=torch.bool,
+        )
+
+        non_final_next_states = None
+        try:
+            non_final_next_states = torch.cat(
+                [s.obs.unsqueeze(0) for s in batch if not s.done], dim=0
+            )
```
Trinkle23897:
```suggestion
            non_final_next_states = batch.obs[~batch.done]
```
so that if all data points are `done==True`, the `non_final_next_states` will be an empty tensor instead of None.

tianshou/policy/modelfree/bcq.py
```diff
@@ -0,0 +1,124 @@
...
+        batch.to_torch()
+
+        non_final_mask = torch.tensor(
+            tuple(map(lambda s: not s, batch.done)),
+            device=self._device,
+            dtype=torch.bool,
+        )
+
+        non_final_next_states = None
+        try:
```
Trinkle23897: Why use try...exception here?

tianshou/policy/modelfree/bcq.py
```diff
@@ -0,0 +1,124 @@
...
+            )
+            non_final_next_states = non_final_next_states.to(self._device)
+        except Exception:
+            pass
+
+        # Compute the target Q value
+        with torch.no_grad():
+            expected_state_action_values = batch.rew.float().to(self._device)
+
+            # Add target Q value for non-final next_state
```
Trinkle23897: Can we extend BCQ to nstep setting instead of only one step?

tianshou/policy/modelfree/bcq.py
Trinkle23897: But I'm a little bit confused that why did you use `s.obs` instead of `s.obs_next`?

tianshou/policy/modelfree/bcq.py
```diff
@@ -0,0 +1,124 @@
...
+        return Batch(act=action, state=state, qvalue=q)
+
+    def learn(self, batch: Batch, **kwargs: Any) -> Dict[str, float]:
+        batch.to_torch()
+
+        non_final_mask = torch.tensor(
+            tuple(map(lambda s: not s, batch.done)),
+            device=self._device,
+            dtype=torch.bool,
+        )
```
Trinkle23897:
```suggestion
        non_final_mask = ~batch.done
```
Please use vectorized operation because that is cleaner and faster.

tianshou/policy/modelfree/bcq.py
```diff
@@ -0,0 +1,124 @@
...
+                next_action = (imt * q + (1 - imt) * -1e8).argmax(
+                    1, keepdim=True
+                )
+                q, _, _ = self._target_net(non_final_next_states)
+                q = q.gather(1, next_action).reshape(-1, 1)
+
+                next_state_values = torch.zeros(
+                    len(batch), device=self._device
+                ).float()
+                next_state_values[non_final_mask] = q.squeeze()
```
Trinkle23897: These lines of code are very similar to `self.forward`. Could you use `self(...)` instead of duplicating again?

lorepieri8:
Looking forward to test the offline trainer.

Trinkle23897:
I'll update this PR tmr.

Trinkle23897:
@zhujl1991 I use the [author's version](https://github.com/sfujim/BCQ/tree/master/discrete_BCQ) and [d3rlpy](https://github.com/takuseno/d3rlpy/blob/b8fa339292778e669871a759d9339bd6a7fcb0e9/tests/algos/test_bcq.py#L42-L46) to play with CartPole-v0 (by given expert data), but none of them can train a DiscreteBCQ agent that can reach the expert level. It's very weird.

zhujl1991:
> @zhujl1991 I use the [author's version](https://github.com/sfujim/BCQ/tree/master/discrete_BCQ) and [d3rlpy](https://github.com/takuseno/d3rlpy/blob/b8fa339292778e669871a759d9339bd6a7fcb0e9/tests/algos/test_bcq.py#L42-L46) to play with CartPole-v0 (by given expert data), but none of them can train a DiscreteBCQ agent that can reach the expert level. It's very weird.

Have you tried with different seed? How did you get the training data?

The original version of BCQ works for us. But internally we actually use a modified version, which gives a better result -- train imitation model first. then when training BCQ, keep the imitation model fixed. This is mentioned here https://danieltakeshi.github.io/2020/06/28/offline-rl/:

`I was confused about why BCQ does the behavior cloning update of ω inside the for loop, rather than beforehand. Since the data is fixed, this seems suboptimal since the optimization for θ will rely on an inaccurate model Gω during the first few iterations. After contacting Fujimoto, he agreed that it is probably better to move the optimization before the loop, but his results were not significantly better.`

Trinkle23897:
> Have you tried with different seed?

No, but I think offline learning should not be unstable as the online setting, i.e., its performance should be seed-invariant.

> How did you get the training data?

For author's version, I just replace the task with "CartPole-v0" and run the 3 commands; for d3rlpy, I run the given test scripts (as shown in the link above, 4 lines).

Trinkle23897:
I found the problem: in the author's version, the imitation network has a ReLU layer in the end, however, this is definitely wrong because it will cause dead ReLU problem. You can see [Conv](https://github.com/sfujim/BCQ/blob/master/discrete_BCQ/discrete_BCQ.py#L30-L31) network has no ReLU layer in the final but [FC](https://github.com/sfujim/BCQ/blob/master/discrete_BCQ/discrete_BCQ.py#L53-L54) has a ReLU layer. The code can work on Atari but cannot work on toy examples due to this reason.

Trinkle23897:
ready for review, and please have a look @zhujl1991

tianshou/trainer/offline.py
```diff
@@ -0,0 +1,95 @@
...
+) -> Dict[str, Union[float, str]]:
+    """A wrapper for offline trainer procedure.
+
+    The "step" in trainer means a policy network update.
+
+    :param policy: an instance of the :class:`~tianshou.policy.BasePolicy`
+        class.
+    :param test_collector: the collector used for testing.
+    :type test_collector: :class:`~tianshou.data.Collector`
+    :param int max_epoch: the maximum of epochs for training. The training
```
duburcqa: maximum number of epochs

tianshou/trainer/offline.py
```diff
@@ -0,0 +1,95 @@
...
+        process might be finished before reaching the ``max_epoch``.
+    :param int step_per_epoch: the number of step for updating policy network
+        in one epoch.
+    :param episode_per_test: the number of episodes for one policy evaluation.
+    :param int batch_size: the batch size of sample data, which is going to
+        feed in the policy network.
+    :param function test_fn: a function receives the current number of epoch
```
duburcqa: A function defined by the user and called automatically by the algorithm is called a "hook", aka [Pytorch doc](https://pytorch.org/tutorials/beginner/former_torchies/nnft_tutorial.html#forward-and-backward-function-hooks).

:param function test_fn: a hook called at the beginning of testing in each epoch. It can be used to perform custom additional operations. It must have the following signature: f(num_peoch: int, step_idx: int) -> None.

tianshou/trainer/offline.py
```diff
@@ -0,0 +1,95 @@
...
+        and step index, and performs some operations at the beginning of
+        testing in this epoch.
+    :param function save_fn: a function for saving policy when the undiscounted
```
duburcqa: Same, it would be nice to emphasis they are hooks (same for the other trainers)

tianshou/trainer/offline.py
```diff
@@ -0,0 +1,95 @@
...
+        average mean reward in evaluation phase gets better.
+    :param function stop_fn: a function receives the average undiscounted
+        returns of the testing result, return a boolean which indicates whether
+        reaching the goal.
+    :param torch.utils.tensorboard.SummaryWriter writer: a TensorBoard
```
duburcqa, line 52: Perhaps precise what happens if none is given in argument.

tianshou/trainer/offpolicy.py
```diff
@@ -73,7 +74,7 @@ def offpolicy_trainer(
     """
     env_step, gradient_step = 0, 0
     best_epoch, best_reward, best_reward_std = -1, -1.0, 0.0
-    stat: Dict[str, MovAvg] = {}
+    stat: Dict[str, MovAvg] = defaultdict(MovAvg)
```
duburcqa, line 78: Nice trick !

duburcqa:
Done with review, nothing much to say :+1:

duburcqa (dismissed):

ChenDRAG:
I think to commit a new policy, experimental result and quantitative analyzsis/comparison with original paper should at least be provided to ensure the correctness of the algorithm.
Also, I worry that the way offline_trainer trains an agent isn't the general meaning of "offline training" in rl literature. (From what I understand offline training usually means that you update the agent after whole episode data is collected, rather than collecting all the data at first in imitation learning). ~So I suggest that offline trainer not added to `tianshou/data/trainer` because this is not what people usually use.(And it is not too hard to implement so not necessary to be officially supported). One option is to add this in `test/discrete/test_bcq.py`. If a lot of users need this 'offline trainer', we then consider officially supporting it.~

duburcqa:
> I think to commit a new policy, experimental result and quantitative analyzsis/comparison with original paper should at least be provided to ensure the correctness of the algorithm.

It would be nice yes, but apart from documenting the algorithms, I don't know any framework actually providing this kind of analysis (I mean extensively) since it is extremely time consuming. Yet it could be possible to compare the performance wrt to a signle other algorithm that is considered to be the state-of-the-art.

ChenDRAG:
> Yet it could be possible to compare the performance wrt to a signle other algorithm that is considered to be the state-of-the-art.

standard library of bcq algorithm is provided here, I think it is not hard to provide at least a fair and detailed comparison over one single environment.

duburcqa:
> standard library of bcq algorithm is provided here, I think it is not hard to provide at least a fair and detailed comparison over one single environment.

Of course, but that's the whole point. Fair and detailed seems to much work to me. Correctness and implementation design must be clear to the user, but the user is also expected to document himself and fond articles analyzing such benchmark.

ChenDRAG:
> > standard library of bcq algorithm is provided here, I think it is not hard to provide at least a fair and detailed comparison over one single environment.
> 
> Of course, but that's the whole point. Fair and detailed seems to much work to me. Correctness and implementation design must be clear to the user, but the user is also expected to document himself and fond articles analyzing such benchmark.

I do not mean to add extra work to developers, any experiments that can prove the correctness to a certain range is acceptable, but here is what I believe: algorithm whose correctness cannot be assured is actually a burden to Tianshou if officially supported. 

Currently, tianshou actually doesn't achieve very good results on large environments. Most of its policies are only demonstrated in toy environments like Cartpole, but many problems are cannot be exposed in toy environments. (Even on toy environments, I see there are still arguments about whether this algorithm works). In other words, tianshou lacks experiment and is a little bit hard to be used for research for now (Some issues on GitHub are already talking about this). This is a critical problem and is what I believe of the highest priority. 

I'm currently working on benchmarking mujoco environments using Tianshou, and found some small problems in policy or in tianshou/data.(will make a pr soon) I also found out that it is actually very hard to modify the code because any small change in tianshou/data you will have to take care of all policies officially supported by tianshou, Even if you do not really understand that algorithm. That's why I suggest at first that new policy can go to /test or /examples because to add code there problems will be a lot less. (You don't even need to change docs, etc). and time can be given to let users to try new policy.  We can consider make it officially supported after some time. If urgently needed, graphs that show the correctness or efficiency of code on **one single** environment does not seem like too much burder?

Trinkle23897:
I'm working on the result of BCQ with Atari games, don't worry about that :)

Trinkle23897:
@zhujl1991 have you successfully reproduce the result in the paper by the author's code? I tried with Pong and Breakout. It seems that Pong can easily reach +20 but Breakout cannot reach above 100 (most of the time it is around 30~60 and continuing going down).
I use 1e7 max timestamp for training DQN behavior agent with https://github.com/sfujim/BCQ/blob/master/discrete_BCQ/main.py.

zhujl1991:
> @zhujl1991 have you successfully reproduce the result in the paper by the author's code? I tried with Pong and Breakout. It seems that Pong can easily reach +20 but Breakout cannot reach above 100 (most of the time it is around 30~60 and continuing going down).
> I use 1e7 max timestamp for training DQN behavior agent with https://github.com/sfujim/BCQ/blob/master/discrete_BCQ/main.py.

I haven't tried to reproduce the result in the paper. We directly use BCQ for our own problem, which gives pretty much the same result as imitation learning.

Trinkle23897:
The Atari result should be updated after `done` issue fixed. Mark as TODO (@Trinkle23897, @ChenDRAG) and currently it is ready to be merged (after the utils.network updating PR #275).

duburcqa (approved):

tianshou/trainer/offline.py
```diff
@@ -0,0 +1,97 @@
...
+from tianshou.policy import BasePolicy
+from tianshou.utils import tqdm_config, MovAvg
+from tianshou.data import Collector, ReplayBuffer
+from tianshou.trainer import test_episode, gather_info
+
+
+def offline_trainer(
+    policy: BasePolicy,
+    buffer: ReplayBuffer,
+    test_collector: Collector,
```
zhujl1991, line 16: @Trinkle23897 I just noticed the test_collector, which needs to be initialized by an env, here is not optional. But actually, in practice, the main reason to use these offline algorithms is the lack of an env. So it might be better to make it optional. But I'm not sure what the alternative way to do the test given we don't have an env.

test/discrete/test_il_bcq.py
```diff
@@ -0,0 +1,111 @@
...
+    parser.add_argument("--seed", type=int, default=1626)
+    parser.add_argument("--eps-test", type=float, default=0.001)
+    parser.add_argument("--lr", type=float, default=3e-4)
+    parser.add_argument("--gamma", type=float, default=0.9)
+    parser.add_argument("--n-step", type=int, default=3)
+    parser.add_argument("--target-update-freq", type=int, default=320)
+    parser.add_argument("--unlikely-action-threshold", type=float, default=0.3)
+    parser.add_argument("--imitation-logits-penalty", type=float, default=0.01)
+    parser.add_argument("--epoch", type=int, default=5)
+    parser.add_argument("--step-per-epoch", type=int, default=1000)
```
zhujl1991, line 29: @Trinkle23897 Shouldn't this parameter equal to `len(buffer) // batch_size` all the time?

Trinkle23897: Well, the definition of `epoch` here is aligned with onpolicy_trainer and offpolicy_trainer. You can change the `step_per_epoch` param to control your evaluation frequency.

tianshou/trainer/offline.py
Trinkle23897: Hmm... remove the test_collector and set it to an optional argument mean we don't have any evaluation metric to measure the performance of the current policy. If users don't have any runnable envs, he/she can give self-definied fake envs to test_collector.

zhujl1991: All right. But I feel like that is sort of hacky. Anyway, let's leave it as it is here.

PR #266, opened 2020-12-25
Add C51 algorithm

shengxiang19:
Distributional RL algorithms are very powerful in atari environments. I am going to implement a series of typical algorithms, i.e. C51, QR-DQN, IQN, FQF, based on the  reinforcement learning platform Tianshou.

This is my frist PR for C51algorithm: [https://arxiv.org/abs/1707.06887](url)

1. add C51 policy in tianshou/policy/modelfree/c51.py.
2. add C51 net in tianshou/utils/net/discrete.py.
3. add C51 atari example in examples/atari/atari_c51.py.
4. add C51 statement in tianshou/policy/__init__.py.
5. add C51 test in test/discrete/test_c51.py.
6. add C51 atari results in examples/atari/results/c51/.

By running "python3 atari_c51.py --task "PongNoFrameskip-v4" --batch-size 64", get  best_result': '20.50 ± 0.50', in epoch 9.

By running "python3 atari_c51.py --task "BreakoutNoFrameskip-v4" --n-step 1 --epoch 40", get best_reward: 407.400000 ± 31.155096 in epoch 39.


Trinkle23897:
Nice job! Could you please also modify the `README.md` and `docs/index.rst` (add C51 description)?

Trinkle23897:
That's ok I can help you fix the PEP8.

shengxiang19:
Thank you very much. I will modify the README.md and docs/index.rst in my next PR.

Trinkle23897:
> Thank you very much. I will modify the README.md and docs/index.rst in my next PR.

Just in this PR is okay.

shengxiang19:
> > Thank you very much. I will modify the README.md and docs/index.rst in my next PR.
> 
> Just in this PR is okay.

I'm not very good at GitHub, and I could not find how to do it  in this PR.

Trinkle23897:
Okay, that's fine :) I'll take a look later on.

shengxiang19:
All checks have passed :) It was a tough journey.

tianshou/policy/modelfree/c51.py
```diff
@@ -0,0 +1,225 @@
...
+                         reward_normalization, **kwargs)
+        self._num_atoms = num_atoms
+        self._v_min = v_min
+        self._v_max = v_max
+        self.support = torch.linspace(self._v_min, self._v_max,
+                                      self._num_atoms)
+        self.delta_z = (v_max - v_min) / (num_atoms - 1)
+
+    @staticmethod
+    def prepare_n_step(
```
Trinkle23897: I'm not quite sure what's the difference between this n_step reimplementation and the existing implementation. Could you please explain more?
Also, a good way is to modify the original implementation instead of a duplicated new one.

Trinkle23897 (commented):
Also, could you please add a test script for CartPole-v0 under `test/discrete/`?

tianshou/policy/modelfree/c51.py
shengxiang19: the existing implementation directly calculates the target_q.  This n_step reimplementation just calculates the n-step return, S_{t+n} and done_{t+n}.

shengxiang19: In Distributional RL, the target_q is two dimensional array. I think it is inefficient to convert this array many times.

Trinkle23897: So why not extend the original implementation? If it is inefficient in your case, change it!
You don't need to care about the `PEP8, Types and Docs Check` test, I will fix that if fails.

shengxiang19: Since the original implementation is used in most Q-learing based algorithms, I'm cautious about modifying it. Maybe I will change it in next PR after I finish the Atari tests on C51.

Trinkle23897: > I'm cautious about modifying it.

Don't worry about that, I will take care of it. If the new implementation can pass all the tests under `test/base/test_returns.py`, it is probably no problem with other existing components.

shengxiang19: OK. I'll try to modify it later.

Trinkle23897:
> Will implement a test/discrete/test_c51.py in the next PR.

I think you can directly add this file. No need to make a separate PR.

shengxiang19:
> > Will implement a test/discrete/test_c51.py in the next PR.
> 
> I think you can directly add this file. No need to make a separate PR.

I hope to combine the results of C51 in a variety of atrai games with it in the future PR.

Trinkle23897:
> I hope to combine the results of C51 in a variety of Atari games with it in the future PR.

Cool, and I think you can add these results here so that this PR can be a complete version of C51 implementation :)

shengxiang19:
> > I hope to combine the results of C51 in a variety of Atari games with it in the future PR.
> 
> Cool, and I think you can add these results here so that this PR can be a complete version of C51 implementation :)

I'm not sure how to add new files under the current PR. So, add a new PR is more convenient for me. In addition, I'm not quite sure when I can finish this work.

Trinkle23897:
> I'm not sure how to add new files under the current PR. So, add a new PR is more convenient for me.

Just add the file in shengxiang19/C51 branch instead of here. See https://stackoverflow.com/questions/10147445/github-adding-commits-to-existing-pull-request

> In addition, I'm not quite sure when I can finish this work.

I can wait for you.

shengxiang19:
> > I'm not sure how to add new files under the current PR. So, add a new PR is more convenient for me.
> 
> Just add the file in shengxiang19/C51 branch instead of here. See https://stackoverflow.com/questions/10147445/github-adding-commits-to-existing-pull-request
> 
> > In addition, I'm not quite sure when I can finish this work.
> 
> I can wait for you.

Thank you. I can try it later.

shengxiang19:
> > Will implement a test/discrete/test_c51.py in the next PR.
> 
> I think you can directly add this file. No need to make a separate PR.

I have add a test_c51 for CartPole-v0 under test/discrete/. Hope you can help me check it.

shengxiang19:
I have add the results of C51 in three typical atari environments. My current plan of C51 is done.

tianshou/utils/net/common.py
```diff
@@ -159,3 +164,55 @@ def forward(
...
+        logits = self.model(s)
+        if self.dueling is not None:  # Dueling DQN
+            q, v = self.Q(logits), self.V(logits)
+            v = v.view(-1, 1, self.num_atoms),
+            q = q.view(-1, np.prod(self.action_shape), self.num_atoms)
+            logits = q - q.mean(dim=1, keepdim=True) + v
+        else:
+            logits = logits.view(
+                -1, np.prod(self.action_shape), self.num_atoms)
+        logits = torch.softmax(logits, dim=-1)
```
Trinkle23897: Can these lines be simplified? 
```python
logits, state = super().forward(s, state, info)
logits = logits.view(-1, np.prod(self.action_shape), self.num_atoms)
if self.softmax:
    logits = torch.softmax(logits, dim=-1)
return logits, state
```

Trinkle23897 (commented):
I'll optimize the n_step code this week (in this PR). Thanks for your great work!

tianshou/utils/net/common.py
shengxiang19: I don't think so. The mean of q in CategoricalNet is of size (batch_size, num_atoms). However, the mean of q in net is of size (batch_size) .

Trinkle23897: Got it

shengxiang19: Maybe we can change the way of calculating the mean of q in net. But it may need extra judgment that whether num_atoms is 1.

tianshou/policy/modelfree/c51.py
Trinkle23897: > I think it is inefficient to convert this array many times.

I think it is negligible.

```python
import time
import numpy as np
from numba import njit

np.random.seed(0)

nstep = 3
rew = np.random.rand(nstep)
target = np.random.rand()
# assume done == False
gamma = .99
bsz = 32
n_atoms = 51

@njit
def compute_target(gamma, nstep, rew, target):
    for n in range(nstep - 1, -1, -1):
        target = target * gamma + rew[n]
    return target

def timeit(n, func):
    t0 = time.time()
    for _ in range(n):
        func()
    return (time.time() - t0) / n

# single_target
single_target = compute_target(gamma, nstep, rew, target)

# batch_target
batch_target = compute_target(gamma, nstep, rew, np.array([target] * bsz))

# c51_target
c51_target = compute_target(gamma, nstep, rew, np.full([bsz, n_atoms], target))

assert single_target == batch_target[0] == c51_target[0, 0]

# timeit
n = 100000
single_target_time = timeit(n, lambda: compute_target(gamma, nstep, rew, target))
batch_target_time = timeit(n, lambda: compute_target(gamma, nstep, rew, np.full([bsz, 1], target)))
c51_target_time = timeit(n, lambda: compute_target(gamma, nstep, rew, np.full([bsz, n_atoms], target)))
ref_target_time = timeit(n, lambda: compute_target(gamma, nstep, rew, np.full([bsz, 1], target)) + c51_target)
print(single_target_time)
print(batch_target_time)
print(ref_target_time)
print(c51_target_time)
print(c51_target_time / ref_target_time)
```

The result of `c51_target_time / ref_target_time` is about 1 (without njit) and <1 (with njit). 
I'll modify the current nstep implementation to extend any shape of the given `target_q`.

shengxiang19: After you modify the current nstep implementation, I can modify the current implementation of C51.

tianshou/policy/modelfree/c51.py
```diff
@@ -0,0 +1,142 @@
...
+        self.optim.zero_grad()
+        with torch.no_grad():
+            target_dist = self._target_dist(batch)
+        weight = batch.pop("weight", 1.0)
+        curr_dist = self(batch).logits
+        act = batch.act
+        curr_dist = curr_dist[np.arange(len(act)), act, :]
+        cross_entropy = - (target_dist * torch.log(curr_dist + 1e-8)).sum(1)
+        loss = (cross_entropy * weight).mean()
+        batch.weight = cross_entropy.detach()  # prio-buffer
```
Trinkle23897, line 139: Could you please explain more on the `weight`?

Trinkle23897 (commented):
I'm running QbertNoFrameskip-v4 for evaluation. Now in epoch 31 it reaches 14047.
Please double-check my implementation.

tianshou/policy/modelfree/c51.py
shengxiang19: The choose of weight follows the setting of line 94-100 in https://github.com/Kaixhin/Rainbow/blob/master/agent.py

shengxiang19: The performance of QbertNoFrameskip is consistent with my previous experiment. I will check the code as soon as possible.

shengxiang19: I just checked the code. Everything is ok. Thank you for making C51 more concise.

Trinkle23897 (approved):

tianshou/policy/modelfree/c51.py
```diff
@@ -0,0 +1,143 @@
...
+            * ``state`` the hidden state.
+
+        .. seealso::
+
+            Please refer to :meth:`~tianshou.policy.DQNPolicy.forward` for
+            more detailed explanation.
+        """
+        model = getattr(self, model)
+        obs = batch[input]
+        obs_ = obs.obs if hasattr(obs, "obs") else obs
```
duburcqa, line 88: Why `hasattr(obs, "obs")` could be false ?

tianshou/policy/modelfree/c51.py
```diff
@@ -0,0 +1,143 @@
...
+        dist, h = model(obs_, state=state, info=batch.info)
+        q = (dist * self.support).sum(2)
+        act: np.ndarray = to_numpy(q.max(dim=1)[1])
+        if hasattr(obs, "mask"):
```
duburcqa, line 92: I don't like much this approach, but right now I have no idea about to avoid it. Maybe adding `masked_array` method to `Batch` class to offer something similar to [numpy's masked arrays](https://numpy.org/doc/stable/reference/maskedarray.generic.html). Internally it would use the same mechanism, but it would be hidden in `Batch`, which is way better in by opinion.

tianshou/policy/modelfree/c51.py
```diff
@@ -0,0 +1,143 @@
...
+        curr_dist = self(batch).logits
+        act = batch.act
+        curr_dist = curr_dist[np.arange(len(act)), act, :]
+        cross_entropy = - (target_dist * torch.log(curr_dist + 1e-8)).sum(1)
+        loss = (cross_entropy * weight).mean()
+        # ref: https://github.com/Kaixhin/Rainbow/blob/master/agent.py L94-100
+        batch.weight = cross_entropy.detach()  # prio-buffer
+        loss.backward()
+        self.optim.step()
+        self._cnt += 1
```
duburcqa, line 142: I recommend explicit variable names `_cnt`

tianshou/policy/modelfree/c51.py
Trinkle23897: These three are the same as existing DQNPolicy. I guess we can make a separate PR to enhance these things :)

duburcqa: Yes I noticed that :)

PR #275, opened 2021-01-18
update utils.network

### LINKED ISSUE #274: Plans of releasing mujoco benchmark with ddpg/sac/td3 on Tianshou
## Purpose
The purpose of this issue(discussion) is to introduce a series of pr in the near future targeted to releasing a benchmark(sac, td3, ddpg) on mujoco environments. Some features of tianshou platform will be enhanced along the way.
## Introduction
By the time this issue is proposed, tianshou platform has attracted 2.4k star on github, and has become a very popular deep rl library based purely on Pytorch(in contrast with [openai baseline](https://github.com/openai/baselines), [rl lab](https://github.com/rll/rllab), etc), thanks to the contributions of @Trinkle23897  @duburcqa, @youkaichao, etc. However, with the users growing day by day, some problems start to spring up. One critical problem is that although Tianshou is a fast speed, structured, flexible library and supports many classic algorithms officially, it has done a relatively poor job on benchmarking the algorithm it supports. Examples and demonstrations are mostly tested on toy environments of gym, and we have not yet provided detailed comparison and analysis with classic papers on officially supported algorithms, which might make users worry about the correctness and efficiency of algorithms, make it a bit hard for researchers using Tianshou to reproduce results of classic papers because of the lack of trustworthy hyperparameters(baseline in other words).

Tianshou hopes to provide users with a lightweight and efficient drl platform and reduce the burden of rl researchers as much as possible. Even if users are only starters and might not be so familiar with drl algorithms or baselines, they can design their own algorithm with minimal lines of code by inheriting and using official data/algorithm structures, understand source code and compare their idea with standard algorithms easily. In order to achieve this, one thing we have to do is to provide a detailed benchmark for widely used algorithms and environments.

This is what I have been trying to do, and the first step has been taken. Using tianshou, I have managed to create a state-of-the-art benchmark on three algorithms on mujoco's mostly widely used 9/14 environments.

### ddpg
|Environment| Tianshou| [spining up(Pytorch)](https://spinningup.openai.com/en/latest/spinningup/bench.html)| [TD3 paper(ddpg)](https://arxiv.org/abs/1802.09477)|[TD3 paper(our ddpg)](https://arxiv.org/abs/1802.09477)|
| --------------- | --------------- | --------------- | --------------- | --------------- |
|Ant|990.4±4.3 | ~840 | **1005.3** | 888.8 |
|HalfCheetah|**11718.7±465.6** | ~11000 | 3305.6 | 8577.3 |
|Hopper|**2197.0±971.6** | ~1800 | **2020.5** | 1860.0 |
|Walker2d|1400.6±905.0 | ~1950 | 1843.6 | **3098.1** |
|Swimmer|**144.1±6.5** | ~137 | N | N |
|Humanoid|**177.3±77.6** | N | N | N |
|Reacher|**-3.3±0.3** | N | -6.51 | -4.01 |
|InvertedPendulum|**1000.0±0.0** | N | **1000.0** | **1000.0** |
|InvertedDoublePendulum|8364.3±2778.9 | N | **9355.5** | 8370.0 |

### td3
|Environment| Tianshou| [spining up(Pytorch)](https://spinningup.openai.com/en/latest/spinningup/bench.html)| [TD3 paper](https://arxiv.org/abs/1802.09477)|
| --------------- | --------------- | --------------- | --------------- |
|Ant|**5116.4±799.9** | ~3800 | 4372.4±1000.3 |
|HalfCheetah|**10201.2±772.8** | ~9750 | 9637.0±859.1 |
|Hopper|3472.2±116.8 | ~2860 | **3564.1±114.7** |
|Walker2d|3982.4±274.5 | ~4000 | **4682.8±539.6** |
|Swimmer|**104.2±34.2** | ~78 | N |
|Humanoid|**5189.5±178.5** | N | N |
|Reacher|**-2.7±0.2** | N | -3.6±0.6 |
|InvertedPendulum|**1000.0±0.0** | N | **1000.0±0.0** |
|InvertedDoublePendulum|**9349.2±14.3** | N | **9337.5±15.0** |

### sac
|Environment| Tianshou| [spining up(Pytorch)](https://spinningup.openai.com/en/latest/spinningup/bench.html)| [SAC paper](https://arxiv.org/abs/1801.01290)|
| --------------- | --------------- | --------------- | --------------- |
|Ant|**5850.2±475.7** | ~3980 | ~3720 |
|HalfCheetah|**12138.8±1049.3** | ~11520 | ~10400 |
|Hopper|**3542.2±51.5** | ~3150 | ~3370 |
|Walker2d|**5007.0±251.5** | ~4250 | ~3740 |
|Swimmer|**44.4±0.5** | ~41.7 | N |
|Humanoid|**5488.5±81.2** | N | ~5200 |
|Reacher|**-2.6±0.2** | N | N |
|InvertedPendulum|**1000.0±0.0** | N | N |
|InvertedDoublePendulum|**9359.5±0.4** | N | N |

\*\*\*\* We didn't compare to OPENAI baselines, because for now I think its [benchmark](https://github.com/openai/baselines/blob/master/benchmarks_mujoco1M.htm) is corrupted(?), and I haven't been able to find the information I need. But in [spining up docs](https://spinningup.openai.com/en/latest/spinningup/bench.html) they stated that "Spinning Up implementations of DDPG, TD3, and SAC are roughly at-parity with the best-reported results for these algorithms", so I guess lack of comparisons with OPENAI baselines is okay.

figure

**I only show one figure here as an example, all other figures for tianshou mujoco benchmark can be found** [here](https://cloud.tsinghua.edu.cn/d/356e0f5d1e66426b9828/).

To achieve the results is not easy, because it requires not only hyperparameter tuning, but some features of Tianshou platform have to be changed first, most of which are already mentioned in different issues by different users. For example:

- #140 #245 #255 metioned that collector collect whole episodes of data when setting n_step = 1.
- #249 mentioned that speed comparison might be unfair because originally Tianshou uses update step as absciss.
- #209 mentioned that original mujoco results can no longer be reproduced because we have changed the code a lot in the past few months.
- discussion on #194 indicates that some policies officially supported by Tianshou can be refactored to be easier to use.
- #161 requires curve-drawing examples or tools, which is also urgently needed when creating a benchmark.

There are also other problems issues haven't mentioned or I haven't noticed. For instance:

- In trainer, log_interval for update step and env step can only be the same, which will cause inconvenience. A flexible logger might help.
- In net utils, Net function can only create MLP in which all hidden layer numbers is the same.
- In policies, some policies will add explore noise when evaluating the algorithm.
- Buffer and Collector now in Tianshou is a little bit too complex because they try to support all features in one single class, which will cause great inconvenience when trying to understand source code inheriting from those class to create customized data structure.

All the problems above will be taken care of to a certain extent when trying to release the benchmark. Scripts that achieve this benchmark is hosted on my fork of Tianshou, and can be found [here](https://github.com/ChenDRAG/tianshou/tree/step). However, it cannot be directly merged, because it is only what we use to demonstrate our idea, so it's not well organized (Lack of consistency, docs, comments, tests, etc.). Another reason is that this will be a big merge on Tianshou and we want to try hard to enhance Tianshou without causing too much interference for our users. As a result, I make a plan and hope to merge all the codes in 6 commits in total in the next few weeks. All of these commits are targeted to releasing the benchmark above eventually.

## Plans
Here I briefly introduce what these 6 commits try to do.

1.  In net utils, enhance `Net ` function, and make it support any type of MLP.

- This is the most urgent commit because Net function will be needed in another pr.

2. Minor fix of batch and adding a new ReplayBuffer class called CachedReplayBuffer.

- CachedReplayBuffer is used to replace `_cached_buf ` of Collector in the next commit, which is critical to solving the n_step problem mentioned in #245.
- change the definition of ReplayBuffer to certain management of Batch, because chronologically organized ReplayBuffer might not be suitable for all scenarios.
- give all buffer types inheriting from ReplayBuffer the same API (indexing method for instance), let the developers worry about the underlying implementation of different types of ReplayBuffer, not the users.
- [Probably] Separation of stack option and other abilities of ReplayBuffer, to make the source code easier to understand or rewrite. Gain efficiency at the same time.
- docs, tests, etc.

3. Refactor of collector to support both ReplayBuffer and CachedReplayBuffer.

- fix #245 by supporting CachedReplayBuffer and not allowing ReplayBuffer to work when n_env > 1.
- removed those not widely used return info, make code more lightweight.
- change BasePolicy to be prepared for the incoming change of indexing method of CachedReplayBuffer.
- fix a bug in BasePolicy: when ignoring done and setting n_step > 1 in offpolicy algorithms, a small amount of targer q will have calculation error.
- change the behavior of action noise, expl noise will all be added in collector from now on, making it easier to redefine, less possible to cause bug when added in forward function. Partly sovle #194.
- little change in trainer, to coordinate collector's change.
- docs, tests, etc.

4. Refactor of trainer to add a self-defined logger in trainer.

- add a logger in trainer which can be self-defined and will be used in benchmarking.
- remove original log_interval, save_fn, writer, etc. (All logging function).
- add a default logger which basically do all jobs of original logging function. Partly solve #161. 
- docs, tests, etc.

5. Some small fixes in tianshou/policy to make policies easier to use and add some standard tricks to it.

- take consideration of gym's 'TimeLimit.truncated' flag, to make the policy more efficient.

6. Releasing mujoco benchmark(source code, data, graphs, detailed comparison, analysis of hyperparameters, etc) on 3 algorithms.

## Future work
1. Remove warnings and implementations for originally supported but now unsupported methods.
2. Adding support(benchmark in the same way) for other algorithms(VPG, PPO, TRPO, etc).
2. Speed analysis, and provide a set of hyperparameters that can be trained in Parallel using tianshou to speed up training.
3. Consider discrete-action environments like Atari (Maybe support rainbow on Tianshou).
4. A tutorial on how to tune hyperparameters of certain rl problem.
5. ......

ChenDRAG:
This is the first commit of 6 commits mentioned in #274, which features

1. Refactor of `Class Net` to support any form of MLP.
2. Enable type check in utils.network.
3. Relative change in docs/test/examples.
4. Move atari-related network to examples/atari/atari_network.py


test/base/test_utils.py
```diff
@@ -40,9 +40,13 @@ def test_net():
     action_shape = (5, )
     data = torch.rand([bsz, *state_shape])
     expect_output_shape = [bsz, *action_shape]
-    net = Net(3, state_shape, action_shape, norm_layer=torch.nn.LayerNorm)
+    net = Net([128, 128, 128, 128], state_shape,
+              action_shape, norm_layer=torch.nn.LayerNorm)
     assert list(net(data)[0].shape) == expect_output_shape
-    net = Net(3, state_shape, action_shape, dueling=(2, 2))
+    dueling = (({"hidden_layer_size": [128, 128]},
```
duburcqa: Double parenthesis ''((" ?

tianshou/utils/net/common.py
```diff
@@ -9,46 +9,134 @@
...
+        Default to empty list (no normalization).
+    :param activation: which activation to use after each layer, can be both
+        the same actvition for all layers if passed in nn.Module, or different
+        activation for different Modules if passed in a list.
+        Default to nn.ReLU.
+    """
+
+    def __init__(
+        self,
+        hidden_layer_size: Union[int, List[int]] = [],
```
duburcqa: 1. Never ever use mutable types as default value. 
2. Why not (128, 128) instead ? I still more usual as default

tianshou/utils/net/common.py
```diff
@@ -9,46 +9,134 @@
...
+    ret += [activation(inplace=True)]
     return ret
 
 
-class Net(nn.Module):
+class MLP(nn.Module):
     """Simple MLP backbone.
+    Create a MLP of size inp_shape*hidden_layer_size[0]*hidden_layer_size[1]...
+
+    :param hidden_layer_size: shape of MLP passed in as a list, not incluing
```
duburcqa: 'hidden_layers_sizes' instead 'hidden_layer_size'

tianshou/utils/net/common.py
```diff
@@ -9,46 +9,134 @@
...
+        hidden_layer_size: Union[int, List[int]] = [],
+        norm_layer: Optional[Union[nn.modules.Module,
+                                   List[nn.modules.Module]]] = [],
+        activation: Optional[Union[nn.modules.Module,
+                                   List[nn.modules.Module]]] = nn.ReLU,
+        inp_shape: Optional[Union[tuple, int]] = None,
+        device: Union[str, int, torch.device] = "cpu",
+    ) -> None:
+        super().__init__()
+        if isinstance(hidden_layer_size, int):
```
duburcqa: I would rather not allow such usecase, it is ill-defined and misleading.

tianshou/utils/net/common.py
```diff
@@ -9,46 +9,134 @@
...
+            hidden_layer_size = [hidden_layer_size]
+        layer_size = hidden_layer_size.copy()
+        if inp_shape is not None:
+            layer_size.insert(0, np.prod(inp_shape))
```
duburcqa: why not just `layer_size = [np.prod(inp_shape)] + layer_size` ? It is easier to read and it is probably doing the same internally.

tianshou/utils/net/common.py
```diff
@@ -9,46 +9,134 @@
...
+                                   List[nn.modules.Module]]] = [],
+        activation: Optional[Union[nn.modules.Module,
+                                   List[nn.modules.Module]]] = nn.ReLU,
+        inp_shape: Optional[Union[tuple, int]] = None,
+        device: Union[str, int, torch.device] = "cpu",
+    ) -> None:
+        super().__init__()
+        if isinstance(hidden_layer_size, int):
+            hidden_layer_size = [hidden_layer_size]
+        layer_size = hidden_layer_size.copy()
```
duburcqa: I would rather cast to `list` instead doing `list(hidden_layer_size)` ? It would perform the copy automatically, but also enable to support any `Typing.Sequence` type, which is desirable since at least `Tuple` are quite usual.

examples/box2d/acrobot_dualdqn.py
```diff
@@ -56,8 +57,10 @@ def test_dqn(args=get_args()):
     train_envs.seed(args.seed)
     test_envs.seed(args.seed)
     # model
-    net = Net(args.layer_num, args.state_shape,
-              args.action_shape, args.device, dueling=(2, 2)).to(args.device)
+    dueling = ({"hidden_layer_size": [128, 128]},
+               {"hidden_layer_size": [128, 128]})
```
Trinkle23897: Tuple[List[int], List[int]] or Dict[str, List[int]] instead of Tuple[Dict[str, List[int]]]

e.g., 
```
dueling = ([128, 128], [128, 128])
# or
dueling = {"hidden_layer_size_q": [128, 128], "hidden_layer_size_v": [128, 128]}
```

tianshou/utils/net/common.py
```diff
@@ -9,46 +9,134 @@
...
+        self.inp_dim = layer_size[0]
+        self.out_dim = layer_size[-1]
+        self.device = device
+        if norm_layer:
+            if isinstance(norm_layer, list):
+                assert len(norm_layer) == len(layer_size) - 1, (
+                    "length of norm_layer should match the "
+                    "length of hidden_layer_size.")
+            else:
+                norm_layer = [norm_layer]*(len(layer_size) - 1)
```
duburcqa: `[norm_layer for _ in range(len(layer_size) - 1)]` to avoid any side effects.

tianshou/utils/net/common.py
```diff
@@ -9,46 +9,134 @@
...
+        if activation:
+            if isinstance(activation, list):
+                assert len(activation) == len(layer_size) - 1, (
+                    "length of activation should match the "
+                    "length of hidden_layer_size.")
+            else:
+                activation = [activation]*(len(layer_size) - 1)
```
duburcqa: same.

tianshou/utils/net/common.py
```diff
@@ -9,46 +9,134 @@
...
+                kwargs["norm_layer"] = norm_layer[i]
+            if activation:
+                kwargs["activation"] = activation[i]
+            model += miniblock(
+                layer_size[i], layer_size[i+1], **kwargs)
+        self.model = nn.Sequential(*model)
+
+    def forward(self, inp):
+        inp = to_torch(inp, device=self.device, dtype=torch.float32)
+        inp = inp.reshape(inp.size(0), -1)
```
duburcqa: `inp.shape[0]` is more usual

tianshou/utils/net/common.py
```diff
@@ -9,46 +9,134 @@
...
+        for i in range(len(layer_size) - 1):
+            if norm_layer:
+                kwargs["norm_layer"] = norm_layer[i]
+            if activation:
+                kwargs["activation"] = activation[i]
+            model += miniblock(
+                layer_size[i], layer_size[i+1], **kwargs)
+        self.model = nn.Sequential(*model)
+
+    def forward(self, inp):
```
duburcqa: `inp` is not a very good name.

test/discrete/test_a2c_with_il.py
```diff
@@ -107,7 +108,7 @@ def stop_fn(mean_rewards):
     # here we define an imitation collector with a trivial policy
     if args.task == 'CartPole-v0':
         env.spec.reward_threshold = 190  # lower the goal
-    net = Net(1, args.state_shape, device=args.device)
+    net = Net([128], args.state_shape, device=args.device)
```
Trinkle23897: change to args.imitation_hidden_layer_size

tianshou/utils/net/common.py
ChenDRAG: because [] means, this Net do nothing, it outputs the input

tianshou/utils/net/common.py
ChenDRAG: I still suggest using hidden_layers_size for consistency with history.

tianshou/utils/net/common.py
ChenDRAG: No, hidden_layer_size as int is not allowed now, I only check this because originally users pass in layer_num(int) as first argument of Net. This is a reminder to them that original class has been changed when they see the error. Again, for consistency.

tianshou/utils/net/common.py
ChenDRAG: I apporve.

test/base/test_utils.py
ChenDRAG: othwise there is an error because line too long?

tianshou/utils/net/common.py
duburcqa: > because [] means, this Net do nothing, it outputs the input

Yes, and why would it be a good choice of default parameters ? I don't see why it would be better than no default.

tianshou/utils/net/common.py
Trinkle23897: > I still suggest using hidden_layers_size for consistency with history.

I don't think so. We should break consistency at this point, otherwise it may be confusing (since we change API, an explicit warning should be generated when users use the network with old fashion)

How about `hidden_size` or some other names?

duburcqa: > I don't think so. We should break consistency at this point, otherwise it may be confusing (since we change API, an explicit warning should be generated when users use the network with old fashion)

exactly. `hidden_sizes` since it is a list, not a scalar.

tianshou/utils/net/common.py
Trinkle23897: I add both `input_dim` and `output_dim` in MLP. Now if `hidden_sizes == []`, the MLP equals to a single `nn.Linear(input_dim, output_dim)`.

tianshou/utils/net/common.py
Trinkle23897: resolved as replacing hidden_layer_size with hidden_sizes

Trinkle23897:
Broken support yet, I'll fix them tmr

test/base/test_utils.py
duburcqa: You should not need it then.

tianshou/utils/net/common.py
duburcqa: Fine. Yet, using `[]` as default parameter is forbidden. See [this article](https://florimond.dev/blog/articles/2018/08/python-mutable-defaults-are-the-source-of-all-evil/) about it for instance.

So everywhere where `[]` is used as default should be replaced by either `None` or `()`, and typing accordingly (respectively `Optional[List[T]]` or `Sequence[T]`).

Trinkle23897:
ready now

examples/box2d/lunarlander_dqn.py
```diff
@@ -29,7 +29,8 @@ def get_args():
     parser.add_argument('--step-per-epoch', type=int, default=5000)
     parser.add_argument('--collect-per-step', type=int, default=16)
     parser.add_argument('--batch-size', type=int, default=128)
-    parser.add_argument('--layer-num', type=int, default=1)
+    parser.add_argument('--hidden-sizes', type=int,
+                        nargs='*', default=[128, 128])
```
duburcqa, line 33: same.

examples/box2d/lunarlander_dqn.py
```diff
@@ -57,9 +58,10 @@ def test_dqn(args=get_args()):
     train_envs.seed(args.seed)
     test_envs.seed(args.seed)
     # model
-    net = Net(args.layer_num, args.state_shape,
-              args.action_shape, args.device,
-              dueling=(2, 2)).to(args.device)
+    Q_param = V_param = {"hidden_sizes": [128, 128]}
```
duburcqa: same.

examples/box2d/mcc_sac.py
```diff
@@ -32,7 +32,8 @@ def get_args():
     parser.add_argument('--step-per-epoch', type=int, default=2400)
     parser.add_argument('--collect-per-step', type=int, default=5)
     parser.add_argument('--batch-size', type=int, default=128)
-    parser.add_argument('--layer-num', type=int, default=1)
+    parser.add_argument('--hidden-sizes', type=int,
+                        nargs='*', default=[128, 128])
```
duburcqa, line 36: same.

examples/mujoco/mujoco_sac.py
```diff
@@ -33,8 +33,8 @@ def get_args():
     parser.add_argument('--update-per-step', type=int, default=1)
     parser.add_argument('--pre-collect-step', type=int, default=10000)
     parser.add_argument('--batch-size', type=int, default=256)
-    parser.add_argument('--hidden-layer-size', type=int, default=256)
-    parser.add_argument('--layer-num', type=int, default=1)
+    parser.add_argument('--hidden-sizes', type=int,
+                        nargs='*', default=[128, 128])
```
duburcqa, line 37: same.

tianshou/utils/net/common.py
```diff
@@ -1,112 +1,198 @@
...
+                    norm_layer for i in range(len(hidden_sizes))]
+        else:
+            norm_layer_list = [None] * len(hidden_sizes)
+        if activation:
+            if isinstance(activation, list):
+                assert len(activation) == len(hidden_sizes)
+                activation_list = activation
+            else:
+                activation_list = [
+                    activation for i in range(len(hidden_sizes))]
```
duburcqa: `for _ in `

tianshou/utils/net/common.py
```diff
@@ -1,112 +1,198 @@
...
+    ) -> None:
+        super().__init__()
+        self.device = device
+        if norm_layer:
+            if isinstance(norm_layer, list):
+                assert len(norm_layer) == len(hidden_sizes)
+                norm_layer_list = norm_layer
+            else:
+                norm_layer_list = [
+                    norm_layer for i in range(len(hidden_sizes))]
```
duburcqa: `for _ in `

examples/atari/atari_network.py
```diff
@@ -0,0 +1,82 @@
...
+                self.net,
+                nn.Linear(self.output_dim, 512), nn.ReLU(inplace=True),
+                nn.Linear(512, np.prod(action_shape)))
+            self.output_dim = np.prod(action_shape)
+
+    def forward(
+        self,
+        x: Union[np.ndarray, torch.Tensor],
+        state: Optional[Any] = None,
+        info: Dict[str, Any] = {},
```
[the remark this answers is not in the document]
Trinkle23897: However this is gym's default return value. The default value of info is `{}`.

examples/box2d/bipedal_hardcore_sac.py
```diff
@@ -30,7 +30,8 @@ def get_args():
     parser.add_argument('--step-per-epoch', type=int, default=10000)
     parser.add_argument('--collect-per-step', type=int, default=10)
     parser.add_argument('--batch-size', type=int, default=128)
-    parser.add_argument('--layer-num', type=int, default=1)
+    parser.add_argument('--hidden-sizes', type=int,
+                        nargs='*', default=[128, 128])
```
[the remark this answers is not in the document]
Trinkle23897: But this is the standard method for receiving a list from command line, see 
https://stackoverflow.com/a/32763023/8773993
https://stackoverflow.com/questions/33564246/passing-a-tuple-as-command-line-argument/33564487

examples/atari/atari_network.py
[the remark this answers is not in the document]
duburcqa: Ok then. Forgot about that.

examples/box2d/bipedal_hardcore_sac.py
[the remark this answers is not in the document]
duburcqa: ok ! i'm not familiar with argparse.

Trinkle23897 (approved):

PR #278, opened 2021-01-20
Add CachedReplayBuffer and ReplayBufferManager

### LINKED ISSUE #274: Plans of releasing mujoco benchmark with ddpg/sac/td3 on Tianshou

ChenDRAG:
This is the second commit of 6 commits mentioned in #274, which features minor refactor of ReplayBuffer and adding two new ReplayBuffer classes called CachedReplayBuffer and ReplayBufferManager. You can check #274 for more detail.

1. Add ReplayBufferManager (handle a list of buffers) and CachedReplayBuffer;
2. Make sure the reserved keys cannot be edited by methods like `buffer.done = xxx`;
3. Add `set_batch` method for manually choosing the batch the ReplayBuffer wants to handle;
4. Add `sample_index` method, same as `sample` but only return index instead of both index and batch data;
5. Add `prev` (one-step previous transition index), `next` (one-step next transition index) and `unfinished_index` (the last modified index whose done==False);
6. Separate `alloc_fn` method for allocating new memory for `self._meta` when a new `(key, value)` pair comes in;
7. Move buffer's documentation to `docs/tutorials/concepts.rst`.


ChenDRAG:
@Trinkle23897 Could you please help me check if it is possible to separate of stack option and other abilities of ReplayBuffer? I have not done that part. And it is a little diffcult for me because I'm not familar with RNN stack usage. Thanks a lot.

Trinkle23897:
**Update: resolved**

@duburcqa I have some trouble with the new buffer type's pickle. (I think the current code is clean but lack of unit test, will be fixed soon...)

Because the CachedReplayBuffer needs to perform `getitem` from many sub-buffers, if each sub-buffer's `_meta` is a separated batch (no memory contiguous), it will first `getitem` then `Batch.cat` all of the data, which may cause a lot of overhead. To avoid this, our approach is to create `_meta` on the top, and create batch slice and send to the sub-buffers by `set_batch`.

Here is an example: we now have a buffer that contains three sub-buffers, with each of size=10 (total=30). At this time, we add a new key-value pair called `info["Timelimit.truncate"]: bool`. It will allocate new memory as follows:
```
Step 1: one sub-buffer's `_add_to_buffer` will raise exception, and will call a hook function `_alloc` to
        require new memory for this key;
Step 2: The top of this buffer class receives this signal, and generate `np.zeros(30, np.bool_)` to 
        `top_buffer._meta.info["Timelimit.truncate"]`;
Step 3: Dispatch the whole `top_buffer._meta` to sub-buffers, i.e., 
        `subbuffer[0].set_batch(top_buffer._meta[0:10])`
```

The current implementation use `alloc_fn` hook as an input argument, but in `VectorReplayBuffer` and `CachedReplayBuffer`, this approach is not compatible with pickle, because the user-defined hook cannot be pickled. You can run the script under `test/base/test_buffer.py` to see this issue.

I haven't figure out the solution to this issue, could you please give us some suggestions?

duburcqa:
I will do my best to have a look today, but I'm a little bit sick since a few days so maybe tomorrow if it is ok for you

Trinkle23897:
Sorry to hear about that... hope you get well soon.

tianshou/data/buffer.py
```diff
@@ -520,13 +437,215 @@ def update_weight(
...
+    :param int buffer_list: a list of ReplayBuffers needed to be handled.
+
+    .. seealso::
+
+        Please refer to :class:`~tianshou.data.ReplayBuffer` for more detailed
+        explanation.
+    """
+
+    def __init__(self, buffer_list: List[ReplayBuffer], **kwargs: Any) -> None:
+        self.buffer_num = len(buffer_list)
```
ChenDRAG, line 474: I think we need to check if buffer_list meta is legal(empty), otherwise it will cause error

tianshou/data/buffer.py
```diff
@@ -520,13 +437,215 @@ def update_weight(
...
-            info=self.get(index, "info"),
-            policy=self.get(index, "policy"),
-            weight=self.weight[index],
-        )
+        batch = super().__getitem__(index)
+        batch.weight = self.weight[index]
+        return batch
+
+
+class ReplayBuffers(ReplayBuffer):
```
ChenDRAG: I suggest we do not make offset a value inside each buffer(because it belongs to replaybuffers but not that buffer). we can use a list or dict instead?

tianshou/data/buffer.py
```diff
@@ -520,13 +437,215 @@ def update_weight(
...
+        size: int,
+        cached_buffer_num: int,
+        max_episode_length: int,
+        **kwargs: Any,
+    ) -> None:
+        assert cached_buffer_num > 0 and max_episode_length > 0
+        main_buffer = ReplayBuffer(size, **kwargs)
+        buffers = [main_buffer] + [ReplayBuffer(max_episode_length, **kwargs)
+                                   for _ in range(cached_buffer_num)]
+        super().__init__(buffer_list=buffers, **kwargs)
```
ChenDRAG, line 632: suggestions:
self.main_buffer = main_buffer  (to replace buffers[0]) in line 648,etc.
self.cached_buffers = buffers[1:]

we use 'buffers ' to  initialise super(). but keep api clear to user

tianshou/data/buffer.py
```diff
@@ -520,13 +437,215 @@ def update_weight(
...
+
+        Return the array of episode_length and episode_reward with shape
+        (len(cached_buffer_ids), ...), where (episode_length[i],
+        episode_reward[i]) refers to the cached_buffer_ids[i]th cached buffer's
+        corresponding episode result.
+        """
+        if cached_buffer_ids is None:
+            cached_buffer_ids = np.arange(self.buffer_num - 1)
+        # in self.buffers, the first buffer is main_buffer
+        buffer_ids = np.asarray(cached_buffer_ids) + 1
```
ChenDRAG: suggest using api we just talk about to replace '+1'

tianshou/data/buffer.py
```diff
@@ -241,117 +174,109 @@ def add(
...
+        else:  # TODO: remove this after deleting ListReplayBuffer
+            self._size = self._index = self._size + 1
+
+        if done:
+            result = (self._episode_length, self._episode_reward)
+            self._episode_length, self._episode_reward = 0, 0.0
+            return result
         else:
-            self._size = self._index = self._index + 1
+            return (0, 0.0)
```
ChenDRAG: can 'add' in Replaybuffer return np.array just to be consistent with cachedbuffers?

tianshou/data/buffer.py
```diff
@@ -520,13 +437,215 @@ def update_weight(
...
+            buf.offset = offset
+            offset += buf.maxsize
+        super().__init__(size=offset, **kwargs)
+
+    def __len__(self) -> int:
+        return sum([len(buf) for buf in self.buffers])
+
+    def reset(self) -> None:
+        for buf in self.buffers:
+            buf.reset()
```
ChenDRAG, line 492: should we delete self._index, _size, _episode_reward, _episode_length, _save_only_last_obs, _save_obs_next, _indices since they are never used?

tianshou/data/buffer.py
ChenDRAG: Also, we should make sure data type is the most basic ReplayBuffer

tianshou/data/buffer.py
Trinkle23897: No because self.cached_buffers is a list instead of `ReplayBuffers`

tianshou/data/buffer.py
Trinkle23897: What's the dimension you want, 0-dim or 1-dim?

tianshou/data/buffer.py
Trinkle23897: No, self._save_obs_next and self._indices is needed because ReplayBuffers inherits the bsae class's `__getitem__`. If it is removed, the cached replay buffer cannot work as we expected. I add it in the unit test and you can have a try.

tianshou/data/buffer.py
ChenDRAG: Both seems okay, 0-dim is clearer but 1-dim is more consistent

tianshou/data/buffer.py
```diff
@@ -464,69 +397,295 @@ def add(
...
+    def __init__(
+        self,
+        main_buffer: ReplayBuffer,
+        cached_buffer_num: int,
+        max_episode_length: int,
+    ) -> None:
+        assert cached_buffer_num > 0 and max_episode_length > 0
+        self.main_buffer = main_buffer
+        self._is_prioritized = isinstance(main_buffer, PrioritizedReplayBuffer)
+        kwargs = self.main_buffer.options
```
ChenDRAG: I think self.cached_buffer_num = cached_buffer_num is needed because len(self.buffers) - 1 seems not good and unclear.
Also, suggest add self.cached_buffers = self.buffers[1:] (will be needed in later commit)

tianshou/data/buffer.py
ChenDRAG: what about self._index, _size,_episode_reward, _episode_length, then

tianshou/data/buffer.py
```diff
@@ -464,69 +397,295 @@ def add(
...
+
+    .. seealso::
+
+        Please refer to :class:`~tianshou.data.ReplayBuffer` for more detailed
+        explanation.
+    """
+
+    def __init__(self, buffer_list: List[ReplayBuffer], **kwargs: Any) -> None:
+        self.buffer_num = len(buffer_list)
+        self.buffers = buffer_list
```
ChenDRAG, line 475: can we use self.buffers = np.array(buffer_list) to support advanced indexing?

tianshou/data/buffer.py
```diff
@@ -464,69 +397,295 @@ def add(
...
+        if buffer_ids is None:
+            buffer_ids = np.arange(self.buffer_num)
+        # assume each element in buffer_ids is unique
+        assert np.bincount(buffer_ids).max() == 1
+        batch = Batch(obs=obs, act=act, rew=rew, done=done,
+                      obs_next=obs_next, info=info, policy=policy)
+        assert len(buffer_ids) == len(batch)
+        episode_lengths = []  # (len(buffer_ids),)
+        episode_rewards = []  # (len(buffer_ids), ...)
+        for batch_idx, env_id in enumerate(buffer_ids):
```
ChenDRAG: I prefer buffer_id rather than env_id

Trinkle23897 (dismissed):

ChenDRAG:
I make a minor change in CachedBuffer because of the need of #280

duburcqa:
I'm felling better now and I will review your PR right now :)

duburcqa:
> The current implementation use alloc_fn hook as an input argument, but in VectorReplayBuffer and CachedReplayBuffer, this approach is not compatible with pickle, because the user-defined hook cannot be pickled. You can run the script under test/base/test_buffer.py to see this issue.
> I haven't figure out the solution to this issue, could you please give us some suggestions?

Yes ! Just use `cloudpickle`, it is pretty standard and it is already a dependency of Tianshou through `gym` requirement ! It should allow pickling customs functions such as lambda for instance. I think it should solve this issue.

You mentioned that you solved this issue, but not how you did it x)

test/base/test_buffer.py
```diff
@@ -320,35 +410,204 @@ def test_hdf5():
...
     # ndarray with data type not supported by HDF5 that cannot be pickled
-    data = {"not_supported": np.array(lambda x: x*x)}
+    data = {"not_supported": np.array(lambda x: x * x)}
     grp = h5py.Group
     with pytest.raises(RuntimeError):
         to_hdf5(data, grp)
 
 
+def test_vectorbuffer():
+    buf = ReplayBuffers([ReplayBuffer(size=5) for i in range(4)])
```
duburcqa: I would rather split this test, for the same reason that the previous one. It is no longer clear what you are checking, making hard to maintain and update if the specification changes.

tianshou/data/buffer.py
```diff
@@ -11,123 +12,10 @@
 
 class ReplayBuffer:
     """:class:`~tianshou.data.ReplayBuffer` stores data generated from \
-    interaction between the policy and environment.
```
duburcqa, line 14: Maybe add a reference to the new location of this documentation.

tianshou/data/buffer.py
```diff
@@ -181,20 +79,95 @@ def __setstate__(self, state: Dict[str, Any]) -> None:
         We need it because pickling buffer does not work out-of-the-box
         ("buffer.__getattr__" is customized).
         """
-        self._indices = np.arange(state["_maxsize"])
         self.__dict__.update(state)
+        # compatible with previous version HDF5
```
duburcqa: what does 'previous version HDF5' means ? Not informative enough to be useful except for the original writer.

tianshou/data/buffer.py
```diff
@@ -181,20 +79,95 @@ def __setstate__(self, state: Dict[str, Any]) -> None:
...
+        self._indices = np.arange(self.maxsize)
 
-    def __getstate__(self) -> dict:
-        exclude = {"_indices"}
-        state = {k: v for k, v in self.__dict__.items() if k not in exclude}
-        return state
+    def __getstate__(self) -> Dict[str, Any]:
+        return self.__dict__
```
duburcqa: Why not to get rid of this method completely since it is the [default](https://docs.python.org/3/library/pickle.html#object.__getstate__).

tianshou/data/buffer.py
```diff
@@ -181,20 +79,95 @@ def __setstate__(self, state: Dict[str, Any]) -> None:
...
+        stack_num, buffer.stack_num = buffer.stack_num, 1
+        save_only_last_obs, self._save_only_last_obs = \
+            self._save_only_last_obs, False
+        indices = buffer.sample_index(0)  # get all available indices
+        for i in indices:
+            self.add(**buffer[i])  # type: ignore
+        buffer.stack_num = stack_num
+        self._save_only_last_obs = save_only_last_obs
+
+    def alloc_fn(self, key: List[str], value: Any) -> None:
```
duburcqa: I would rather define this in the constructor as a nested function by default, and stored as a private attribute.So that it is clear it is not supposed to be modified on the fly, but maybe I'm wrong about the intended use-cases.

tianshou/data/buffer.py
```diff
@@ -181,20 +79,95 @@ def __setstate__(self, state: Dict[str, Any]) -> None:
...
+        """
+        end_flag = self.done[index] | np.isin(index, self.unfinished_index())
+        return (index + (1 - end_flag)) % self._size
+
+    def update(self, buffer: "ReplayBuffer") -> None:
+        """Move the data from the given buffer to current buffer."""
+        if len(buffer) == 0 or self.maxsize == 0:
+            return
+        stack_num, buffer.stack_num = buffer.stack_num, 1
+        save_only_last_obs, self._save_only_last_obs = \
```
duburcqa: At this point, why not just split this line in two, since you need them anyway ? In particular if there is the same variable on both side of the assignment.

tianshou/data/buffer.py
```diff
@@ -241,117 +191,111 @@ def add(
...
             obs = obs[-1]
         self._add_to_buffer("obs", obs)
         self._add_to_buffer("act", act)
-        # make sure the reward is a float instead of an int
-        self._add_to_buffer("rew", rew * 1.0)  # type: ignore
-        self._add_to_buffer("done", done)
-        if self._save_s_:
+        # make sure the data type of reward is float instead of int
+        rew = rew * 1.0  # type: ignore
+        self._add_to_buffer("rew", rew)
```
duburcqa, line 208: better just do `self._add_to_buffer("rew", float(rew))` and remove all this mess (including comment since it is self-explainatory)

tianshou/data/buffer.py
```diff
@@ -241,117 +191,111 @@ def add(
...
-        if self._maxsize > 0:
-            self._size = min(self._size + 1, self._maxsize)
-            self._index = (self._index + 1) % self._maxsize
+        self._episode_reward += rew
+        self._episode_length += 1
+
+        if self.maxsize > 0:
+            self._size = min(self._size + 1, self.maxsize)
+            self._index = (self._index + 1) % self.maxsize
+        else:  # TODO: remove this after deleting ListReplayBuffer
```
duburcqa, line 222: can you explain me this comment a bit more ?

tianshou/data/buffer.py
```diff
@@ -241,117 +191,111 @@ def add(
...
+        index: Union[int, np.integer, np.ndarray],
         key: str,
         stack_num: Optional[int] = None,
     ) -> Union[Batch, np.ndarray]:
         """Return the stacked result.
 
         E.g. [s_{t-3}, s_{t-2}, s_{t-1}, s_t], where s is self.key, t is the
-        indice. The stack_num (here equals to 4) is given from buffer
-        initialization procedure.
+        indice.
```
duburcqa: index.

duburcqa:
Can you do the profiling before and after those modifications ? I'm wondering if it is harming the overall speed or the contrary.

docs/tutorials/concepts.rst
```diff
@@ -53,11 +53,157 @@ In short, you can define a :class:`~tianshou.data.Batch` with any key-value pair
...
+* ``obs_next`` the observation of step :math:`t+1` ;
+* ``info`` the info of step :math:`t` (in ``gym.Env``, the ``env.step()`` function returns 4 arguments, and the last one is ``info``);
+* ``policy`` the data computed by policy in step :math:`t`;
+
+The following code snippet illustrates its usage:
+
+::
+
+    >>> import pickle, numpy as np
+    >>> from tianshou.data import ReplayBuffer
```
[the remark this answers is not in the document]
Trinkle23897: But this is the ReplayBuffer's documentation. The Batch's usage is right above.

test/base/test_buffer.py
```diff
@@ -38,11 +39,13 @@ def test_replaybuffer(size=10, bufsize=20):
     assert (data.obs < size).all()
     assert (0 <= data.done).all() and (data.done <= 1).all()
     b = ReplayBuffer(size=10)
-    b.add(1, 1, 1, 'str', 1, {'a': 3, 'b': {'c': 5.0}})
+    # neg bsz should return empty index
+    assert b.sample_index(-1).tolist() == []
+    b.add(1, 1, 1, 1, 'str', {'a': 3, 'b': {'c': 5.0}})
     assert b.obs[0] == 1
-    assert b.done[0] == 'str'
```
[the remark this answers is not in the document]
Trinkle23897: Because in `buffer.add()`, done is convert to boolean

tianshou/data/buffer.py
Trinkle23897: My bad. 
```
# compatible with version == 0.3.1's HDF5 data format
```

tianshou/data/buffer.py
Trinkle23897: Could you please elaborate it (`just two lines`) a little bit more?

tianshou/data/buffer.py
Trinkle23897: `rew` may be `np.ndarray`:
```
In [1]: import numpy as np

In [2]: a=np.array([0,0,0])

In [3]: float(a)
`---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-3-6c8cade3f87c> in <module>
----> 1 float(a)

TypeError: only size-1 arrays can be converted to Python scalars
```

tianshou/data/buffer.py
Trinkle23897: This is used in ListReplayBuffer's add method. Its maxsize is 0 (theoretically it doesn't have maxsize because it saves data in a list).
BTW ListReplayBuffer will not be used in the new collector (#280)

Trinkle23897:
> Can you do the profiling before and after those modifications ? I'm wondering if it is harming the overall speed or the contrary.

2000 -> 1900, I'll have a look right now.

Trinkle23897:
> You mentioned that you solved this issue, but not how you did it x)

I remove `@staticmethod` in the previous implementation and overwrite the sub-buffer's alloc_fn in the initialization of ReplayBuffers:
```python
        for buf in self.buffers:
            # overwrite sub-buffers' alloc_fn so that
            # the top buffer can allocate new memory for all sub-buffers
            buf.alloc_fn = self.alloc_fn  # type: ignore
```

Trinkle23897:
> 2000 -> 1900, I'll have a look right now.

commit 425c2bd made the speed from 1980 to 1900

Trinkle23897:
> Yes ! Just use `cloudpickle`, it is pretty standard and it is already a dependency of Tianshou through `gym` requirement ! It should allow pickling customs functions such as lambda for instance. I think it should solve this issue.

But this still leads `pickle.save(buffer)` not working. The users should change it to `cloudpickle.save(buffer)`. :(

docs/tutorials/concepts.rst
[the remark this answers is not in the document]
duburcqa: Indeed x) I misread I don't know why x)

test/base/test_buffer.py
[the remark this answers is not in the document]
duburcqa: why not to check that then ?

tianshou/data/buffer.py
duburcqa:
```
save_only_last_obs = self._save_only_last_obs
self._save_only_last_obs = False
```

tianshou/data/buffer.py
duburcqa: AFAIK, the reward must be a scalar float according to OpenAI gym API. And `float(np.array([0]))` is fine.

tianshou/data/buffer.py
duburcqa: OK !

tianshou/data/buffer.py
Trinkle23897: Indeed, but we currently have a unittest for multi-agent (multi-agent env's reward is np.ndarray)

test/base/test_buffer.py
Trinkle23897: I'll re-organize this unit test and add some comments. I think it is messy too. That's my bad.

duburcqa:
> But this still leads pickle.save(buffer) not working. The users should change it to cloudpickle.save(buffer). :(

Indeed. You can load with pickle, but you must dump with `cloudpickle`:

```
>>> import cloudpickle
>>> squared = lambda x: x ** 2
>>> pickled_lambda = cloudpickle.dumps(squared)

>>> import pickle
>>> new_squared = pickle.loads(pickled_lambda)
>>> new_squared(2)
4
```

What you could do is to create a `Mixin` class, that is defining a generic `pickle_dump` method, that it calling `cloudpickle` instead of `pickle`, so that the user only as to call this method to dump (or even to load) objects that must be pickle-able.

It depends how much you expect to rely on hook that must be pickle-able on the future. If you think it is a special case, your workaround is fine,otherwise it is likely to make things for complicated than the contrary, just to avoid using a library that the user already has installed on the system.

Trinkle23897:
> If you think it is a special case, your workaround is fine

In my current thought it is a special case. I haven't come up with other demands or requirements similar to `alloc_fn`.

tianshou/data/buffer.py
```diff
@@ -232,8 +232,7 @@ def add(
             self._episode_length, self._episode_reward = 0, 0.0
             return result
         else:
-            return np.zeros_like(self._episode_length), \
-                np.zeros_like(self._episode_reward)
+            return self._episode_length * 0, self._episode_reward * 0.
```
duburcqa: what about `np.zeros(self._episode_length.shape)` ? Since `np.zeros_like` internal implementation since overly complicated.

duburcqa:
> In my current thought it is a special case. I haven't come up with other demands or requirements similar to alloc_fn.

Why not then ! But may be the name is not very appropriate if it is a instance method (supposedly private) instead of a hook ?

duburcqa:
> commit 425c2bd made the speed from 1980 to 1900

What are you using for profiling ? Personally I'm now using `py-spy` and it is just mesmerising (open the interactive `svg` in webbrowser)

`py-spy record --native -o profiling.svg -- python3 bench.py`

duburcqa:
> I remove @staticmethod in the previous implementation and overwrite the sub-buffer's alloc_fn in the initialization of ReplayBuffers

Hum... that's look kind of hacky but not meaningless actually. Is it  still pickle correctly after that change at instance-level ?

Trinkle23897:
> Hum... that's look kind of hacky but not meaningless actually. 

I think so, but I haven't come up with other approaches better than this. (compatible with pickle)

> Is it still pickle correctly after that change at instance-level ?

Sure. I add the related test under `test_hdf5`

Trinkle23897:
> But may be the name is not very appropriate if it is a instance method (supposedly private) instead of a hook ?

What's your suggestion? (Also I'm not sure if the name `ReplayBuffers` is good enough since it is only one character different than `ReplayBuffer`, but changing its name to `ReplayBufferList` will be further misleading because we already have `ListReplayBuffer`)

duburcqa:
> I'm not sure if the name ReplayBuffers is good enough since it is only one character different than ReplayBuffer, but changing its name to ReplayBufferList will be further misleading because we already have ListReplayBuffer

What about `ReplayBufferManager`, `ReplayBufferSupervisor`, `ReplayBufferOrchestrator`, `ReplayBufferDispatcher` ...

duburcqa:
> What's your suggestion?

I don't know, maybe just `_buffer_allocator` or `_buffer_dynamic_allocator`.

Trinkle23897:
> What are you using for profiling ? 

output

duburcqa:
Your profiling graph is polluted by non-significant method calls (such as import.load modules), you should use a script running for a longer duration. 

I was using this tool before, but I find it less readable than `py-spy` at the end. But both are good :)

Trinkle23897:
py-spy result [profiling.zip](https://github.com/thu-ml/tianshou/files/5888165/profiling.zip)

I can't find the part related to replaybuffer. Is there anything wrong?

duburcqa:
> I can't find the part related to replaybuffer. Is there anything wrong?

And yet it is here! Just use the search functionality integrated in the svg itself, and look for `buffer.py`

duburcqa:
By looking at the time spend in this module, I don't think it is relevant to optimize it further x)

NB: Just disable `--native` option to avoid showing time spend in `torch` if you prefer.

duburcqa (dismissed):

ChenDRAG:
I think cachedbuffer is ready now.

Trinkle23897 (approved):

Trinkle23897:
@duburcqa There is a question that makes me confused.

```python
import numpy as np
from tianshou.data import ReplayBuffer, CachedReplayBuffer
buf = CachedReplayBuffer(ReplayBuffer(10000), 1, 10000)
buf.add(obs=[np.random.rand(4, 84, 84)], act=[1], rew=[1], done=[1])
buf = CachedReplayBuffer(ReplayBuffer(10000), 1, 10000)
buf.add(obs=[np.random.rand(4, 84, 84)], act=[1], rew=[1], done=[1])
```

Repeating executing the last two lines, you'll see that the actual memory is still increasing. That's weird because theoretically the reference count of previous `buf = CachedReplayBuffer(...)` is 0. However it seems like it doesn't do the garbage collection. Do you have any idea?

duburcqa:
What about this ?
```
import gc
import numpy as np
from tianshou.data import ReplayBuffer, CachedReplayBuffer
for _ in range(100):
    buf = CachedReplayBuffer(ReplayBuffer(10000), 1, 10000)
    buf.add(obs=[np.random.rand(4, 84, 84)], act=[1], rew=[1], done=[1])
    gc.collect()
```

If there is an actual memory leak, you should be able to find the line that is responsible of it using [memory_profiler](https://github.com/pythonprofilers/memory_profiler). I'm using it frequently and it is quite good, even though time consuming to use. `mprof` especially is pretty useful.

Trinkle23897:
`gc.collect()` works fine. Thanks~

PR #284, opened 2021-02-02
fix 2 bugs of batch

Trinkle23897:
1. `_create_value(Batch(a={}, b=[1, 2, 3]), 10, False)`

before:
```python
TypeError: cannot concatenate with Batch() which is scalar
```
after:
```python
Batch(
    a: Batch(),
    b: array([0, 0, 0, 0, 0, 0, 0, 0, 0, 0]),
)
```

2. creating keys in a batch's subkey, e.g. 
```python
a = Batch(info={"key1": [0, 1], "key2": [2, 3]})
a[0] = Batch(info={"key1": 2, "key3": 4})
print(a)
```
before:
```python
Batch(
    info: Batch(
              key1: array([0, 1]),
              key2: array([0, 3]),
          ),
)
```
after:
```python
ValueError: Creating keys is not supported by item assignment.
```

3. small optimization for `Batch.stack_` and `Batch.cat_`

### DIFF
--- a/tianshou/data/batch.py
+++ b/tianshou/data/batch.py
@@ -246,8 +246,9 @@ class Batch:
         value: Any,
     ) -> None:
         """Assign value to self[index]."""
+        value = _parse_value(value)
         if isinstance(index, str):
-            self.__dict__[index] = _parse_value(value)
+            self.__dict__[index] = value
             return
         if not isinstance(value, (dict, Batch)):
             raise ValueError("Batch does not supported tensor assignment. "
--- a/tianshou/data/batch.py
+++ b/tianshou/data/batch.py
@@ -177,18 +177,17 @@ class Batch:
         copy: bool = False,
         **kwargs: Any,
     ) -> None:
-        if batch_dict is None:
-            if len(kwargs) == 0:
-                return
-            batch_dict = kwargs
         if copy:
             batch_dict = deepcopy(batch_dict)
-        if isinstance(batch_dict, (dict, Batch)):
-            _assert_type_keys(batch_dict.keys())
-            for k, v in batch_dict.items():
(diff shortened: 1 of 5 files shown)


tianshou/data/batch.py
```diff
@@ -496,6 +494,8 @@ def stack_(
         self, batches: Sequence[Union[dict, "Batch"]], axis: int = 0
     ) -> None:
         """Stack a list of Batch object into current batch."""
+        batches = [x for x in batches if isinstance(
```
duburcqa: I think at this point and noting the other quite complexe comprehension loop below, maybe a good old plain for-loop would be clearer and more efficient.

Trinkle23897: Agree. But we tend to optimize batch in the future, this is just a workaround.
Also if I don't add this line of code, what's the following result:
```python
Batch([[Batch(a=1)], [Batch(a=1)]])
```
I think it should be an illegal input, because it is not a `Sequence[Union[dict, Batch]]`.

duburcqa: Hum... I don't really know, it is hard to tell for sure. It may be useful to support it, but I can't come up with an example for which it would be actually useful. So I guess it is better to forbid it if there is no relevant usecase of it.

Trinkle23897: Yep, so that it returns an empty Batch as the temporary solution currently.

duburcqa: Could you document this point please ? It would be nice to keep track of it to not forget taking this aspect into consideration for future rework.

duburcqa: Emptiness sounds quite misleading, don't you think ? throwing an exception seems more appropriate no ?

duburcqa (approved):

PR #294, opened 2021-02-20
Improve buffer.prev() & buffer.next()

Trinkle23897:
Typically in atari training pipeline, buf.prev takes over 30% time. This pr aims to optimize buffer.prev/next: use njit to make it faster.


tianshou/data/buffer.py
```diff
@@ -470,29 +487,23 @@ def set_batch(self, batch: Batch) -> None:
...
     def prev(self, index: Union[int, np.integer, np.ndarray]) -> np.ndarray:
-        index = np.asarray(index) % self.maxsize
-        prev_indices = np.zeros_like(index)
-        for offset, buf in zip(self._offset, self.buffers):
-            mask = (offset <= index) & (index < offset + buf.maxsize)
-            if np.any(mask):
-                prev_indices[mask] = buf.prev(index[mask] - offset) + offset
-        return prev_indices
+        try:
+            len(index)  # type: ignore
```
Trinkle23897: This is ugly but I haven't figure out a better way, because index may be `np.array(3)`, at this time `type(index) == np.ndarray`, `hasattr(index, "__len__") == True`

ChenDRAG:
I prefer unfinished index over last_index.
Also, I think _lengths is redundant, there might be a better way

ChenDRAG (approved):

PR #300, opened 2021-02-28
fix vecenv action_space randomness

### LINKED ISSUE #299: action_space in SubprocVectorEnv make it harder to reproduce result.
```
            [lambda: gym.make(args.task) for _ in range(args.training_num)])
train_envs.action_space[0]
>Box(8,)
a = train_envs.action_space[0]
b = train_envs.action_space[0]
a is b
>False  # For SubprocVectorEnv, but for DummyVectorEnv the result is true.
```
This is because that  SubprocVectorEnv returns a deep copy but DummyVectorEnv return a reference.
As a result,
If you use SubprocVectorEnv and tianshou/collector, when you use 'random' option in collect() method, you will get different random action everytime you run the scripts even when using same random seeds.

So you have to be very careful about when you call 'train_envs.action_space[0].seed(1)'

Trinkle23897:
fix #299

### DIFF
--- a/tianshou/env/venvs.py
+++ b/tianshou/env/venvs.py
@@ -43,8 +43,7 @@ class BaseVectorEnv(gym.Env):
 
         Otherwise, the outputs of these envs may be the same with each other.
 
-    :param env_fns: a list of callable envs, ``env_fns[i]()`` generates the ith
-        env.
+    :param env_fns: a list of callable envs, ``env_fns[i]()`` generates the ith env.
     :param worker_fn: a callable worker, ``worker_fn(env_fns[i])`` generates a
         worker which contains the i-th env.
     :param int wait_num: use in asynchronous simulation if the time cost of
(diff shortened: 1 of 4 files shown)


tianshou/env/venvs.py
```diff
@@ -123,7 +120,12 @@ def __getattr__(self, key: str) -> List[Any]:
         wrapped environment, if it does not belong to the wrapping vector
         environment class.
         """
-        return [getattr(worker, key) for worker in self.workers]
+        result = [getattr(worker, key) for worker in self.workers]
+        if key == "action_space":
```
ChenDRAG: I don't know, but can we do better than just check for the single key:"action_space"

ChenDRAG:
There perhaps is another way to do this: In collector, set seed for self._action_spcae rather than train_envs. Both seems okay to me.

ChenDRAG (approved):

PR #303, opened 2021-03-02
fix venv seed, add TOC in docs, and split buffer.py into several files

### LINKED ISSUE #299: action_space in SubprocVectorEnv make it harder to reproduce result.

Trinkle23897:
patch for #299


Trinkle23897:
2021-03-01 22-48-30 的屏幕截图
- [x] Hyperlink doesn't work well
- [x] `E.g. [s_{t-3}, s_{t-2}, s_{t-1}, s_t], where s is self.key, t is the index.` This is unclear.
- [x] other types of buffer should mention "APIs are the same as ReplayBuffer"
- [x] move collector's example usage to docs/concepts.rst
- [x] modify trainer's `__init__` order
- [x] split buffer and collector into several files
- [x] policy docs need a TOC
- [x] venv.action_space should set seed in venv.seed(), keep a copy in local worker space

ChenDRAG (approved):

Issue #309, opened 2021-03-10
action mask is not used in 'compute_nstep_return'


ArthurWang10086:
It leads to unexpected results when the max q-value of the next state is calculated based on illegal actions.

Trinkle23897:
Do you mean in `_target_q` below, the target_q isn't masked? https://github.com/thu-ml/tianshou/blob/5c53f8c1f8aa0f4ed68ce9ea8455a89518db619c/tianshou/policy/modelfree/dqn.py#L74-L83

ArthurWang10086:
> Do you mean in `_target_q` below, the target_q isn't masked?
> 
> https://github.com/thu-ml/tianshou/blob/5c53f8c1f8aa0f4ed68ce9ea8455a89518db619c/tianshou/policy/modelfree/dqn.py#L74-L83
> 
> That's indeed a bug, thanks for pointing out!

only the action mask of obs is considered while the action mask of obs_next  not.

Trinkle23897:
Exactly. And would you like to make a pull request to fix it?

PR #310, opened 2021-03-14
fix qvalue mask_action error for obs_next

Trinkle23897:


danagi (commented):
I think the problem in #309 is caused by the function `_target_q` which uses unmasked logits to get target max q values in dqn.py. Since self(...).act is the legal max q value action, my personal suggestion is as follow:
```
model = 'model_old' if self._target else 'model'
result = self(batch, model=model, input='obs_next')
a = result.act
target_q = result.logits[np.arange(len(a)), a]
```

danagi (approved):

PR #319, opened 2021-03-22
Refactor PG algorithm and change behavior of `compute_episodic_return`

### LINKED ISSUE #307: Plans of releasing mujoco benchmark of onpolicy algorithms(VPG, A2C, PPO)
## Purpose
The purpose of this issue(discussion) is to introduce a series of pr in the near future targeted to releasing tianshou's benchmark for MuJoCo Gym task suite using onpolicy algorithms already supported by tianshou(VPG, A2C, PPO).
## Introduction
This issue is closely related #274, which mainly focus on benchmarking mujoco environments using offpolicy algorithm and enhancing tianshou along the way. Since Mujoco is a widely used task suite in literature, and most drl libraries haven't released satisfying benchmark on such important tasks. (While there are open-source implementations available that can be used by practitioners, lacking of graphs, source data, comparasion, and specific deatils still make it uneasy for starters to compare performance between algorithms.) We decided to try buiding a complete benchmark for MuJoCo Gym task suite. First step is #274, and the second is this one.

The most closely related work to ours is probably [spining up(Pytorch)](https://spinningup.openai.com/en/latest/spinningup/bench.html), which benchmarked 3 onpolicy algorithms(VPG, PPO, TRPO) on 5 mujoco environments, while our benchmark will try supporting more algorithms on 9 out of 13 environments. (Pusher, Thrower, Striker and HumanoidStandup have not been considered to be supported because they are not commonly seen in literatures.) While spinningup is intended for beginners and thus hasn't implemented standard tricks in drl (such as observation normalization and normalized value regression targets), we intend to do so.

Beyond that, like [offpolicy benchmark](https://github.com/thu-ml/tianshou/tree/master/examples/mujoco). for each supported algorithm, we will try providing:
- Default hyperparameters used for benchmark and scripts to reproduce the benchmark.
- A comparasion of performance(or code level details) with other open source implementations or classic papers.
- graphs and raw data that can be used for research purposes.
- Log details obtained during training.
- Pretrained agents.
- Some hints on how to tune the algorithm.

I make a plan and hope to finish the tasks above in the coming few weeks. Some features of tianshou will be enhanced along the way. I have done some experiments on my [fork](https://github.com/ChenDRAG/tianshou/tree/onmujoco_valuenorm) of Tianshou, which makes the benchmark below of PPO.

|Environment| Tianshou| [ikostrikov/pytorch-a2c-ppo-acktr-gail](https://github.com/ikostrikov/pytorch-a2c-ppo-acktr-gail)| [PPO paper](https://arxiv.org/pdf/1707.06347.pdf)|[baselines](http://htmlpreview.github.io/?https://github.com/openai/baselines/blob/master/benchmarks_mujoco1M.htm)|[spinningup(pytorch)](https://spinningup.openai.com/en/latest/spinningup/bench.html)|
| :---------------: | :---------------: | :---------------: | :---------------: | :---------------: |:---------------: |
|Ant|**3253.5+-894.3** | N | N | N | ~650 |
|HalfCheetah|**5460.5+-1319.3** | ~3120 | ~1800 | ~1700 | ~1670 |
|Hopper|2139.7+-776.6 | ~2300| ~2330 | **~2400** | ~1850 |
|Walker2d|3407.3+-1011.3 | **~4000** | ~3460 | ~3510 | ~1230 |
|Swimmer|99.8+-33.9 | N | ~108 | ~111 | **~120** |
|Humanoid|**597.4+-29.7** | N | N | N | N |
|Reacher|**-3.9+-0.4** | ~-5 | ~-7 | ~-6 | N |
|InvertedPendulum| N | N | **~1000** | ~940 | N |
|InvertedDoublePendulum|**8407.6+-1136.0** | N | ~8000 | ~7350 | N |

(This figure is outdated by now check examples/mujoco for better SOTA benchmark)

Example graph:
figure

## Plans
Here I briefly list 7 to do prs:

1.  Add a observation normalization wrapper of vecenv, because obs normalization affect onpolicy algorithm much.

2. Refactor VPG algorithm
- support value normalization and obs normalization.
- try different versions of VPG algorithm and make a benchmark on mujoco. 

3. Probably support natural policy gradient and make a benchmark.

4. Possibly support learning rate decay.

5.  Refactor A2C algorithm and make a benchmark

6. Refactor and benchmark PPO algorithm

7. Other enhancement
- providing drawing tools to reproduce benchmark to fix #161 .
- more loggger that can be used right away.
- maybe a fine_tuned version of ppo to use tricks guided by this [paper](https://arxiv.org/pdf/2006.05990.pdf).

## Future work
1.. Adding support(benchmark in the same way) for other algorithms(TRPO, ACER etc).

### LINKED ISSUE #317: Suggestion: Abandon name 'vpg' but use REINFORCE to replace it
As is stated in #307 , as far as what I know, vpg itself is not a very formal algorithm in the literature (first appears in Spinningup's docs, I think) and is loosely defined.  In SpinningUp's implementation, they use a generative advantage estimator(GAE) which requires a dnn value predictor(critic). This is more close to the definition of A2C(both in OPENAI baselines and in Tianshou). However, In tianshou, our current 'vpg' algorithm still follows strictly REINFORCE (use rew to go to update actor) algorithm pipeline.  As a result, I suggest we do not use the definition of vpg to avoid possible confusion, but use REINFORCE instead.

ChenDRAG:
See #307 , #317 for details. Note that I remove 'rew_norm' option from compute_episodic_return() because latter I will redefine how  to do value normalization(different algorithm may need different rew_norm schedule?)

### DIFF
--- a/tianshou/policy/base.py
+++ b/tianshou/policy/base.py
@@ -4,7 +4,7 @@ import numpy as np
 from torch import nn
 from numba import njit
 from abc import ABC, abstractmethod
-from typing import Any, Dict, Union, Optional, Callable
+from typing import Any, Dict, Union, Optional, Callable, Tuple
 
 from tianshou.data import Batch, ReplayBuffer, to_torch_as, to_numpy
 
@@ -257,7 +257,7 @@ class BasePolicy(ABC, nn.Module):
         v_s: Optional[Union[np.ndarray, torch.Tensor]] = None,
         gamma: float = 0.99,
         gae_lambda: float = 0.95,
-    ) -> Batch:
+    ) -> Tuple[np.ndarray, np.ndarray]:
         """Compute returns over given batch.
 
         Use Implementation of Generalized Advantage Estimator (arXiv:1506.02438)
--- a/tianshou/policy/base.py
+++ b/tianshou/policy/base.py
@@ -274,9 +274,7 @@ class BasePolicy(ABC, nn.Module):
         :param float gae_lambda: the parameter for Generalized Advantage Estimation,
             should be in [0, 1]. Default to 0.95.
 
-        # TODO change doc
-        :return: a Batch. The result will be stored in batch.returns as a numpy
-            array with shape (bsz, ).
+        :return: two numpy arrays (returns, advantage) with each shape (bsz, ).
         """
         rew = batch.rew
         if v_s_ is None:
--- a/tianshou/policy/base.py
+++ b/tianshou/policy/base.py
@@ -254,14 +254,14 @@ class BasePolicy(ABC, nn.Module):
         buffer: ReplayBuffer,
         indice: np.ndarray,
         v_s_: Optional[Union[np.ndarray, torch.Tensor]] = None,
+        v_s: Optional[Union[np.ndarray, torch.Tensor]] = None,
         gamma: float = 0.99,
         gae_lambda: float = 0.95,
-        rew_norm: bool = False,
     ) -> Batch:
         """Compute returns over given batch.
 
         Use Implementation of Generalized Advantage Estimator (arXiv:1506.02438)
-        to calculate q function/reward to go of given batch.
+        to calculate Q&A function/reward to go of given batch.
 
         :param Batch batch: a data batch which contains several episodes of data in
             sequential order. Mind that the end of each finished episode of batch
@@ -273,8 +273,8 @@ class BasePolicy(ABC, nn.Module):
         :param float gamma: the discount factor, should be in [0, 1]. Default to 0.99.
         :param float gae_lambda: the parameter for Generalized Advantage Estimation,
             should be in [0, 1]. Default to 0.95.
-        :param bool rew_norm: normalize the reward to Normal(0, 1). Default to False.
 
+        # TODO change doc
         :return: a Batch. The result will be stored in batch.returns as a numpy
             array with shape (bsz, ).
         """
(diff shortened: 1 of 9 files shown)


tianshou/policy/base.py
```diff
@@ -254,14 +254,14 @@ def compute_episodic_return(
...
+        v_s: Optional[Union[np.ndarray, torch.Tensor]] = None,
         gamma: float = 0.99,
         gae_lambda: float = 0.95,
-        rew_norm: bool = False,
     ) -> Batch:
         """Compute returns over given batch.
 
         Use Implementation of Generalized Advantage Estimator (arXiv:1506.02438)
-        to calculate q function/reward to go of given batch.
+        to calculate Q&A function/reward to go of given batch.
```
Trinkle23897: Q&A is not so meaningful, how about `Q value and advantage`?

tianshou/policy/modelfree/a2c.py
```diff
@@ -53,69 +53,42 @@ def __init__(
...
-        """
-        logits, h = self.actor(batch.obs, state=state, info=batch.info)
-        if isinstance(logits, tuple):
-            dist = self.dist_fn(*logits)
+                v_s_.append(to_numpy(self.critic(b.obs_next)))
+        v_s_ = np.concatenate(v_s_, axis=0)
+        if self._rew_norm:
+            # unnormalize v_s_
+            v_s_ = v_s_ * np.sqrt(self.ret_rms.var + self._eps) + self.ret_rms.mean
+        un_normalized_returns, _ = self.compute_episodic_return(
```
Trinkle23897: unnormalized_returns

tianshou/policy/modelfree/a2c.py
```diff
@@ -53,69 +53,42 @@ def __init__(
...
+            batch, buffer, indice, v_s_, gamma=self._gamma, gae_lambda=self._lambda)
```
Trinkle23897: better indicate v_s_=v_s_ for less ambiguous

Trinkle23897 (approved):

PR #328, opened 2021-03-27
Hotfix : keep statisics of buffer when reset buffer in on policy trainer

ChenDRAG:
For now, in on policy trainer, collector.reset_buffer() will reset statistics of buffer, which affects reward/length statistics. This pr offers  an option to fix that.


ChenDRAG:
@Trinkle23897  I'm not confident with this implementation(not familar with buffer implementation), can you check?

Trinkle23897 (approved):

PR #350, opened 2021-04-23
Make trainer resumable

### LINKED ISSUE #349: Resumable trainer?
I wonder if it is a common demand to save training progress and resume it. In my story I bothered by an unstable server that often shut down and made my train wasted. I made some changes to the trainer and, if it is a common demand, I will create a pr. But a discussion is needed to decide what param should be added to the trainer, like ``begin_step``, ``last_gradient_step``, etc. Should we read them from log? Or read it from a slot function that users should implement (like ``load_fn``).

Also, what data should be saved in the progress. Buffer, policy, and what else?

StephenArk30:
This is my simple idea: put everything (best_epoch and so on) in a class (TrainLog), and load/save all these things with hooked functions. I use a param ``epoch-per-save`` to save all these things every ``epoch-per-save`` epoch.

But the problem is that the loggers log every n step/epoch, and we restart our training process from ``k * epoch-per-save * step-per-epoch`` env_step, so the loggers will write reduplicative log. I wonder if we can log only when saving the training data.

### DIFF
--- a/docs/tutorials/cheatsheet.rst
+++ b/docs/tutorials/cheatsheet.rst
@@ -41,6 +41,7 @@ To resume training process from an existing checkpoint, you need to do the follo
 
 1. Make sure you write ``save_checkpoint_fn`` which saves everything needed in the training process, i.e., policy, optim, buffer; pass it to trainer;
 2. Use ``BasicLogger`` which contains a tensorboard;
+3. To adjust the save frequency, specify ``save_interval`` when initializing BasicLogger.
 
 And to successfully resume from a checkpoint:
 
--- a/docs/tutorials/cheatsheet.rst
+++ b/docs/tutorials/cheatsheet.rst
@@ -47,7 +47,7 @@ And to successfully resume from a checkpoint:
 1. Load everything needed in the training process **before trainer initialization**, i.e., policy, optim, buffer;
 2. set ``resume_from_log=True`` with trainer;
 
-We provide an example to show how these steps work: checkout `test_c51.py <https://github.com/thu-ml/tianshou/blob/master/test/discrete/test_c51.py>`_ by running
+We provide an example to show how these steps work: checkout `test_c51.py <https://github.com/thu-ml/tianshou/blob/master/test/discrete/test_c51.py>`_ or `test_ppo.py <https://github.com/thu-ml/tianshou/blob/master/test/continuous/test_ppo.py>`_ by running
 
 .. code-block:: bash
 
--- a/docs/tutorials/cheatsheet.rst
+++ b/docs/tutorials/cheatsheet.rst
@@ -30,6 +30,31 @@ Customize Training Process
 See :ref:`customized_trainer`.
 
 
+.. _resume_training:
+
+Resume Training Process
+-----------------------
+
+This is related to `Issue 349 <https://github.com/thu-ml/tianshou/issues/349>`_.
+
+To resume training process from an existing checkpoint, you need to do the following things in the training process:
+
+1. Make sure you write ``save_train_fn`` which saves everything needed in the training process, i.e., policy, optim, buffer; pass it to trainer;
+2. Use ``BasicLogger`` which contains a tensorboard;
+
+And to successfully resume from a checkpoint:
+
+1. Load everything needed in the training process **before trainer initialization**, i.e., policy, optim, buffer;
+2. set ``resume_from_log=True`` with trainer;
+
+We provide an example to show how these steps work: checkout `test_c51.py <https://github.com/thu-ml/tianshou/blob/master/test/discrete/test_c51.py>`_ by running
+
+.. code-block:: bash
+
+    $ python3 test/discrete/test_c51.py  # train some epoch
+    $ python3 test/discrete/test_c51.py --resume  # restore from existing log and continuing training
+
+
 .. _parallel_sampling:
 
 Parallel Sampling
--- a/docs/tutorials/cheatsheet.rst
+++ b/docs/tutorials/cheatsheet.rst
@@ -45,11 +45,11 @@ To resume training process from an existing checkpoint, you need to do the follo
 And to successfully resume from a checkpoint:
 
 1. Load everything needed in the training process **before trainer initialization**, i.e., policy, optim, buffer;
-2. set ``resume_from_log=True`` with trainer;
+2. Set ``resume_from_log=True`` with trainer;
 
-We provide an example to show how these steps work: checkout `test_c51.py <https://github.com/thu-ml/tianshou/blob/master/test/discrete/test_c51.py>`_ or `test_ppo.py <https://github.com/thu-ml/tianshou/blob/master/test/continuous/test_ppo.py>`_ by running
+We provide an example to show how these steps work: checkout `test_c51.py <https://github.com/thu-ml/tianshou/blob/master/test/discrete/test_c51.py>`_, `test_ppo.py <https://github.com/thu-ml/tianshou/blob/master/test/continuous/test_ppo.py>`_ or `test_il_bcq.py <https://github.com/thu-ml/tianshou/blob/master/test/discrete/test_il_bcq.py>`_ by running
 
-.. code-block:: bash
+.. code-block:: console
 
     $ python3 test/discrete/test_c51.py  # train some epoch
     $ python3 test/discrete/test_c51.py --resume  # restore from existing log and continuing training
--- a/docs/tutorials/cheatsheet.rst
+++ b/docs/tutorials/cheatsheet.rst
@@ -56,7 +56,7 @@ We provide an example to show how these steps work: checkout `test_c51.py <https
     $ python3 test/discrete/test_c51.py --resume  # restore from existing log and continuing training
 
 
-To correctly render the data (including several tfevent files), we highly recommend using ``tensorboard >= 2.5.0``.
+To correctly render the data (including several tfevent files), we highly recommend using ``tensorboard >= 2.5.0`` (see `here <https://github.com/thu-ml/tianshou/pull/350#issuecomment-829123378>`_ for the reason). Otherwise, it may cause overlapping issue that you need to manually handle with.
 
 .. _parallel_sampling:
 
--- a/docs/tutorials/cheatsheet.rst
+++ b/docs/tutorials/cheatsheet.rst
@@ -39,7 +39,7 @@ This is related to `Issue 349 <https://github.com/thu-ml/tianshou/issues/349>`_.
 
 To resume training process from an existing checkpoint, you need to do the following things in the training process:
 
-1. Make sure you write ``save_train_fn`` which saves everything needed in the training process, i.e., policy, optim, buffer; pass it to trainer;
+1. Make sure you write ``save_checkpoint_fn`` which saves everything needed in the training process, i.e., policy, optim, buffer; pass it to trainer;
 2. Use ``BasicLogger`` which contains a tensorboard;
 
 And to successfully resume from a checkpoint:
--- a/docs/tutorials/cheatsheet.rst
+++ b/docs/tutorials/cheatsheet.rst
@@ -56,6 +56,8 @@ We provide an example to show how these steps work: checkout `test_c51.py <https
     $ python3 test/discrete/test_c51.py --resume  # restore from existing log and continuing training
 
 
+To correctly render the data (including several tfevent files), we highly recommend using ``tensorboard >= 2.5.0``.
+
 .. _parallel_sampling:
 
 Parallel Sampling
--- a/test/discrete/test_il_bcq.py
+++ b/test/discrete/test_il_bcq.py
@@ -43,6 +43,7 @@ def get_args():
         default="cuda" if torch.cuda.is_available() else "cpu",
     )
     parser.add_argument("--resume", action="store_true")
+    parser.add_argument("--save-interval", type=int, default=4)
     args = parser.parse_known_args()[0]
     return args
 
@@ -86,7 +87,7 @@ def test_discrete_bcq(args=get_args()):
 
     log_path = os.path.join(args.logdir, args.task, 'discrete_bcq')
     writer = SummaryWriter(log_path)
-    logger = BasicLogger(writer)
+    logger = BasicLogger(writer, save_interval=args.save_interval)
 
     def save_fn(policy):
         torch.save(policy.state_dict(), os.path.join(log_path, 'policy.pth'))
--- a/test/discrete/test_il_bcq.py
+++ b/test/discrete/test_il_bcq.py
@@ -68,7 +68,7 @@ def test_discrete_bcq(args=get_args()):
         args.state_shape, args.action_shape,
         hidden_sizes=args.hidden_sizes, device=args.device).to(args.device)
     optim = torch.optim.Adam(
-        set(policy_net.parameters()).union(imitation_net.parameters()),
+        list(policy_net.parameters()) + list(imitation_net.parameters()),
         lr=args.lr)
 
     policy = DiscreteBCQPolicy(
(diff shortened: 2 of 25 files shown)


Trinkle23897:
I'll take a look tmr

Trinkle23897:
can we directly restore the metadata from tensorboard? can we continue writing to the same tensorboard?

Trinkle23897:
Here is my proposal: we can use `tensorboard.backend.event_processing.event_accumulator` reading the existing log, and extract them before each trainer's initialization process.

|                                        | offpolicy/onpolicy                                 | offline                                            |
| -------------------------------------- | -------------------------------------------------- | -------------------------------------------------- |
| best_epoch/best_reward/best_reward_std | load from existing tensorboard                     | load from existing tensorboard                     |
| env_step                               | load from existing tensorboard (the latest record) | /                                                  |
| gradient_step                          | load from existing tensorboard (the latest record) | load from existing tensorboard (the latest record) |
| last_rew/last_len                      | load from existing tensorboard (the latest record) | /                                                  |

I agree with adding extra fn like `epoch_save_fn` but actually we don't need `load_train_fn` -- because it only executes once, why not explicitly execute it by user before the definition of trainer?

test/discrete/test_il_bcq.py
```diff
@@ -93,10 +94,30 @@ def save_fn(policy):
...
+        }, os.path.join(log_path, 'checkpoint.pth'))
+
+    if args.resume:
+        # load from existing checkpoint
+        print(f"Loading agent under {log_path}")
+        ckpt_path = os.path.join(log_path, 'checkpoint.pth')
+        if os.path.exists(ckpt_path):
+            checkpoint = torch.load(ckpt_path, map_location=args.device)
+            policy.load_state_dict(checkpoint['model'])
+            # optim.load_state_dict(checkpoint['optim'])  # don't know why
```
Trinkle23897: This is strange. If I load optimizer from an existing checkpoint, it doesn't work:
```python
$ python3 test/discrete/test_il_bcq.py --resume                                                                                                                                                          [15:03:52]
2021-04-25 15:05:14.273532: I tensorflow/stream_executor/platform/default/dso_loader.cc:49] Successfully opened dynamic library libcudart.so.11.0
Loading agent under log/CartPole-v0/discrete_bcq
Successfully restore policy and optim.
Epoch #2:   0%|                                                                                                                                                                            | 0/2000 [00:00<?, ?it/s]
  File "test/discrete/test_il_bcq.py", line 142, in <module>
    test_discrete_bcq(get_args())
  File "test/discrete/test_il_bcq.py", line 116, in test_discrete_bcq
    result = offline_trainer(
  File "/home/trinkle/github/tianshou-new/tianshou/trainer/offline.py", line 94, in offline_trainer
    losses = policy.update(batch_size, buffer)
  File "/home/trinkle/github/tianshou-new/tianshou/policy/base.py", line 235, in update
    result = self.learn(batch, **kwargs)
  File "/home/trinkle/github/tianshou-new/tianshou/policy/imitation/discrete_bcq.py", line 117, in learn
    self.optim.step()
  File "/home/trinkle/.local/lib/python3.8/site-packages/torch/optim/optimizer.py", line 89, in wrapper
    return func(*args, **kwargs)
  File "/home/trinkle/.local/lib/python3.8/site-packages/torch/autograd/grad_mode.py", line 27, in decorate_context
    return func(*args, **kwargs)
  File "/home/trinkle/.local/lib/python3.8/site-packages/torch/optim/adam.py", line 108, in step
    F.adam(params_with_grad,
  File "/home/trinkle/.local/lib/python3.8/site-packages/torch/optim/_functional.py", line 96, in adam
    param.addcdiv_(exp_avg, denom, value=-step_size)
RuntimeError: output with shape [64] doesn't match the broadcast shape [64, 64]
```

Not sure what happens :(

Trinkle23897:
@StephenArk30 could you please help me verify the implementation (though I've added the test)?

test/discrete/test_il_bcq.py
Trinkle23897: This is because adam use `set(actor.parameters()).union(critic_parameters())` which causes random ordering in adam param groups.

StephenArk30:
But the logger logs (train) every n step (and test every epoch), and we save the model and the buffer every k epoch. Won't it be a problem? @Trinkle23897

Trinkle23897:
No, you can have a try.
Current approach: restore the env_step and gradient_step unrelated to epoch. So you will see the following situation:
- epoch 1: train 1000 env_step and 100 gradient_step
- epoch 1 test: save
- epoch 2: train 500 env_step and 50 gradient_step, killed
- restore
- epoch 3: start from env_step=1500 and gradient_step=150, load checkpoint from epoch 1 test

docs/tutorials/cheatsheet.rst
```diff
@@ -30,6 +30,31 @@ Customize Training Process
...
+
+1. Make sure you write ``save_train_fn`` which saves everything needed in the training process, i.e., policy, optim, buffer; pass it to trainer;
+2. Use ``BasicLogger`` which contains a tensorboard;
+
+And to successfully resume from a checkpoint:
+
+1. Load everything needed in the training process **before trainer initialization**, i.e., policy, optim, buffer;
+2. Set ``resume_from_log=True`` with trainer;
+
+We provide an example to show how these steps work: checkout `test_c51.py <https://github.com/thu-ml/tianshou/blob/master/test/discrete/test_c51.py>`_, `test_ppo.py <https://github.com/thu-ml/tianshou/blob/master/test/continuous/test_ppo.py>`_ or `test_il_bcq.py <https://github.com/thu-ml/tianshou/blob/master/test/discrete/test_il_bcq.py>`_ by running
```
Trinkle23897, line 51: scripts here

StephenArk30 (commented):
> No, you can have a try.
> Current approach: restore the env_step and gradient_step unrelated to epoch. So you will see the following situation:
> 
> * epoch 1: train 1000 env_step and 100 gradient_step
> * epoch 1 test: save
> * epoch 2: train 500 env_step and 50 gradient_step, killed
> * restore
> * epoch 3: start from env_step=1500 and gradient_step=150, load checkpoint from epoch 1 test

What if ``epoch-per-save = 2``

* epoch 2: train 1000 env_step and 100 gradient_step
* epoch 2 test: save
* **epoch 2 save model and buffer**
* epoch 3: train 1000 env_step and 100 gradient_step
* epoch 3 test: save
* epoch 4: train 500 env_step and 50 gradient_step, killed
* restore
* epoch 5 (not 4?): start from env_step=3500 and gradient_step=350, load checkpoint from epoch 3 test, **load epoch 2 model and buffer**

tianshou/trainer/offpolicy.py
```diff
@@ -118,6 +133,8 @@ def offpolicy_trainer(
                         if stop_fn(test_result["rew"]):
                             if save_fn:
                                 save_fn(policy)
+                            if save_train_fn:
+                                save_train_fn(epoch, env_step, gradient_step)
```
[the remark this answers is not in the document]
Trinkle23897: save_fn saves the best model while save_tain_fn saves periodically (including optim)

Trinkle23897:
> * epoch 5 (not 4?): start from env_step=3500 and gradient_step=350, load checkpoint from epoch 3 test, **load epoch 2 model and buffer**

Yeah but if you don't start at epoch5, the tensorboard log would be messy because it will overlap with previous epoch4

StephenArk30:
I agree with you. I think the key is to make model/buffer sync with step, otherwise we might restore a model with only 1000 gradient_step while resuming from 1500 gradient step read from log. In that case we will lose the 500 gradient step and make things unprecise.

So can we log data only before the save_train_fn? For instance:

```python
# offpolicy
for epoch in range(start_epoch + 1, max_epoch + 1):
    while t.n < t.total:
        result = train_collector.collect(n_step)
        env_step += n/st
        store env_step and result in a "train_log" batch
        ......  # store gradient_step and losses too
    test_result = test_episode(......)  # test but do not log
    if time to log:
        logger.log_train_data(train_log)
        logger.log_test_data(test_result)
        save_train_fn()
```

Trinkle23897:
How about now?
BTW, is there any other name better than `save_train_fn`?

danagi:
> How about now?
> BTW, is there any other name better than `save_train_fn`?

How about `save_checkpoint_fn` and `epoch_per_checkpoint`

StephenArk30:
``save_checkpoint_fn`` would be great. I think ``save/epoch`` still can't solve the problem, the model and the buffer still not sync with the log, or I may have some misunderstandings... I will try some examples later @Trinkle23897. @danagi please review the commits and run some tests if time permits.

Trinkle23897:
> I think `save/epoch` still can't solve the problem, the model and the buffer still not sync with the log,

why? `save/epoch`, `save/env_step` and `save/gradient_step` are only triggered when calling `logger.save_data`.

tianshou/utils/log_tools.py
```diff
@@ -133,15 +168,58 @@ def log_update_data(self, update_result: dict, step: int) -> None:
...
+        env_step: int,
+        gradient_step: int,
+        save_checkpoint_fn: Optional[Callable[[int, int, int], None]] = None,
+    ) -> None:
+        super().save_data(epoch, env_step, gradient_step, save_checkpoint_fn)
+        self.write("save/epoch", epoch, epoch)
+        self.write("save/env_step", env_step, env_step)
+        self.write("save/gradient_step", gradient_step, gradient_step)
+
+    def restore_data(self) -> Tuple[int, float, float, int, int, int, float, int]:
```
ChenDRAG: I think `restore` method should not be implemented in `Logger`, seems quite weird.

tianshou/utils/log_tools.py
```diff
@@ -42,15 +45,47 @@ def log_update_data(self, update_result: dict, step: int) -> None:
...
+        :param int epoch: the epoch in trainer.
+        :param int env_step: the env_step in trainer.
+        :param int gradient_step: the gradient_step in trainer.
+        :param function save_checkpoint_fn: a hook defined by user, see trainer
+            documentation for detail.
+        """
+        if save_checkpoint_fn:
+            save_checkpoint_fn(epoch, env_step, gradient_step)
+
+    def restore_data(self) -> Tuple[int, float, float, int, int, int, float, int]:
```
ChenDRAG: Same as L174

tianshou/trainer/offpolicy.py
```diff
@@ -24,6 +24,9 @@ def offpolicy_trainer(
     test_fn: Optional[Callable[[int, Optional[int]], None]] = None,
     stop_fn: Optional[Callable[[float], bool]] = None,
     save_fn: Optional[Callable[[BasePolicy], None]] = None,
+    save_checkpoint_fn: Optional[Callable[[int, int, int], None]] = None,
```
ChenDRAG, line 28: 1. What is the differences between save_fn and save_checkpoint_fn?
2. I think we add too much code just to support "resume training". More code means harder to read/rewrite. I don't know whether this single method worth this.

docs/tutorials/cheatsheet.rst
ChenDRAG: TBD

tianshou/utils/log_tools.py
```diff
@@ -104,12 +138,13 @@ def log_train_data(self, collect_result: dict, step: int) -> None:
...
                 self.last_log_train_step = step
 
-    def log_test_data(self, collect_result: dict, step: int) -> None:
+    def log_test_data(self, collect_result: dict, step: int, epoch: int) -> None:
         """Use writer to log statistics generated during evaluating.
 
         :param collect_result: a dict containing information of data collected in
             evaluating stage, i.e., returns of collector.collect().
         :param int step: stands for the timestep the collect_result being logged.
+        :param int epoch: stands for the epoch the collect_result being logged.
```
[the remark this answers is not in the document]
ChenDRAG: Also I think epoch should not appeare in `Logger`. For the logger itself, it shouldn't care about whether xlabel is `epoch` or `step` or anything else. That is `trainer`'s concern.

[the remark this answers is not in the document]
Trinkle23897: just wait for my next commit (not pushed yet)

tianshou/utils/log_tools.py
Trinkle23897: But if not, where should we put this functionality?

tianshou/trainer/offpolicy.py
Trinkle23897: save_fn saves the best policy, while save_checkpoint_fn saves periodically 

but this issue is required by many users (even one years ago! https://discuss.pytorch.org/t/is-there-a-way-to-resume-logging-checkpoint-while-using-tensorboard/82307)

StephenArk30:
try this:

```python
# test_c51.py
def stop_fn(mean_rewards):
    return False
    # return mean_rewards >= env.spec.reward_threshold
```

```shell
python test/discrete/test_c51.py --epoch 20 --epoch-per-save 4 --step-per-epoch 1000
# run 14 epoch
python test/discrete/test_c51.py --epoch 20 --epoch-per-save 4 --step-per-epoch 1000 --resume
```

I get this in tensorboard:

tensorboard

Trinkle23897:
> I get this in tensorboard:

@StephenArk30 how about now?

StephenArk30:
> > I get this in tensorboard:
> 
> @StephenArk30 how about now?

Looks better, thanks. I will try later.

ChenDRAG (dismissed):

Trinkle23897:
> Looks better, thanks. I will try later.

Please ping me if you have tested the code, thanks!

StephenArk30:
Still get messy log:




```shell
python test/discrete/test_c51.py --epoch 20 --step-per-epoch 500
# run 14 epoch and some step in 15 epoch (7344 steps in total)
python test/discrete/test_c51.py --epoch 20 --step-per-epoch 5000 --resume
```

We need to ensure ``step in train == step in test == save/env_step``.

Now if I stop at 7344 step:

- step in train: 7344
- step in test: 7056 (test after epoch 14)
- step in save: 5500 (save at epoch 11)

When I resume, the train and the test start from 5500, so the tensorboard draws backwards. You can debug to see this issue.

I suggest putting ``log_train_data`` and ``log_test_data`` into (or right after) ``save_data``, or log save/* in ``log_test_data`` @Trinkle23897 .

Trinkle23897:
which version of tensorboard do you use now? I use 2.5.0 and everything looks fine here.
2021-04-29 17-42-14 的屏幕截图

**Update: 2.4.1 is messy, so that I specify `tensorboard>=2.5.0` in setup.py**

Trinkle23897:
https://github.com/tensorflow/tensorboard/pull/4696

StephenArk30:
No, mine is also 2.5.0. @ChenDRAG  @danagi can you try this at your convenience?

However, I think this is not a tensorboard issue. If users use a custom logger with a different writer (like plt), they will find it difficult to handle the overlap (which is handled automatically by tensorboard according to the issue mentioned). If we insist on this implementation, we need to warn users about the overlap in the doc.

Trinkle23897:
But https://github.com/tensorflow/tensorboard/pull/4696 fix this issue. Could you please remove all logs and run again with tensorboard==2.5.0? I run several times and only <=2.4.1 will cause messy.

> However, I think this is not a tensorboard issue. If users use a custom logger with a different writer (like plt), they will find it difficult to handle the overlap (which is handled automatically by tensorboard according to the issue mentioned).

I don't think we can ignore this issue, because we cannot predict precisely to save checkpoint right before the program being killed. Instead, if we change the env_step and gradient_step as mentioned above, the actual behavior is similar to the plot in your comment https://github.com/thu-ml/tianshou/pull/350#issuecomment-827612796, so that I switch to current implementation.

Trinkle23897:
I test on mac M1 and the result is exactly the same: 2.5.0 works fine and 2.4.1 cannot work well.

Trinkle23897:
@StephenArk30 do you have any further comments?

danagi (dismissed):

Trinkle23897 (approved):

StephenArk30:
> @StephenArk30 do you have any further comments?

Went on a holiday, sorry. No further comments. Thanks for the great work!

PR #354, opened 2021-04-25
Support deterministic evaluation for onpolicy algorithms

### LINKED ISSUE #353: Deterministic sampling for PGPolicy
    + [ ] exception-raising bug
    + [ ] RL algorithm bug
    + [x] new feature request
  ```python
  ```

```
0.4.1 1.8.0 1.19.2 3.8.8 (default, Feb 24 2021, 21:46:12) 
[GCC 7.3.0] linux
```

I was wondering why tianshou's PGPolicy, A2CPolicy and PPOPolicy always use `dist.sample()`. This causes undeterminstic behavior. On the other hand, many other RL frameworks (e.g., rllib) provides ways to make it determinstic by using argmax (https://github.com/ray-project/ray/blob/cecfc3b43b7a1a7ec4398645556ef64c21704ea8/rllib/models/torch/torch_action_dist.py#L362).

It would be nice if tianshou can provide an init parameter that enables deterministic sampling in case of evaluation mode, e.g.,

```python
if self.training:
  act = self.dist_fn(logits).sample()
else:
  act = logits.argmax()
```

or maybe can we make the `dist_fn` aware of current mode?

ultmaster:
    + [ ] exception-raising fix
    + [x] new feature
- [x] If applicable, I have mentioned the relevant/related issue(s) #353

  ```python
  ```

### DIFF
--- a/tianshou/policy/base.py
+++ b/tianshou/policy/base.py
@@ -5,6 +5,7 @@ from torch import nn
 from numba import njit
 from abc import ABC, abstractmethod
 from typing import Any, Dict, Tuple, Union, Optional, Callable
+from gym.spaces import Box, Discrete, MultiDiscrete, MultiBinary
 
 from tianshou.data import Batch, ReplayBuffer, to_torch_as, to_numpy
 
@@ -66,6 +67,11 @@ class BasePolicy(ABC, nn.Module):
         super().__init__()
         self.observation_space = observation_space
         self.action_space = action_space
+        self.action_type = ""
+        if isinstance(action_space, (Discrete, MultiDiscrete, MultiBinary)):
+            self.action_type = "discrete"
+        elif isinstance(action_space, Box):
+            self.action_type = "continuous"
         self.agent_id = 0
         self.updating = False
         self.action_scaling = action_scaling
--- a/tianshou/policy/modelfree/pg.py
+++ b/tianshou/policy/modelfree/pg.py
@@ -25,6 +25,9 @@ class PGPolicy(BasePolicy):
         to use option "action_scaling" or "action_bound_method". Default to None.
     :param lr_scheduler: a learning rate scheduler that adjusts the learning rate in
         optimizer in each policy.update(). Default to None (no lr_scheduler).
+    :param deterministic_eval: whether to use deterministic action (maximum
+        value in the logits) instead of stochastic action sampled by the policy.
+        Default to False.
 
     .. seealso::
 
--- a/tianshou/policy/modelfree/pg.py
+++ b/tianshou/policy/modelfree/pg.py
@@ -1,11 +1,10 @@
-import gym
 import torch
 import numpy as np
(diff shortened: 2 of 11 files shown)


tianshou/policy/modelfree/pg.py
```diff
@@ -103,7 +108,13 @@ def forward(
             dist = self.dist_fn(*logits)
         else:
             dist = self.dist_fn(logits)
-        act = dist.sample()
+        if self._deterministic_eval and not self.training:
+            if isinstance(logits, tuple):
+                act = logits[0].argmax(-1)
```
Trinkle23897: Currently tuple means `(mu, sigma)` for continuous action space, and it seems like using `act = logits[0]` is better.

Another thing: we should check the action space's type in this case, discrete or continuous.

ultmaster: Maybe we can check whether dist_fn is `torch.distributions.Categorical`. If it is, we apply an extra argmax?

Does it make sense?

I think it would be more convenient if `torch.distributions.Distribution` had a `deterministic_sample()`, but it didn't...

Trinkle23897: > Maybe we can check whether dist_fn is `torch.distributions.Categorical`.

That's not a universal approach. If someone wants to use another dist_fn for discrete action space and `deterministic_eval=True`, it won't result in the expected behavior.

ultmaster: Yes. But without argmax, the action shape (action space) would be different in training and evaluation. I guess it's not a good idea to change the corresponding environment in evaluation. What's your suggestion?

Can we enforce user to implement a `determinstic_sample` in `dist_fn` if they want to use this feature, and throw an error if flag is true but the method is not implemented?

Trinkle23897: > Can we enforce user to implement

Better not. It would be much difficult when using DiagGaussian distribution.

> What's your suggestion?

Well, my current thought is to check the type of action_space in `BasePolicy.__init__`, and use it here.

ultmaster: Make sense to me. Implementation updated.

tianshou/policy/base.py
```diff
@@ -66,6 +67,11 @@ def __init__(
         super().__init__()
         self.observation_space = observation_space
         self.action_space = action_space
+        self.action_type = ""
+        if isinstance(action_space, (Discrete, MultiDiscrete, MultiBinary)):
+            self.action_type = "discrete"
+        elif isinstance(action_space, Box):
+            self.action_type = "continuous"
```
ultmaster, line 74: I think `gym.spaces` is more than 4 types. Would `self.action_type == ""` cause any problems?

Trinkle23897: "" eventually handles Dict/Tuple ~or other types~ no other types.

danagi (approved):

ChenDRAG (approved):

PR #359, opened 2021-05-05
Add discrete Conservative Q-Learning for offline RL

### LINKED ISSUE #358: Conservative Q-Learning (CQL) for offline/batch reinforcement learning
    + [ ] exception-raising bug
    + [ ] RL algorithm bug
    + [x] new feature request
  ```python
  ```
I would like to contribute an implementation of the discrete version of [Conservative Q-Learning (CQL)](https://arxiv.org/pdf/2006.04779.pdf), which claims to be better than prior works in some offline/batch tasks. I mostly used the authors' [code](https://github.com/aviralkumar2907/CQL) as reference and tried to follow the contributing guidelines.

A PR will be created soon. Your feedback is highly appreciated.

nuance1979:
- A tianshou implementation of the discrete Conservative Q-Learning for offline/batch RL
- Reference: [paper](https://arxiv.org/pdf/2006.04779.pdf) and [code](https://github.com/aviralkumar2907/CQL)


examples/atari/README.md
```diff
@@ -67,4 +67,19 @@ We test our BCQ implementation on two example tasks (different from author's ver
...
+- Train an expert, by using the command listed in the above QRDQN section;
+- Generate buffer with noise: `python3 atari_qrdqn.py --task {your_task} --watch --resume-path log/{your_task}/qrdqn/policy.pth --eps-test 0.2 --buffer-size 1000000 --save-buffer-name expert.hdf5` (note that 1M Atari buffer cannot be saved as `.pkl` format because it is too large and will cause error);
+- Train CQL: `python3 atari_cql.py --task {your_task} --load-buffer-name expert.hdf5`.
+
+We test our CQL implementation on two example tasks (different from author's version, we use v4 instead of v0; one epoch means 10k gradient step):
+
+| Task                   | Online QRDQN | Behavioral | CQL                               |
+| ---------------------- | ---------- | ---------- | --------------------------------- |
+| PongNoFrameskip-v4     | 20.5         |         | 21 (epoch 5)                      |
+| BreakoutNoFrameskip-v4 | 394.3        |          |  (epoch 12, could be higher) |
```
Trinkle23897: `behavioral` is the mean episodic return over all noisy data in replay buffer (eps=0.2)

Trinkle23897:
`pytest` uses alphabetical order to test scripts under the same directory (for now), so that `cql` would run before `dqn`. You can either modify the code to add test dependency in pytest, or change the filename to `test_il_cql`.

examples/atari/README.md
nuance1979: Great! I was about to ask. Is there a way to compute it quickly? I feel this should be a script or a utility function.

Trinkle23897: just print out the `collector.collect` result, `result['rews'].mean()` is what you want

nuance1979: After tweaking some hyper parameters, I was able to get a decent result on BreakoutNoFrameskip-v4. Turned out I need to increase the weight for the min_q_loss to 50 for Breakout. I also tried 0 weight, which is equivalent to an offline QRDQN, and the test result was 0, which means the min_q_loss/regularizer was indeed doing its job.

I think this PR is complete for now. Please review.

Trinkle23897: So how about the performance?

nuance1979: Sorry. Forgot to push. See README.md .

examples/atari/README.md
```diff
@@ -67,4 +67,19 @@ We test our BCQ implementation on two example tasks (different from author's ver
...
+- Train CQL: `python3 atari_cql.py --task {your_task} --load-buffer-name expert.hdf5`.
+
+We test our CQL implementation on two example tasks (different from author's version, we use v4 instead of v0; one epoch means 10k gradient step):
+
+| Task                   | Online QRDQN | Behavioral | CQL                               | parameters                                                   |
+| ---------------------- | ---------- | ---------- | --------------------------------- | ------------------------------------------------------------ |
+| PongNoFrameskip-v4     | 20.5         | 6.8        | 19.5 (epoch 5)                      | `python3 atari_cql.py --task "PongNoFrameskip-v4" --load-buffer-name log/PongNoFrameskip-v4/qrdqn/expert.hdf5 --epoch 5` |
+| BreakoutNoFrameskip-v4 | 394.3        | 46.9       | 248.3 (epoch 12) | `python3 atari_cql.py --task "BreakoutNoFrameskip-v4" --load-buffer-name log/BreakoutNoFrameskip-v4/qrdqn/expert.hdf5 --epoch 12 --min-q-weight 50` |
```
Trinkle23897, line 85: I'd like to suggest using `--eps-test 0.9` or `--eps-test 0.99` when generating buffer to see what happens because, in the original paper, the author used 1% and 10% of the expert data to train CQL and got a good result (Table 3).

Trinkle23897: I've tried `--eps-test 0.9` on Pong and Breakout. Pong can easily achieve +20 reward but it is not so stable; Breakout cannot achieve >20 reward when `min-q-weight` is either 10 or 50. I also test with only first 10% random data instead of mixed 10% data, the results are the same. Could you please help check what's going wrong?

examples/atari/README.md
```diff
@@ -83,3 +83,19 @@ We test our CQL implementation on two example tasks (different from author's ver
...
+| BreakoutNoFrameskip-v4 | 394.3        | 41.4       | 40.8 (epoch 12) | `python3 atari_cql.py --task "BreakoutNoFrameskip-v4" --load-buffer-name log/BreakoutNoFrameskip-v4/qrdqn/expert.size_1e5.hdf5 --epoch 12 --min-q-weight 20` |
+
+Buffer size 10000:
+
+| Task                   | Online QRDQN | Behavioral | CQL                               | parameters                                                   |
+| ---------------------- | ---------- | ---------- | --------------------------------- | ------------------------------------------------------------ |
+| PongNoFrameskip-v4     | 20.5         | nan        | 1.8 (epoch 5)                      | `python3 atari_cql.py --task "PongNoFrameskip-v4" --load-buffer-name log/PongNoFrameskip-v4/qrdqn/expert.size_1e4.hdf5 --epoch 5 --min-q-weight 1` |
+| BreakoutNoFrameskip-v4 | 394.3        | 31.7       | 22.5 (epoch 12) | `python3 atari_cql.py --task "BreakoutNoFrameskip-v4" --load-buffer-name log/BreakoutNoFrameskip-v4/qrdqn/expert.size_1e4.hdf5 --epoch 12 --min-q-weight 10` |
```
Trinkle23897, line 101: There's no need to stop at epoch 5 and 12 -- that's too small and not converge yet. In fact, the original paper indicates that they use ~25 iterations to get the optimal performance in Breakout, is that (25*1M/64/10000=39) epoch in our setting?

Trinkle23897 (dismissed):

Trinkle23897 (approved):

examples/atari/README.md
nuance1979: It's tensorboard. I usually run it like this: `tensorboard --logdir log/BreakoutNoFrameskip-v4/cql/ --host $(hostname -i) --port 8086`. Then click the server address to open in a browser.

Trinkle23897: May I ask how to change the css style? Because mine is orange and looks quite different to yours:
2021-05-12 09-09-58 的屏幕截图

nuance1979: I don't know. I had many runs in my directory so tensorboard used all kinds of colors and it just happened that the two runs that I cared about had nice colors.

PR #367, opened 2021-05-14
Add discrete Critic Regularized Regression

### LINKED ISSUE #366: Critic Regularized Regression (CRR) for offline/batch reinforcement learning
    + [ ] exception-raising bug
    + [ ] RL algorithm bug
    + [x] new feature request
  ```python
  ```

I would like to contribute an implementation of the discrete version of [Critic Regularized Regression (CRR)](https://arxiv.org/pdf/2006.15134.pdf) for offline/batch reinforcement learning, which came out around the same time as CQL. Although the original paper only had experiments on continuous control tasks, it was equally applicable to discrete control tasks, on which I was interested in evaluating its performance. I used [this repo](https://github.com/Kmeco/offline-rl) as reference and tried to follow the contributing guidelines.

A PR will be created soon. Your feedback is highly appreciated.

nuance1979:
- Add discrete Critic Regularized Regression
- Reference: [paper](https://arxiv.org/pdf/2006.15134.pdf) and [code](https://github.com/Kmeco/offline-rl)
- Note that CRR itself does not seem to work well on Atari tasks but adding CQL loss/regularizer helps.

### DIFF
--- a/test/discrete/test_il_crr.py
+++ b/test/discrete/test_il_crr.py
@@ -1,0 +1,112 @@
+import os
+import gym
+import torch
+import pickle
+import pprint
+import argparse
+import numpy as np
+from torch.utils.tensorboard import SummaryWriter
+
+from tianshou.data import Collector
+from tianshou.utils import BasicLogger
+from tianshou.env import DummyVectorEnv
+from tianshou.utils.net.common import Net
+from tianshou.trainer import offline_trainer
+from tianshou.policy import DiscreteCRRPolicy
+
+
+def get_args():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--task", type=str, default="CartPole-v0")
+    parser.add_argument("--seed", type=int, default=1626)
+    parser.add_argument("--eps-test", type=float, default=0.001)
+    parser.add_argument("--lr", type=float, default=7e-4)
+    parser.add_argument("--gamma", type=float, default=0.99)
+    parser.add_argument("--n-step", type=int, default=3)
+    parser.add_argument("--target-update-freq", type=int, default=320)
+    parser.add_argument("--epoch", type=int, default=5)
+    parser.add_argument("--update-per-epoch", type=int, default=1000)
+    parser.add_argument("--batch-size", type=int, default=64)
+    parser.add_argument('--hidden-sizes', type=int,
+                        nargs='*', default=[64, 64])
+    parser.add_argument("--test-num", type=int, default=100)
+    parser.add_argument("--logdir", type=str, default="log")
+    parser.add_argument("--render", type=float, default=0.)
+    parser.add_argument(
+        "--load-buffer-name", type=str,
+        default="./expert_DQN_CartPole-v0.pkl",
+    )
+    parser.add_argument(
+        "--device", type=str,
+        default="cuda" if torch.cuda.is_available() else "cpu",
+    )
+    args = parser.parse_known_args()[0]
+    return args
+
+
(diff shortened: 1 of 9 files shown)


test/discrete/test_il_crr.py
```diff
@@ -0,0 +1,112 @@
...
+        stop_fn=stop_fn, save_fn=save_fn, logger=logger)
+
+    assert stop_fn(result['best_reward'])
+
+    if __name__ == '__main__':
+        pprint.pprint(result)
+        # Let's watch its performance!
+        env = gym.make(args.task)
+        policy.eval()
+        policy.set_eps(args.eps_test)
```
Trinkle23897:
```python
  File "test/discrete/test_il_crr.py", line 112, in <module>
    test_discrete_crr(get_args())
  File "test/discrete/test_il_crr.py", line 104, in test_discrete_crr
    policy.set_eps(args.eps_test)
  File "/Users/trinkle23897/miniconda3/lib/python3.8/site-packages/torch/nn/modules/module.py", line 778, in __getattr__
    raise ModuleAttributeError("'{}' object has no attribute '{}'".format(
torch.nn.modules.module.ModuleAttributeError: 'DiscreteCRRPolicy' object has no attribute 'set_eps'
```

nuance1979:
FYI, I tried a variant of CRR where I replaced the DQN critic with a QRDQN (num_quantiles=200). It was slightly better but not much: For Breakout, after 40 iterations, I got 161.4 for QRCRR vs 137.2 for CRR; for Pong, after 40 iterations, I got 18.5 for QRCRR vs 16.8 for CRR.

tianshou/policy/imitation/discrete_crr.py
```diff
@@ -0,0 +1,121 @@
...
+    ) -> None:
+        super().__init__(
+            actor,
+            optim,
+            lambda x: Categorical(logits=x),  # type: ignore
+            discount_factor,
+            reward_normalization,
+            **kwargs,
+        )
+        self.actor = actor
```
Trinkle23897: remove this line if base class is PGPolicy
```suggestion
```

Trinkle23897 (commented):
Have you ever considered inheriting from QRDQN or DiscreteCQL instead of PG? Because the only reused part is `forward` and it is really simple (just a few line, just `discrete`, no other cases in PGPolicy) -- and maybe that would be more concise?

nuance1979:
> Have you ever considered inheriting from QRDQN or DiscreteCQL instead of PG? Because the only reused part is `forward` and it is really simple (just a few line, just `discrete`, no other cases in PGPolicy) -- and maybe that would be more concise?

Yes, I have. But after some thought, I chose to inherit from PGPolicy to emphasize the fact that CRR is a policy gradient model, i.e., the critic, be it DQN, QRDQN or else, is only a by-product of the training process and does **NOT** contribute to the evaluation of the policy. I also thought about multiple inheritance, i.e., let CRR inherits from both PGPolicy and DQN or QRDQN or DiscreteCQL but I realized that it would not make the implementation simpler. For example, because none of them has a `self.actor_old` model, I still need to override the `def sync_weight()` method in order to synchronize the target actor model.

The only place I can think of improving is to refactor the `def learn()` method of the parent class in order to share some code. But even if I do that, that's only going to save me <10 lines of code at the expense of fragmenting the logic of `def learn()` function.

Anyway that's just my opinion. I can certainly change it if you insist.

Trinkle23897 (approved):

PR #368, opened 2021-05-19
Allow researchers to choose whether to use Double DQN

StephenArk30:
In my research, I need to compare the performance of modified DQN(double, dueling, ...) with the origin DQN (2013 and 2015), and I couldn't find a way to make the DQN "origin". So I add a param to control whether to use double DQN in DQN policy.

Please check if it's the right implementation.

### DIFF
--- a/.github/workflows/extra_sys.yml
+++ b/.github/workflows/extra_sys.yml
@@ -9,7 +9,7 @@ jobs:
     strategy:
       matrix:
         os: [macos-latest, windows-latest]
-        python-version: [3.6, 3.7, 3.8]
+        python-version: [3.7, 3.8]
     steps:
     - uses: actions/checkout@v2
     - name: Set up Python ${{ matrix.python-version }}
@@ -24,4 +24,4 @@ jobs:
         python -m pip install ".[dev]" --upgrade
     - name: Test with pytest
       run: |
-        pytest test/base test/continuous --ignore-glob "*env.py" --cov=tianshou --durations=0 -v
+        pytest test/base test/continuous --cov=tianshou --durations=0 -v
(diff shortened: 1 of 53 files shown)


Trinkle23897:
Why not set target_update_freq=0?

StephenArk30:
> Why not set target_update_freq=0?

I think target q network is from [DQN version 2015 (NIPS 2015)](https://web.stanford.edu/class/psych209/Readings/MnihEtAlHassibis15NatureControlDeepRL.pdf) but not Double DQN. The [DDQN paper](https://arxiv.org/pdf/1509.06461.pdf) also mentioned this.

The difference between DDQN and Nature DQN is that it uses the online network to get the best action, and uses target network to calculate Q. As the paper wrote:

> We therefore propose to evaluate the greedy policy according to the online network, but using the target network to estimate its value.

Trinkle23897:
I see, the difference is target_q estimation. But only adding such an argument is kind of weird because DQNPolicy has many subclasses -- C51, QRDQN, DiscreteBCQ, ..., how about them related to this issue...?

StephenArk30:
What about adding a DDQNPolicy?

Trinkle23897:
Hmm... I need other's opinion. @danagi

Trinkle23897 (approved):

p-veloso:
How do I access the tianshou version with DDQN? I installed the latest version/0.4.1 using

`pip install git+https://github.com/thu-ml/tianshou.git@master --upgrade 
`

What is confusing for me is that according to github 0.4.1 is the latest version. However, while the documentations for the master and latest versions have the argument is_double, the installation and documentation for 0.4.1 do not have it. How do I get a version with DDQN?

Trinkle23897:
Hmm… my bad. Currently you need to directly install from master branch instead of a specific tagged version.

PR #375, opened 2021-05-23
support arrays with different shape in batch

### LINKED ISSUE #362: What is the old version of numpy and networkx so the nx.Graph() can be saved as obs
    + [ ] exception-raising bug
    + [ ] RL algorithm bug
    + [X] documentation request (i.e. "X is missing from the documentation.")
    + [ ] new feature request
- [X] I have mentioned version numbers, operating system and environment, where applicable:
  ```python
  0.4.1 1.8.1+cpu 1.20.2 3.8.5 (default, Sep  3 2020, 21:29:08) [MSC v.1916 64 bit (AMD64)] win32
  ```

Hi, @Trinkle23897 
 From the usage here, it seems the nx.Graph() cannot be used as before.  May I ask that, how canI still use this feature, since directly saving the graph object is more efficient, instead of the edge_list array.

Maybe in the future, tianshou can support the dgl.graph object. It stores the tensor internally. 

Thanks!
> 这个例子可能现在不太能work，因为numpy升级了，以及nx.Graph重写了__getitem__，导致np.array([nx.Graph()])会出来空的> 数组……
> 不过正常的自定义class应该没啥问题
>    import networkx as nx
>   b = ReplayBuffer(size=3)
>  b.add(obs=nx.Graph(), act=0, rew=0, done=0)
> print(b)

### LINKED ISSUE #369: Batching with a variable action space
I am trying to train an agent for a board game, Blokus, which has a potentially large (~32K) action space, while only a small fraction of them are valid by the rule per step. I am implementing the environment which provides all valid actions as observation per step (the shape of observation differs every step). Next I plan to train a DQN agent, but I am not sure how replay buffer batches work with a variable action space. I have read the tic-tac-toe example in the docs. What should I change from tic-tac-toe to adapt to a variable number of actions per step? Thanks!

Trinkle23897:
partially resolve #362 but I'm not sure if it can work in general cases. Test code:

```python
import gym
import numpy as np
import networkx as nx
from tianshou.data import Batch, VectorReplayBuffer
from tianshou.env import DummyVectorEnv, SubprocVectorEnv


class GraphEnv(gym.Env):
    def __init__(self, size):
        self.size = size
        self.feat_dim = 32
        # generate a random graph with size number of nodes
        self.graph = nx.random_regular_graph(2, size)

    def reset(self):
        # node feats
        node_feats = np.random.rand(self.size, self.feat_dim)
        edge_list = np.array(self.graph.edges())
        print(edge_list.shape, node_feats.shape, edge_list.dtype)
        return {"edge_list": np.asarray(edge_list, dtype=object), "node_feats": node_feats}

    def step(self, action):
        node_feats = np.random.rand(
            self.size, self.feat_dim)
        edge_list = np.array(self.graph.edges())
        print(edge_list.shape, node_feats.shape)
        return {"edge_list": np.asarray(edge_list, dtype=object), "node_feats": node_feats}, 1, False, {}

    def seed(self, seed=None):
        pass


if __name__ == '__main__':
    # train_envs = SubprocVectorEnv([lambda: GraphEnv(i) for i in [5, 10, 15]])
    train_envs = DummyVectorEnv([lambda i=x: GraphEnv(size=i) for x in [5, 10, 15]])

    env_obs = train_envs.reset()
    b = VectorReplayBuffer(total_size=3, buffer_num=3)

    b.add(Batch(obs=env_obs, act=[0, 0, 0], rew=[0, 0, 0],
                done=[0, 0, 0]), buffer_ids=np.arange(len(env_obs)))
    print(b[[1, 2]].obs.shape)
```

### DIFF
--- a/tianshou/data/batch.py
+++ b/tianshou/data/batch.py
@@ -526,7 +526,12 @@ class Batch:
             elif all(isinstance(e, (Batch, dict)) for e in v):  # third often
                 self.__dict__[k] = Batch.stack(v, axis)
             else:  # most often case is np.ndarray
-                self.__dict__[k] = _to_array_with_correct_type(np.stack(v, axis))
+                try:
+                    self.__dict__[k] = _to_array_with_correct_type(np.stack(v, axis))
+                except ValueError:
+                    warnings.warn("You are using tensors with different shape,"
+                                  " fallback to dtype=object by default.")
+                    self.__dict__[k] = np.array(v, dtype=object)
         # all the keys
         keys_total = set.union(*[set(b.keys()) for b in batches])
         # keys that are reserved in all batches
--- a/tianshou/data/batch.py
+++ b/tianshou/data/batch.py
@@ -153,10 +153,23 @@ def _alloc_by_keys_diff(
 ) -> None:
     for key in batch.keys():
         if key in meta.keys():
-            if isinstance(meta[key], Batch) and isinstance(batch[key], Batch):
-                _alloc_by_keys_diff(meta[key], batch[key], size, stack)
-            elif isinstance(meta[key], Batch) and meta[key].is_empty():
-                meta[key] = _create_value(batch[key], size, stack)
+            if isinstance(meta[key], Batch):
+                if isinstance(batch[key], Batch):
+                    _alloc_by_keys_diff(meta[key], batch[key], size, stack)
+                elif meta[key].is_empty():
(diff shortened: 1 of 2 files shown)


Trinkle23897:
@duburcqa could you please kindly provide some suggestions? Should I overwrite the current code or just add this patch?

PR #376, opened 2021-06-04
Add Fully-parameterized Quantile Function

### LINKED ISSUE #370: Implicit Quantile Network (IQN) and Fully parameterized Quantile Function (FQF)
    + [ ] exception-raising bug
    + [ ] RL algorithm bug
    + [x] new feature request
  ```python
  ```

I would like to take a stab at implementing [IQN](https://arxiv.org/pdf/1806.06923.pdf) and [FQF](https://arxiv.org/pdf/1911.02140.pdf), two distributional RL model with great results on Atari games, following the excellent reference implementations [here](https://github.com/ku2482/fqf-iqn-qrdqn.pytorch).

I am almost done with IQN and will submit a PR soon. Before I move on to FQF, I'd highly appreciate your feedback.

nuance1979:
- Add Fully-parameterized Quantile Function
- Reference: [paper](https://arxiv.org/pdf/1911.02140.pdf) and [code](https://github.com/ku2482/fqf-iqn-qrdqn.pytorch)
- Note:
  - For some reason, in my experiments, I encountered mode collapse (entropy of taus goes to zero; fraction loss skyrockets; quantile loss goes to zero; test rewards go to almost zero) much more often than both the paper and the reference implementation had suggested (they mentioned odds like 1 in 20 seeds). Therefore I set the default value of entropy regularization coefficient to 10.
  - I tried hard to figure out the reason by making my implementation as close to the reference as possible, sometimes at the expense of making the code unnecessarily complicated and the computation, wasteful. However, I still experience mode collapse quite often. Since I didn't see any gain (e.g., in training stability), I removed those modification to be more consistent with the tianshou conventions.
  - The training time is much longer than IQN, as expected.
  - I will rerun all experiments and update results after code review and revisions are done.


tianshou/policy/modelfree/fqf.py
```diff
@@ -0,0 +1,139 @@
...
+
+    .. seealso::
+
+        Please refer to :class:`~tianshou.policy.QRDQNPolicy` for more detailed
+        explanation.
+    """
+
+    def __init__(
+        self,
+        model: torch.nn.Module,
```
Trinkle23897: should we use `model: FullQuantileFunction`?

test/discrete/test_fqf.py
```diff
@@ -0,0 +1,156 @@
...
+        num_fractions=args.num_fractions, num_cosines=args.num_cosines,
+        device=args.device
+    )
+    optim = torch.optim.Adam(
+        list(net.preprocess.parameters()) + list(net.last.parameters())
+        + list(net.embed_model.parameters()), lr=args.lr
+    )
+    fraction_optim = torch.optim.RMSprop(
+        net.propose_model.parameters(), lr=args.fraction_lr
+    )
```
Trinkle23897: this part is kind of messy. I know you have tried your best to unify this API with previous IQN network but is it possible to simplify, e.g., avoid using `net.last` and `net.embed_model`?

Trinkle23897:
> * The training time is much longer than IQN, as expected.

Why? Have you aligned with that repo with all details? Because their result shows FQF/IQN/QR-DQN all have similar convergence time.

test/discrete/test_fqf.py
nuance1979: Good point. I made some changes and it looks better now.

nuance1979:
> > * The training time is much longer than IQN, as expected.
> 
> Why? Have you aligned with that repo with all details? Because their result shows FQF/IQN/QR-DQN all have similar convergence time.

I meant it was slower in terms of **wall clock time**, not in terms of training steps, which I believe is what you meant by "convergence time". The slowdown in wall clock time is simply due to the extra computation of the fraction proposal net and its gradient. I'll post some numbers and pictures once I get them.

nuance1979:
For example, a single run of Qbert: ~7h for IQN vs ~11h for FQF. Best reward: 15837.50 for IQN vs 15650 for FQF. (FQF is not a clear winner in Qbert; in some other games, it is.)

IQN reward: 
Qbert_rew

FQF reward:
Qbert_rew_fqf_new

Trinkle23897:
> I meant it was slower in terms of **wall clock time**

Oh I see, so is it possible to reduce the model forward time by reusing the previous result? The profiling result shows it takes a lot of time in the network part.
I'll try to understand the complex operation in the network and see if there's any way we can optimize it.

nuance1979:
> > I meant it was slower in terms of **wall clock time**
> 
> Oh I see, so is it possible to reduce the model forward time by reusing the previous result? The profiling result shows it takes a lot of time in the network part.
> I'll try to understand the complex operation in the network and see if there's any way we can optimize it.

Sure. I know there is a place in `_target_q()` where the same computation (one forward pass of fraction proposal net) is performed twice. I can start by optimizing it.

nuance1979:
> > > I meant it was slower in terms of **wall clock time**
> > 
> > 
> > Oh I see, so is it possible to reduce the model forward time by reusing the previous result? The profiling result shows it takes a lot of time in the network part.
> > I'll try to understand the complex operation in the network and see if there's any way we can optimize it.
> 
> Sure. I know there is a place in `_target_q()` where the same computation (one forward pass of fraction proposal net) is performed twice. I can start by optimizing it.

Wait. I take it back. It wasn't exactly the same computation because the feature net of the target model is different from that of the online model. However, since they usually do not differ much, the reference implementation does take this shortcut to save some computation. I'm running an experiment to see how much wall clock time it can save.

nuance1979:
Using the shortcut mentioned above, a single run of Qbert takes ~9.5h now. Best reward: 16172.5 . So it is faster without loss of performance. However, the code does look a bit ugly.

Screen Shot 2021-06-07 at 8 16 13 PM

tianshou/policy/modelfree/fqf.py
```diff
@@ -0,0 +1,157 @@
...
+    :param torch.optim.Optimizer optim: a torch.optim for optimizing the model.
+    :param float discount_factor: in [0, 1].
+    :param int num_fractions: the number of fractions to use. Default to 32.
+    :param float ent_coef: the coefficient for entropy loss. Default to 0.
+    :param int estimation_step: the number of steps to look ahead. Default to 1.
+    :param int target_update_freq: the target network update frequency (0 if
+        you do not use the target network).
+    :param bool reward_normalization: normalize the reward to Normal(0, 1).
+        Default to False.
+
```
Trinkle23897, line 29: what are `fraction_model` and `fraction_optim`?

Trinkle23897 (commented):
No more other comments, great job!

tianshou/policy/modelfree/fqf.py
Trinkle23897: @nuance1979 ？

nuance1979: Sorry for the late reply. Fixed.

Trinkle23897 (approved):

emrul:
Just came here to say thank you for this contribution but also thank you for explaining why the default entropy was set to 10.0 - I was specifically wondering this and found the explanation helpful!

PR #391, opened 2021-07-02
add env_id in preprocess fn

### LINKED ISSUE #381: How to do self-play correctly for tic-tac-toe?
I am trying to adapt the tic-tac-toe example to train through self-play. I train an agent against a fixed agent (with the same network architecture) until the average win rate reaches 95%. Then I copy the weights from the trained agent to the fixed agent and repeat the process for 10 generations of evolution. 
My first question is how to fix an agent under multi-agent settings. Currently I am using an optimizer with lr=0 on the agent I want to freeze as an ad hoc solution. But I am not sure it is the correct way, and there should be a solution that does not compute gradients. Should I use some API that exists (but I don't know)? Or should I formulate the fixed agent as a part of the environment and train without multi-agent settings?
My second problem is that training has no variances at all (the log at the end of each epoch says `test_reward: 1.000000 ± 0.000000`). Is it because DQN is deterministic at test time and therefore is too easy to beat? What should I change to make meaningful training progress like training against a random policy?

### LINKED ISSUE #390: `preprocess_fn` does not receive env index
    + [ ] exception-raising bug
    + [ ] RL algorithm bug
    + [x] new feature request

 I want to compute one feature per environment only at reset time and persist it. I can not do it inside `preprocess_fn` because although I know when environments are reset, I don't know which one it is and I can not save that value as an attribute. The other solution is to implement this on the environment itself or in the agent. Is this the idea?

Trinkle23897:
close #390

### DIFF
--- a/docs/tutorials/cheatsheet.rst
+++ b/docs/tutorials/cheatsheet.rst
@@ -129,7 +129,7 @@ This is related to `Issue 42 <https://github.com/thu-ml/tianshou/issues/42>`_.
 
 If you want to get log stat from data stream / pre-process batch-image / modify the reward with given env info, use ``preproces_fn`` in :class:`~tianshou.data.Collector`. This is a hook which will be called before the data adding into the buffer.
 
-It will receive with only "obs" when the collector resets the environment, and will receive five keys "obs_next", "rew", "done", "info", "policy" in a normal env step. It returns either a dict or a :class:`~tianshou.data.Batch` with the modified keys and values.
+It will receive with "obs" and "env_id" when the collector resets the environment, and will receive six keys "obs_next", "rew", "done", "info", "policy", "env_id" in a normal env step. It returns either a dict or a :class:`~tianshou.data.Batch` with the modified keys and values.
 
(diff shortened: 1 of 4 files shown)


externalsupplierstaff (changes requested):
Also change
> The "preprocess_fn" is a function called before the data has been added to the
    buffer with batch format. It will receive with only "obs" when the collector resets
    the environment, and will receive five keys "obs_next", "rew", "done", "info", and
    "policy" in a normal env step. It returns either a dict or a
    :class:`~tianshou.data.Batch` with the modified keys and values. Examples are in
    "test/base/test_collector.py".

By
>  The "preprocess_fn" is a function called before the data has been added to the
    buffer with batch format. It will receive only "obs" and "env_id" when the collector resets
    the environment, and will receive six keys "obs_next", "rew", "done", "info", "policy" and "env_id" in a normal env step. It returns either a dict or a
    :class:`~tianshou.data.Batch` with the modified keys and values. Examples are in
    "test/base/test_collector.py".

PR #438, opened 2021-09-03
Adding option for accumulating gradient

### LINKED ISSUE #437: Option to update policy with accumulated gradients
    + [ ] exception-raising bug
    + [ ] RL algorithm bug
    + [X] new feature request
  ```python
  ```
Recently I've been running sub-policies in a wrapper. To keep maximum modularity in my code and make it easier to sub for different sub-policies, the ideal behavior is that the wrapper simply calls the learn() function of the sub-policies. However, they can't be updated simultaneously, since learn() does the gradient step by default. This might be slightly problematic in use cases such as mine, where sub-policies need to be updated simultaneously. (I can still get this behavior by wrapping yet another policy class on top of the Tianshou built-in policy classes, but this losses the modularity)

I propose a very simple additional feature: User can pass flags such as "accumulate_grad=True" as keyword arguments to learn() or update() method calls, and optimizer steps are not executed if this flag is checked to be true. The user now only need to make sure all update() calls are accumulating gradient, and do a final optimizer step in their own wrapper class.

Aceticia:
    + [ ] exception-raising fix
    + [X] documentation modification
    + [X] new feature

### DIFF
--- a/tianshou/policy/base.py
+++ b/tianshou/policy/base.py
@@ -185,6 +185,11 @@ class BasePolicy(ABC, nn.Module):
     def learn(self, batch: Batch, **kwargs: Any) -> Dict[str, Any]:
         """Update policy with a given batch of data.
 
+        If you only want :meth:`learn` calls to accumulate gradients and not
+        actually update the weight, you can specify ``accumulate_grad=True``
+        in the argument. This doesn't block calls to function calls like
+        ``sync_weight``, please use with caution.
+
         :return: A dict, including the data needed to be logged (e.g., loss).
 
         .. note::
--- a/tianshou/policy/imitation/base.py
+++ b/tianshou/policy/imitation/base.py
@@ -48,7 +48,8 @@ class ImitationPolicy(BasePolicy):
         return Batch(logits=logits, act=a, state=h)
 
     def learn(self, batch: Batch, **kwargs: Any) -> Dict[str, float]:
-        self.optim.zero_grad()
+        if not kwargs.get('accumulate_grad'):
+            self.optim.zero_grad()
         if self.action_type == "continuous":  # regression
             a = self(batch).act
             a_ = to_torch(batch.act, dtype=torch.float32, device=a.device)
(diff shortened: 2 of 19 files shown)


tianshou/policy/base.py
```diff
@@ -185,6 +185,11 @@ def process_fn(
     def learn(self, batch: Batch, **kwargs: Any) -> Dict[str, Any]:
         """Update policy with a given batch of data.
 
+        If you only want :meth:`learn` calls to accumulate gradients and not
+        actually update the weight, you can specify ``accumulate_grad=True``
+        in the argument. This doesn't block calls to function calls like
+        ``sync_weight``, please use with caution.
```
Trinkle23897, line 191:
```suggestion
        ``sync_weight``, please use with caution (see issue#437).
```

Trinkle23897 (commented):
So is it possible to directly add some hooks in optimizer? Like the wrapper approach outside optim.zero_grad()and optim.step(), so that there's no need to add this kwargs everywhere...?
Or at least pass in a NULL optimizer which does nothing.

tianshou/policy/base.py
Aceticia: Directly passing a dummy optimizer is an easy way to deal with this, but it will take away the choice of "whether to accumulate grads". Can you elaborate more on this: 

> Like the wrapper approach outside optim.zero_grad()and optim.step()

Trinkle23897: > it will take away the choice of "whether to accumulate grads"

I mean you can set up your choice by passing different `optim` into `policy` class



> Can you elaborate more on this:

For example, create a new optimizer that inherit existing pytorch optimizer (e.g., Adam):

```python
class MyAwesomeOptim(torch.optim):

  def __init__(self, accumulate_grad = True, *args, **kwargs):
    super().__init__(*args, **kwargs)
    self.opt = accumulate_grad

  def zero_grad(self):
    if self.opt:
        super().zero_grad()

  def step(self):
    if self.opt:
      super().step()
```

But I think the other approach is more concise and better.

Aceticia: I agree, the optimizer idea is more concise and looks way better. I'll close this pr then?

Trinkle23897: Sure!

PR #460, opened 2021-10-05
add multi-GPU support

Trinkle23897:


Trinkle23897:
not sure what happens to GitHub CI, it doesn't trigger anything
move to #461

PR #461, opened 2021-10-05
add multi-GPU support

Trinkle23897:


Trinkle23897:
https://www.githubstatus.com/

2021-10-05 10-10-22 的屏幕截图

PR #478, opened 2021-11-02
Implements set_env_attr and get_env_attr for vector environments

### LINKED ISSUE #473: Set Attributes in Vector Environments
    + [ ] exception-raising bug
    + [ ] RL algorithm bug
    + [x] new feature request

Operating System: Ubuntu Linux; Version: Commit c19876179a52acdc4d2e7dcb10dfbb4557adca86

Hey,

I am using tianshou as a submodule in a different project. There I use `ShmemVectorEnv` to simulate an environment. During training I would like to dynamically manipulate the environments by setting certain attributes to some values. Looking at the current code, this operation does not seem to be supported by the vector environments, is that correct? Is there some reason why `BaseVectorEnv` and the workers cannot implement a `__setattr__` method? 
Is there possibly some other solution to this problem or could such methods be implemented (I'd be happy to have a crack at it and raise a pull request)? Although I have not yet performed any benchmarks, I expect the cost of instantiating a new `ShmemVectorEnv` whenever I want to modify an attribute to be prohibitive.  
Thanks for your help!

Best,
Markus

Markus28:
    + [ ] exception-raising fix
    + [x] new feature

This pull request aims to implement the features outlined in issue #473 

- Implement `get_env_attr`, `set_env_attr` for each worker class
- Implement `get_env_attr`, `set_env_attr` for vector environments
- Corresponding unit tests and documentation

### DIFF
--- a/test/throughput/test_batch_profile.py
+++ b/test/throughput/test_batch_profile.py
@@ -10,7 +10,7 @@ from tianshou.data import Batch
 
 @pytest.fixture(scope="module")
 def data():
-    print("Initialising data...")
+    print("Initializing data...")
     np.random.seed(0)
     batch_set = [
         Batch(
@@ -19,7 +19,7 @@ def data():
                 'b1': (3.14, 3.14),
                 'b2': np.arange(1e3)
             },
-            c=i
+            c=i,
         ) for i in np.arange(int(1e4))
     ]
     batch0 = Batch(
@@ -27,8 +27,8 @@ def data():
         b=Batch(
             c=np.ones((1, ), dtype=np.float64),
             d=torch.ones((3, 3, 3), dtype=torch.float32),
-            e=list(range(3))
-        )
+            e=list(range(3)),
+        ),
     )
     batchs1 = [copy.deepcopy(batch0) for _ in np.arange(1e4)]
     batchs2 = [copy.deepcopy(batch0) for _ in np.arange(1e4)]
@@ -39,13 +39,13 @@ def data():
     indexs = np.random.choice(batch_len, size=batch_len // 10, replace=False)
     slice_dict = {
         'obs': [np.arange(20) for _ in np.arange(batch_len // 10)],
-        'reward': np.arange(batch_len // 10)
+        'reward': np.arange(batch_len // 10),
     }
     dict_set = [
         {
             'obs': np.arange(20),
             'info': "this is info",
-            'reward': 0
+            'reward': 0,
         } for _ in np.arange(1e2)
     ]
     batch4 = Batch(
@@ -53,11 +53,11 @@ def data():
         b=Batch(
             c=np.ones((1, ), dtype=np.float64),
             d=torch.ones((1000, 1000), dtype=torch.float32),
-            e=np.arange(1000)
-        )
+            e=np.arange(1000),
+        ),
     )
 
-    print("Initialised")
+    print("Initialized")
     return {
         'batch_set': batch_set,
         'batch0': batch0,
@@ -67,7 +67,7 @@ def data():
         'indexs': indexs,
         'dict_set': dict_set,
         'slice_dict': slice_dict,
-        'batch4': batch4
+        'batch4': batch4,
     }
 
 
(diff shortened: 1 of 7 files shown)


Markus28:
What's up with the failed profile check? I don't see where the segmentation fault could come from :(. Apart from that, how does it look?

Trinkle23897:
not sure what's going wrong. I ran this test locally and it can successfully pass the check without segfault, maybe something went wrong in github container?

Trinkle23897:
I found the cause: it is because torch==1.10.0

test/throughput/test_batch_profile.py
```diff
@@ -106,7 +106,7 @@ def test_set_attr(data):
 
 def test_numpy_torch_convert(data):
     """Test conversion between numpy and torch."""
-    for _ in np.arange(1e5):
+    for _ in np.arange(1e4):  # not sure what's wrong in torch==1.10.0
```
Trinkle23897, line 109: here

tianshou/env/venvs.py
```diff
@@ -130,17 +130,56 @@ def __getattribute__(self, key: str) -> Any:
...
+        self,
+        key: str,
+        id: Optional[Union[int, List[int], np.ndarray]] = None
+    ) -> List[Any]:
+        """Get an attribute from the underlying environments.
+
+        If id is an int, retrieve the attribute denoted by key from the environment
+        underlying the worker at index id. The result is returned as a list with one
+        element. Otherwise, retrieve the attribute for all workers at indices id and
+        return a list that is ordered correspondingly to id.
```
Trinkle23897, line 147: should also mention if id is a list[int], same as set_env_attr

Trinkle23897 (approved):

Markus28:
Awesome! Thanks for the quick review and the polishing :+1:

ultmaster:
I have my own env worker here: https://github.com/microsoft/nni/blob/1eced0a7bff9d4de22487e7cfa45d41ce1f66605/nni/retiarii/strategy/_rl_impl.py#L25

But now it's broken because I need to implement two new methods.

I tried to update my code and delete the `__getattr__` method, but the updated version turns out to be imcompatible the tianshou 0.4.4.

How can I implement an env worker that works with both versions of tianshou? Implement both of `__getattr__` and `get_env_attr`? or use if-clause to check tianshou version?

----------------

My workaround here:

image

Trinkle23897:
how about try...except ?

ultmaster:
Not a good idea. Part of environment is user code. I don't want getattr to be run twice or catch unwanted errors.

Trinkle23897:
Also a small request: can we have a unit test for your nni integration?

ultmaster:
Sure. The related test code is here: https://github.com/microsoft/nni/blob/d50b46650bdc9e86aec8b2e0c07204c599310c9d/test/ut/retiarii/test_strategy.py#L144

I'm not sure whether it's handy right now because you need to install NNI from source (instead of from pip) for the tests.

-----

Or maybe tianshou can have a built-in support for Multi-thread env worker?

ultmaster:
I'm experiencing issues as my CI upgrading to tianshou 0.4.6 breaks my env worker again. It looks like another interface change.

```
>   self.workers = [worker_fn(fn) for fn in env_fns]
E   TypeError: Can't instantiate abstract class MultiThreadEnvWorker with abstract method send
```

I understand that EnvWorker might not be considered as a public API. But as I'm implementing a 3rd-party env worker, it becomes a burden for me to be always compatible with all kinds of versions. I've already written 3 different version of env worker for tianshou 0.4.4 and 0.4.5 and 0.4.6, and unfortunately they are not compatible with each other.

Trinkle23897:
Could you please upgrade to the newest version regardless of previous one? I'm super sorry about that happens but actually there's no CI that can secure this thing would never happen again.

Trinkle23897:
> Or maybe tianshou can have a built-in support for Multi-thread env worker?

This is in [EnvPool](https://github.com/sail-sg/envpool/)

ultmaster:
I've sent a patch. It's currently going through our CI.

microsoft/nni#4586

ultmaster:
> > Or maybe tianshou can have a built-in support for Multi-thread env worker?
> 
> This is in [EnvPool](https://github.com/sail-sg/envpool/)

Looks like yet another framework under intense development. :)

Trinkle23897:
I can make a post version tmr to fix this issue. (Also I just found another issue of consistency and it's critical to be fixed)

Trinkle23897:
fixed in #536

PR #481, opened 2021-11-18
fix tqdm issue

Trinkle23897:
#459 introduces a bug where the progress bar cannot show the last episode reward/length correctly, this PR fix that issue.

### DIFF
--- a/tianshou/trainer/offpolicy.py
+++ b/tianshou/trainer/offpolicy.py
@@ -117,8 +117,8 @@ def offpolicy_trainer(
                 env_step += int(result["n/st"])
                 t.update(result["n/st"])
                 logger.log_train_data(result, env_step)
-                last_rew = result['rew'] if 'rew' in result else last_rew
-                last_len = result['len'] if 'len' in result else last_len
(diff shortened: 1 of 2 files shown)


tianshou/trainer/offpolicy.py
```diff
@@ -117,8 +117,8 @@ def offpolicy_trainer(
                 env_step += int(result["n/st"])
                 t.update(result["n/st"])
                 logger.log_train_data(result, env_step)
-                last_rew = result['rew'] if 'rew' in result else last_rew
-                last_len = result['len'] if 'len' in result else last_len
+                last_rew = result['rew'] if result["n/ep"] > 0 else last_rew
```
thkkk, line 120: `result["n/ep"]` corresponds to episode_count in `collector.py`. If episode_count > 0, then the 'rew' and 'len' must exist.

thkkk (commented):
`result["n/ep"]` corresponds to episode_count in `collector.py`. If episode_count > 0, then the 'rew' and 'len' must exist.

thkkk (approved):

PR #485, opened 2021-11-26
Fix critic network for Discrete CRR

### LINKED ISSUE #482: Visualize result in real time
Hi, how can you see the results (such as rewards, loss, ...) of the visualization in real-time while training a model? Just like using tensorboard or Visdom. Thanks

nuance1979:
    + [ ] exception-raising fix
    + [ ] new feature

This PR:

- Fixes an inconsistency in the implementation of Discrete CRR. Now it uses `Critic` class for its critic, following conventions in other actor-critic policies;
- Updates several offline policies to use `ActorCritic` class for its optimizer to eliminate randomness caused by parameter sharing between actor and critic;
- Adds a `disable_test` option to `offline_trainer` to turn off testing during training;
- Updates the Atari offline results in README.md;
- Moves Atari offline RL examples to `examples/offline`; tests to `test/offline` per review comments.


tianshou/trainer/offline.py
```diff
@@ -75,12 +77,13 @@ def offline_trainer(
     start_time = time.time()
     test_collector.reset_stat()
 
-    test_result = test_episode(
-        policy, test_collector, test_fn, start_epoch, episode_per_test, logger,
-        gradient_step, reward_metric
-    )
-    best_epoch = start_epoch
-    best_reward, best_reward_std = test_result["rew"], test_result["rew_std"]
+    if test_in_train:
```
Trinkle23897: I don't think this is the same logic of `test_in_train` compared with other trainers. It is used for testing performance in the middle of an epoch.

Trinkle23897 (commented):
Also, could you please move the related result from `examples/atari` to `examples/offline` and move the related test to `test/discrete` to `test/offline`?

tianshou/trainer/offline.py
nuance1979: I changed the name to `disable_test`. I just need a switch to turn off testing in offline setting.

nuance1979:
> Also, could you please move the related result from `examples/atari` to `examples/offline` and move the related test to `test/discrete` to `test/offline`?

Sure. I added a few `__init__.py` files in examples directory so that I can import Atari helpers in `examples/atari` from `examples/offline`.

tianshou/trainer/offline.py
Trinkle23897: So why not use `test_collector=None` to indicate? And this can be also applied to other trainers.

Trinkle23897:
how about the other two trainers?

nuance1979:
> how about the other two trainers?

This functionality is mainly for offline RL when we don't even have an env and want to train a model; other than that, I think it's rarely necessary and/or useful to turn off testing during training.

Of course I can add it if you insist.

Trinkle23897:
Well, actually if the environment simulation is expensive and if we have already know the approximate number of steps to use, excluding the test phase can be more efficient for training an agent.

nuance1979:
> Well, actually if the environment simulation is expensive and if we have already know the approximate number of steps to use, excluding the test phase can be more efficient for training an agent.

Ok. I'll add it then.

Trinkle23897 (approved):

PR #501, opened 2022-01-11
Remove reset_buffer() from reset method

### LINKED ISSUE #500: Loading ReplayBuffer checkpoint into Collector
    + [ ] exception-raising bug
    + [ ] RL algorithm bug
    + [x] new feature request

I am using the commit 8f19a8696672fa49ead41e4e93d5d79855e5d6aa with Torch: 1.9.0+cu111 Numpy: 1.20.0 Sys: 3.8.10 (default, Sep 28 2021, 16:10:42) [GCC 9.3.0] linux 

I have a long-running job which is executed on a cluster. To avoid resource fragmentation and keep running costs low, I have to make frequent checkpoints, abort the training process, then schedule a new job to resume it. I save the model and optimizers via PyTorch. The `VectorReplayMemory` is saved via `save_hdf5`. Since `Collector` has no interface for serialization (understandably so), I tried to initialize a new `Collector` object with the deserialized replay memory. However, I learned that `Collector` resets the memory upon initialization. Thus, it is currently impossible to resume training in this fashion. I could set the `buffer` attribute manually, but this seems problematic, especially since there is a private `_assign_buffer` method, that checks certain constraints.

Is there any way around this problem? If not, is there a good reason to reset the replay memory whenever a `Collector` is initialized? What would you think about adding a keyword argument `reset_buffer=True` to `Collector.__init__`? 
I'd be happy to raise a pull request for this.

Thanks for your help!

Best,
Markus

Markus28:
    + [ ] exception-raising fix
    + [x] new feature

This pull request:
- Removes `self.reset_buffer()` from the `reset()` method
- Points out this change in the documentation

This was discussed in #500 

Best, Markus

### DIFF
--- a/tianshou/data/buffer/base.py
+++ b/tianshou/data/buffer/base.py
@@ -96,7 +96,7 @@ class ReplayBuffer:
         """Load replay buffer from HDF5 file."""
         with h5py.File(path, "r") as f:
             buf = cls.__new__(cls)
-            buf.__setstate__(from_hdf5(f, device=device))
+            buf.__setstate__(from_hdf5(f, device=device))  # type: ignore
         return buf
 
     def reset(self, keep_statistics: bool = False) -> None:
--- a/tianshou/data/collector.py
+++ b/tianshou/data/collector.py
@@ -73,7 +73,7 @@ class Collector(object):
         self.preprocess_fn = preprocess_fn
         self._action_space = env.action_space
         # avoid creating attribute outside __init__
-        self.reset()
+        self.reset(False)
 
     def _assign_buffer(self, buffer: Optional[ReplayBuffer]) -> None:
         """Check if the buffer matches the constraint."""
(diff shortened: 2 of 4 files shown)


Markus28:
So there are at least two tests (test_sac_with_il and test_collector_nepisode) that break because they assume that `reset` also resets the buffer. I can adapt those to use `collector.reset_buffer()` along with `collector.reset()`. But indeed, this seems to be a breaking change in many cases.
Not sure why the type check fails.

Trinkle23897:
> Not sure why the type check fails.

I can do for that

> What would you think about adding a keyword argument reset_buffer=True to Collector.__init__?
> So there are at least two tests (test_sac_with_il and test_collector_nepisode) that break because they assume that reset also resets the buffer. 

Indeed, but they all call `collector.reset()`, so why not change the signature to 
```python
collector.reset(reset_buffer: bool = True)
```
and in init: 
```python
def __init__(...):
  ...
  self.reset(reset_buffer=False)
```

Trinkle23897 (approved):

PR #510, opened 2022-01-18
Hindsight Experience Replay for tianshou

jc-bao:
    + [x] exception-raising fix
    + [x] new feature

* Tested on `FetchReach-v2` and `FetchPickAndPlace-v2`. See `examples/robotics/sac_her.py`
* Both online relabel and offline relabel are implemented. 
  * online-relabel relabel at `preprocess` function of `sac_her` (also tried to relabel in the `sample` function of buffer, but in preprocess function of the policy, the original buffer data rather than the sampled relabeled data is used)
  * offline-relabel relabel when storing the trajectory in `her_colletor`.
The results are as follows:
`FetchReach-v2`
Reach
`FetchPickAndPlace-v2`
PickAndPlace

TODO
- [ ] Normalization: The normalizer in the environment will make the training unstable when using relabel data. An external normalizer is needed to fix it. The result shown above is no normalization. (if open normalization, the training failed.)
- [ ] Success Rate Recorder: I found it inconvenient to record user data (like success rate in sparse reward scenario). The trainer needs to be modified. The recording problem can be solved by adding an option to record the `info` of the environment.


docs/_static/js/atari 2
```diff
@@ -0,0 +1 @@
+../../../examples/atari
```
Trinkle23897, line 1: remove

docs/_static/js/mujoco 2
```diff
@@ -0,0 +1 @@
+../../../examples/mujoco
```
Trinkle23897, line 1: remove

examples/modelbased/psrl 2.py
```diff
@@ -0,0 +1 @@
+../../test/modelbased/test_psrl.py
```
Trinkle23897, line 1: remove

examples/vizdoom/network 2.py
```diff
@@ -0,0 +1 @@
+../atari/atari_network.py
```
Trinkle23897, line 1: remove

test/throughput/env 2.py
```diff
@@ -0,0 +1 @@
+../base/env.py
```
Trinkle23897, line 1: remove

tianshou/data/her_collector.py
```diff
@@ -0,0 +1,304 @@
...
+                            trajectory.done[-1] = True
+                            for i in range(length):
+                                env_buffer.add(trajectory[i])
+                    elif self.strategy == 'online':
+                        # record the achieved goal of future steps,
+                        # to reduce the relabel time during the trainning
+                        ag = original_trajectory.obs_next[:, self.obs_index_range[
+                            'achieved_goal']]
+                        for i, idx in enumerate(obs_index_range):
+                            env_buffer.info.achieved_goal[idx] = ag[i:]
```
Trinkle23897, line 257: I don't think it is a good idea to implement HER on the collector side, instead, buffer would be the best fit. In my previous thinking, we can only call a function like `buffer.perform_her(...)` to solve this problem and it is more general and suitable to all algorithms (instead of creating something like SACHERPolicy).

jc-bao: Yes, I agree with that. Relabeling in the buffer is the most intuitive way. There are two ways of implementing HER in other repositories. Both [stable baseline3](https://github.com/DLR-RM/stable-baselines3) and [Tianhong Dai](https://github.com/TianhongDai/hindsight-experience-replay) use online sampling, which means relabeling when sampling the data. This way can generate more fresh data than the original version of HER proposed by the paper, which relabels data in the buffer then sample it as usual. This led to two different ways of implementing `perform_her()`: relabel all data in the buffer or relabel newly collected data. 
Here are my a few concerns about When to perform `buffer.perform_her(...)` and How to perform `buffer.perform_her(...)`:
1. For an offline version of HER:
* Why I choose to modify `collector`: The current version is as follows: After the collection of one trajectory is over, I repeat it for k times, relabel then store it. I have not figured out how to identify the newly collected data and deal with the partially collected trajectory. (add newly relabeled data will break them). 
* Difficulties in implementing it in `buffer`: I think we can perform the relabel per collect by calling `perform_her()`. But I have not figured out how to identify the newly collected data. Moreover, suppose we only relabel the collected data and do not generate more trajectory. In that case, it will hurt the sample efficiency as lots of original data is lost, and more goal choice is ignored.
2. For the online version of HER:
* Why I choose to modify `policy`: My initial idea was to do it in `buffer.sample()` to get relabeled data. Still, I found that the `process_fn()` function only uses the subscript obtained by sampling rather than the data, so it is hard to relabel in the buffer's sample. That's why I had the idea to modify the policy `pre_process`. So I changed the policy `pre_process` function, which can be applied to any off-policy algorithm. I have to admit that it is not a clean way to do this, but it could be general.
* Difficulties to implement it in `buffer`: The `buffer` and `policy.process_fn()` is not disentangled. I understand this design is to achieve an n-step return, but I have not figured out a better solution for the online sampling in the buffer as the sampled data is not used.

I wonder if it is possible to store alternative goals in the `info` though `buffer.perform_her()`  and sample it when updating the network. (This is what I am doing in the online HER). It is efficient(saving a lot of time to sample a new goal in the buffer during the update) and can extend to other relabel methods (e.g., inverse RL). The main obstacle is we still need to redefine the `policy.process_fn()` to recompute reward and get a new goal, as it will get data directly from the buffer. (Or to modify the `buffer.sample()` to get n_step data and decouple the policy and buffer, use the sample function as the only portal to get data). 
Anyway, tianshou shows excellent potential to perform such training task quickly, especially in relabel algorithms like HER, as most repositories use MPI to parallelize, which is resource-consuming. Hindsight Experience Replay is a crucial method to solve sparse reward tasks, and it should not be missing in such a great training framework.
Any information will be highly appreciated.

wkoa:
So, is there another way for implementing HER?

tianshou/data/her_collector.py
nuance1979: @jc-bao Is the online version almost always better than offline version? If so, I'd suggest we implement the online version only because I think modifying `collector` is much worse than modifying `policy`.

How about inheriting `buffer` and overriding `sample()` and/or related methods to perform relabeling on-the-fly?

If there is NO way to implement it as `buffer.perform_her()`, I think overriding `policy.process_fn()` is an acceptable compromise.

jc-bao: 1. Is the online version almost always better than the offline version?
Ans: Yes, the online version has 5 times higher sample efficiency compared to the original HER.
2. How about inheriting buffer and overriding sample()?
Ans: current version of tianshou will only use the index return from the sample function to sample data. The relabeled data will not be used. So it is insufficient to only modify the sample().
3. implement it as buffer.perform_her()
Ans: It is feasible for the offline version.
4. override process_fn():
Ans: [here](https://github.com/jc-bao/tianshou/blob/68f22cafc5e906d783d32964fa7f1aa91e69cf91/tianshou/policy/modelfree/sac_her.py#L92) is my solution. The problem is that it is not a general solution.

nuance1979: > 4\. Ans: [here](https://github.com/jc-bao/tianshou/blob/68f22cafc5e906d783d32964fa7f1aa91e69cf91/tianshou/policy/modelfree/sac_her.py#L92) is my solution. The problem is that it is not a general solution.

What do you mean it's not general? If you mean you want to make HER available to any off policy methods, not just SAC, you can try implementing it as a [decorator](https://en.wikipedia.org/wiki/Decorator_pattern). See an example [here](https://github.com/thu-ml/tianshou/blob/master/tianshou/policy/modelbased/icm.py) and [how it is used](https://github.com/thu-ml/tianshou/blob/master/examples/atari/atari_ppo.py#L169-L172).

jc-bao: Thanks for the information! I will try it later.

Trinkle23897:
Finished by #753

PR #514, opened 2022-01-27
Implement MBPO (#16) and REDQ

### LINKED ISSUE #16: Model-based algorithm?
    + [ ] exception-raising bug
    + [ ] RL algorithm bug
    + [x] new feature request
- [ ] I have searched through the [issue tracker] for duplicates
  ```python
  import tianshou, sys
  print(tianshou.__version__, sys.version, sys.platform)
  ```
Are you thinking about including model-based algorithms in this framwork in the future, e.g., PILCO?
If not, I wonder Is it easy to implement model-based algorithms with the current code structure.

Jimenius:
    + [ ] exception-raising fix
    + [X] algorithm implementation fix
    + [X] documentation modification
    + [X] new feature

- Both tested on `Hopper-v2`. See `examples/modelbased/mujoco_mbpo.py` and `examples/mujoco/mujoco_redq.py`.
- Results are shown below. The orange line represents `SAC` baseline. The blue line represents `MBPO` result. The red line represents `REDQ` result.
- Add ensemble network structures in `tianshou/utils/net/common.py` and `tianshou/utils/net/continuous.py`.
results


tianshou/policy/modelfree/redq.py
```diff
@@ -0,0 +1,207 @@
...
+import numpy as np
+import torch
+from torch.distributions import Independent, Normal
+
+from tianshou.data import Batch, ReplayBuffer, to_torch_as
+from tianshou.exploration import BaseNoise
+from tianshou.policy import DDPGPolicy
+
+
+class REDQPolicy(DDPGPolicy):
```
nuance1979, line 13: Why don't you inherit `SACPolicy` to match the setup in the paper as well as your MBPO implementation?

tianshou/policy/modelfree/redq.py
```diff
@@ -0,0 +1,207 @@
...
+        if self.target_mode == 'min':
+            target_q, _ = torch.min(qs, dim=0)
+        elif self.target_mode == 'mean':
+            target_q = torch.mean(qs, dim=0)
+        target_q -= self._alpha * obs_next_result.log_prob
+
+        return target_q
+
+    def learn(self, batch: Batch, **kwargs: Any) -> Dict[str, float]:
+        # critic ensemble
```
nuance1979, line 167: Is it possible to call `super().learn()` here to reduce code duplication? If so, is it possible to make `REDQPolicy` a decorator so that it can be used with any off-policy algorithm?

tianshou/policy/modelfree/redq.py
Jimenius: Inheriting `SACPolicy` requires changes of currently working code `sac.py` as the 'critics1, ...' parameters are required now. I'm not sure if this is acceptable at this moment.

The same thing happens on `EnsembleMLP` in `tianshou.utils.net.common.py`. There will be less duplicated code if I can pass the `ensemble_size` as a parameter of `EnsembleLinear` to
https://github.com/thu-ml/tianshou/blob/bc53ead273f6f9d3788a78ecc739249eeb96b8c6/tianshou/utils/net/common.py#L15-L19

tianshou/policy/modelfree/redq.py
Jimenius: It is possible if the training of actor and critic are written in helper methods. Then we can overwrite train_critic method in REDQ. As for the decorator idea, I don't have a clear thought of solution at this moment.

tianshou/policy/modelfree/redq.py
nuance1979: I see. Since `SACPolicy` is a sub-class of `DDPGPolicy`, you only go up one level. I think that's a good compromise. Thanks for the explanation!

nuance1979: Regarding `EmsembleMLP` and friends, how about something like this:

```
def miniblock(
    input_size: int,
    output_size: int = 0,
    norm_layer: Optional[ModuleType] = None,
    activation: Optional[ModuleType] = None,
    linear_layer: Type[nn.Linear] = nn.Linear,
    ensemble_size: int = 1,
) -> List[nn.Module]:
    """Construct a miniblock with given input/output-size, norm layer and \
    activation."""
    layers: List[nn.Module] = [linear_layer(input_size, output_size) if ensemble_size == 1 else linear_layer(ensemble_size, input_size, output_size)]
```

Jimenius: I was considering
```
def miniblock(
    input_size: int,
    output_size: int = 0,
    norm_layer: Optional[ModuleType] = None,
    activation: Optional[ModuleType] = None,
    linear_layer: Type[nn.Linear] = nn.Linear,
    **layer_args
) -> List[nn.Module]:
    """Construct a miniblock with given input/output-size, norm layer and \
    activation."""
    layers: List[nn.Module] = [linear_layer(input_size, output_size, **layer_args)]
```
and then change the parameter order of `EnsembleLinear`, so that it can be compatible with some other fancy layers in the future.

nuance1979: Wait. There might be simpler solutions. Keep `EnsembleLinear`, then define a lambda function to pass in to `MLP` or `miniblock()` as `linear_layer`:

```
def linear_layer(input_size, output_size):
  return EnsembleLinear(ensemble_size, input_size, output_size)
```

Similar to https://github.com/thu-ml/tianshou/blob/c25926dd8f5b6179f7f76486ee228982f48b4469/examples/atari/atari_network.py#L110-L114

Trinkle23897:
Sorry for the delay. Could you please kindly answer the following questions because I'm not quite familiar with these two algorithms and need some context?

1. What's the difference between SimpleReplayBuffer and ReplayBuffer?
2. What's the difference between RolloutCollector and Collector?
3. Why did you need to construct TransitionDataset instead of reusing ReplayBuffer?
4. Why did you put training logics under `env/`? Is there a better abstraction or location to put in?
5. What's special in dyna_trainer? (I'm not quite familiar with this algo...)
6. Can we add some extra arguments to the original MLP code to reuse most of the code in many Ensamble networks?
7. Similar issue raised from nuance, can we reuse the code as many as possible in REDQPolicy?
8. `env/mujoco/static.py` is an env-specific configuration from my perspective, maybe move to `examples/mujoco`?

Jimenius:
Thanks @Trinkle23897 and @nuance1979 for the discussions and helpful suggestions. I reimplemented MBPO in response to the questions. I'll submit another PR to have a clear view as there are lots of changes.

Currently I still can't replace SimpleReplayBuffer by VectorReplayBuffer. Take the following code as an example
```python

import time
import numpy as np
from tianshou.data import Batch, VectorReplayBuffer

if __name__ == '__main__':
    buf1 = VectorReplayBuffer(400000, 100000)
    for i in range(3):
        obs = np.ones((100000, 11)) * i
        act = np.ones((100000, 3)) * i
        rew = np.ones((100000, 1)) * i
        done = np.zeros((100000, 1), dtype=bool)
        obs_next = np.ones((100000, 11))
        batch = Batch(obs=obs, act=act, done=done, rew=rew, obs_next=obs_next)
        buf1.add(batch)
    start_time = time.time()
    buf1.sample(256)
    print(time.time() - start_time)
```
The sample call takes over 0.3s on my machine, and there are update_per_step (20) * step_per_epoch (1000) calls in an epoch, about 2 hours in total, which is not acceptable. The SimpleReplayBuffer.sample call is 20-300 times faster, but it breaks the sequential relationship of data, so the settings of algorithm and parameters are restricted.

A solution to optimize the sample method in my mind is to figure out or keep available indices of _meta in an array and sample with a single `np.random.choice` call instead of calculating the number and sampling in each sub-buffer. I'm not sure if this solution contradicts with the logic of other algorithms.

Trinkle23897:
> A solution to optimize the sample method in my mind is to figure out or keep available indices of _meta in an array and sample with a single np.random.choice call instead of calculating the number and sampling in each sub-buffer. I'm not sure if this solution contradicts with the logic of other algorithms.

what if setting a flag:
```python
if not self.full:
  return np.concatenate([sample(sub_buffer[i]) + offset[i] for i in range(self)])
else:
  return np.random.randint(len(self), size=batch_size)
```

```python
In [5]: %timeit np.random.randint(400000, size=256)
9.01 µs ± 121 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

2022-03-06 09-44-06 的屏幕截图


> sample with a single np.random.choice

That also works. From buffer's perspective:

1. At first we choose 1-d vector buffer because of vectorized prioritized replay buffer;
2. The above `flag` approach can only resolve off-policy case because on-policy needs to set `buffer.maxsize >= trajectory_len` to store the whole trajectory, and it is often the case that trajectory length is smaller than buffer size, in that case `np.random.choice` may be a good solution.
3. 2d vecbuf or linklist buffer needs lots of code refactoring.

nuance1979:
> Thanks @Trinkle23897 and @nuance1979 for the discussions and helpful suggestions. I reimplemented MBPO in response to the questions. I'll submit another PR to have a clear view as there are lots of changes.
> 

Is it possible to move REDQ into a separate PR? I think it's an independent algorithm which doesn't really share code with MBPO.

Jimenius:
>what if setting a flag

Reply to @Trinkle23897: However, in MBPO, the model buffer is often not full as the buffer size is increasing with the rollout length (usually linearly increasing.) I am thinking of keeping an "available mask" the same length as  `_meta`. The elements are set to be True when buffer.add. The sampling is gonna be like
```python
available_indices = np.arange(maxsize)[mask]
np.random.choice(indices, size=batch_size)
```
Maybe this solution also handles `stack`? This solution transers the time to the add operation, but it worths as sampling is usually a much more frequent action.
>Is it possible to move REDQ into a separate PR? I think it's an independent algorithm which doesn't really share code with MBPO.

Reply to @nuance1979: Yes, I'm doing so. The only resemblance is that they both use an ensemble network. It is usually implemented as ModuleList, but I implement it as a 3D matrix multiplication as in the original MBPO implementation, which should be faster. Based on it, I'm thinking of treating REDQ as a base algorithm and rewriting SAC and TD3 to use a critic ensemble instead of critic 1 & 2. Intrinsically, they are specific instances of REDQ.

It seems that I have a lot of work to do related to implementing the two algorithms, but currently I'm focusing my time and energy on my own research and I think it's better to keep the code stable. Feel free if you want to submit a PR with part of my code.

nuance1979:
> Based on it, I'm thinking of treating REDQ as a base algorithm and rewriting SAC and TD3 to use a critic ensemble instead of critic 1 & 2. Intrinsically, they are specific instances of REDQ.

While I understand your point, I feel it's better to leave the current implementations of SAC and TD3 unchanged. Of course it's just my personal opinion. What do you think, @Trinkle23897 ?

> It seems that I have a lot of work to do related to implementing the two algorithms, but currently I'm focusing my time and energy on my own research and I think it's better to keep the code stable. Feel free if you want to submit a PR with part of my code.

I'm also currently busy with other work. Might take it up when I have the time.

MischaPanch:
Seems like this was implemented in #623, closing this one

PR #550, opened 2022-03-02
Implement Generative Adversarial Imitation Learning (GAIL)

### LINKED ISSUE #173: Inverse reinforcement learning
    + [ ] exception-raising bug
    + [ ] RL algorithm bug
    + [x] new feature request

  ```
Do you think tianshou may support inverse reinforcement learning? 
https://ai.stanford.edu/~ang/papers/icml04-apprentice.pdf
http://martin.zinkevich.org/publications/maximummarginplanning.pdf

### LINKED ISSUE #531: Is there any bc or gail in the tianshou
    + [ ] exception-raising bug
    + [ ] RL algorithm bug
    + [x] new feature request
  ```python
  ```

nuance1979:
    + [ ] exception-raising fix
    + [x] new feature

Implement GAIL based on PPO and provide example script and sample (i.e., most likely not the best) results with Mujoco tasks. (#531, #173)


tianshou/policy/imitation/gail.py
```diff
@@ -0,0 +1,136 @@
...
+        return super().process_fn(batch, buffer, indices)
+
+    def learn(  # type: ignore
+        self, batch: Batch, batch_size: int, repeat: int, **kwargs: Any
+    ) -> Dict[str, List[float]]:
+        # update discriminator
+        losses = []
+        acc_pis = []
+        acc_exps = []
+        for b in batch.split(len(batch) // self.disc_repeat, merge_last=True):
```
Trinkle23897: is `//` correct?

tianshou/utils/net/continuous.py
```diff
@@ -471,3 +472,41 @@ def decode(
         # decode z with state!
         return self.max_action * \
             torch.tanh(self.decoder(torch.cat([state, latent_z], -1)))
+
+
+class GAILDiscriminator(nn.Module):
```
Trinkle23897: Do we really need it? `Critic(Net(...))`

examples/atari/atari_wrapper.py
```diff
@@ -181,6 +181,20 @@ def reward(self, reward):
...
+    """
+
+    def __init__(self, env):
+        super().__init__(env)
+
+    def reward(self, reward):
+        """Set reward to 0."""
+        return np.zeros_like(reward)
+
+
```
Trinkle23897: Do we really need it? I don't see any call.

tianshou/policy/imitation/gail.py
nuance1979: Good question. I started by doing the following:

```python
for _ in range(self.disc_repeat):
  for b in batch.split(batch_size, merge_last=True):
    ...
```

However, the results were terrible because the discriminator net quickly converged to almost 100% accuracy and near-zero loss, giving almost no reward signal to guide the PPO learning.

So I checked other reference implementations and "slowed down" the learning of the discriminator by reducing the number of updates, all of the sudden, the results improved dramatically. I tried two variants:

- In this version, I sample a few (between 1-4) mini batches (size: 64) from the big batch (size: 2048) and update the discriminator net:
```python
for i, b in enumerate(batch.split(batch_size, merge_last=True):
  if i > self.disc_repeat:
    break
  ...
```

- In this version, I use all the data from the big batch (size: 2048) but control the number of updates on the discriminator net:
```python
for b in batch.split(len(batch) // self.disc_repeat, merge_last=True):
  ...
```

From my experiments, the second version worked much better. So I kept it here.

Reference: In https://github.com/Khrylx/PyTorch-RL/blob/master/gail/gail_gym.py#L121-L130, it takes the whole big batch to update the discriminator just once, then go on to slice the big batch into mini batches and do PPO updates.

tianshou/utils/net/continuous.py
nuance1979: I think so because the discriminator needs to take a (state, action)-pair and output a single number. I could use a MLP but then needs to call it as `self.disc(torch.cat([obs, act], dim=1))` instead of `self.disc(obs, act)`.

In AIRL, the discriminator net will be even more complicated, like `self.disc(obs, next_obs, dones, log_pis)`. Then its almost inevitable to add a new class for it.

examples/atari/atari_wrapper.py
nuance1979: It's called in `irl_gail.py`:

```python
from examples.atari.atari_wrapper import NoRewardEnv
...
        [lambda: NoRewardEnv(gym.make(args.task)) for _ in range(args.training_num)],
        norm_obs=True
    )
```

This is to make sure the implementation doesn't accidentally use the reward signal coming from the env. I like that it gives me more confidence about the correctness.

examples/atari/README.md
```diff
@@ -121,3 +121,4 @@ One epoch here is equal to 100,000 env step, 100 epochs stand for 10M.
 | MsPacmanNoFrameskip-v4      | 1930       | ![](results/ppo/MsPacman_rew.png)     | `python3 atari_ppo.py --task "MsPacmanNoFrameskip-v4"`  |
 | SeaquestNoFrameskip-v4      | 904      | ![](results/ppo/Seaquest_rew.png)     | `python3 atari_ppo.py --task "SeaquestNoFrameskip-v4" --lr 2.5e-5`  |
 | SpaceInvadersNoFrameskip-v4 | 843       | ![](results/ppo/SpaceInvaders_rew.png) | `python3 atari_ppo.py --task "SpaceInvadersNoFrameskip-v4"`  |
+Note that CRR itself does not work well in Atari tasks but adding CQL loss/regularizer helps.
```
Trinkle23897:
```suggestion

Note that CRR itself does not work well in Atari tasks but adding CQL loss/regularizer helps.
```

And is it in the right place?

examples/atari/atari_wrapper.py
Trinkle23897: But you apply an Atari wrapper for Mujoco task? I mean it's not the right place to put this wrapper.

tianshou/policy/imitation/gail.py
Trinkle23897: I'm a little bit confused about the variable name, maybe it should change to something like `disc_update_num` or `disc_num_update_per_iter`?

tianshou/utils/net/continuous.py
Trinkle23897: > In AIRL, the discriminator net will be even more complicated, like self.disc(obs, next_obs, dones, log_pis). Then its almost inevitable to add a new class for it.

Here are my first thoughts: 
1. Is it possible to design the APIs here for AIRL?
2. Is it possible to inherit from Critic?

tianshou/policy/imitation/gail.py
nuance1979: You are right. I have changed the name of the parameter to `disc_update_num`.

tianshou/utils/net/continuous.py
nuance1979: You have a point. I refactored the code and now it didn't need a special discriminator class and it would be easy to extend to AIRL by overriding `def disc(self, batch: Batch)` method since all the information (obs, next_obs, dones, log_pis) can be passed through a `batch`.

examples/atari/atari_wrapper.py
nuance1979: I moved it to `irl_gail.py`.

examples/atari/README.md
nuance1979: Removed.

Trinkle23897 (approved):

Issue #426, opened 2021-08-24
wandb logger


drozzy:
If anyone needs a weights and biases (wandb) logger, here is one I hacked together.
Note - you have to select different step values for each of train/test/update panels in your https://wandb.ai/ panels:

```python
from tianshou.utils import BaseLogger
from tianshou.utils.log_tools import WRITE_TYPE
from typing import Any, Tuple, Union, Callable, Optional
import wandb

class WandBLogger(BaseLogger):
    def __init__(
        self,
        train_interval: int = 1000,
        test_interval: int = 1,
        update_interval: int = 1000
    ) -> None:
        super().__init__(writer=None)
        
        self.train_interval = train_interval
        self.test_interval = test_interval
        self.update_interval = update_interval
        self.last_log_train_step = -1
        self.last_log_test_step = -1
        self.last_log_update_step = -1        

    def write(self, key: str, x: int, y: WRITE_TYPE, **kwargs: Any) -> None:
        pass

    def log_train_data(self, collect_result: dict, step: int) -> None:
        if collect_result["n/ep"] > 0:
            collect_result["rew"] = collect_result["rews"].mean()
            collect_result["len"] = collect_result["lens"].mean()
            if step - self.last_log_train_step >= self.train_interval:

                log_data = {                    
                    "train/env_step": step,
                    "train/episode": collect_result["n/ep"],
                    "train/reward": collect_result["rew"],
                    "train/length": collect_result["len"]}
                wandb.log(log_data)

                self.last_log_train_step = step

    def log_test_data(self, collect_result: dict, step: int) -> None:
        assert collect_result["n/ep"] > 0
        rews, lens = collect_result["rews"], collect_result["lens"]
        rew, rew_std, len_, len_std = rews.mean(), rews.std(), lens.mean(), lens.std()
        collect_result.update(rew=rew, rew_std=rew_std, len=len_, len_std=len_std)
        if step - self.last_log_test_step >= self.test_interval:

            log_data = {                
                "test/env_step": step,
                "test/reward": rew,
                "test/length": len_,
                "test/reward_std": rew_std,
                "test/length_std": len_std}

            wandb.log(log_data)
            self.last_log_test_step = step

    def log_update_data(self, update_result: dict, step: int) -> None:        
        if step - self.last_log_update_step >= self.update_interval:
            log_data = {}
            
            for k,v in update_result.items():
                log_data[f'update/{k}'] = v

            log_data['update/gradient_step'] = step
[...]

PR #558, opened 2022-03-05
Use `global_step` as the x-axis for wandb

vwxyzjn:
    + [ ] exception-raising fix
    + [x] new feature

Tianshou already supports W&B logging via #426. The current logging solution uses two custom x-axises `train/env_step` and `test/env_step`. Such usage might be less desirable because 
1. `train/env_step` and `test/env_step` share virtually the same values, so we should use the same key such as `global_step`; with `global_step` as the x-axis we can still see the `train/reward` and `test_reward` as the y-axis, 
2. it's hard to compare tianshou's experiments with those from SB3 and CleanRL, which have adopted `global_step` as the common x-axis (see https://github.com/DLR-RM/rl-baselines3-zoo/pull/213).

To help address this issue, this PR uses `global_step` as the x-axis for wandb logging. Additionally, this PR allows the users to override the default wandb project via environment variables like:

```
WANDB_PROJECT=myproject python3 atari_dqn.py --task "BreakoutNoFrameskip-v4" --test-num 100 --logger wandb
```

## Alternatives considered

An alternative plan is to remove the `WandbLogger` altogether and instead use wandb's tensorboard integration like

```
wandb.init(..., sync_tensorboard=True)
```

While this is possible, `WandbLogger` currently does more such as resume training, so removing it is a bit more complicated.


Trinkle23897:
> WandbLogger currently does more such as resume training

TensorboardLogger also has this function. So I think that's fine to create something like

```python
class WandbLogger(TensorboardLogger):
  def __init__(self, *args, **kwargs):
    wandb.init(..., sync_tensorboard=True)
    super().__init__(*args, **kwargs)
```

vwxyzjn:
> > WandbLogger currently does more such as resume training
> 
> TensorboardLogger also has this function. So I think that's fine to create something like
> 
> ```python
> class WandbLogger(TensorboardLogger):
>   def __init__(self, *args, **kwargs):
>     wandb.init(..., sync_tensorboard=True)
>     super().__init__(*args, **kwargs)
> ```

There is an issue with this approach. The `SummaryWriter` needs to be initialized **after** the `wandb.init(..., sync_tensorboard=True)`, which requires the refactoring from the `TensorboardLogger`. Maybe we should revert the changes back?

Trinkle23897:
How about this:
```python
# logger/wandb_init.py
import wandb
wandb.init(..., sync_tensorboard=True)
```

```python
# logger/wandb.py
# from tianshou.utils.logger import wandb_init
from tianshou.utils.logger.tensorboard import TensorboardLogger

class WandbLogger(TensorboardLogger):
  pass
```

```python
# utils/__init__.py
do not import wandb_init here
```

and in main.py:
```python
...
from tianshou.utils.logger import wandb_init, WandbLogger
...
if __name__ == "__main__":
  ...
```

vwxyzjn:
This is the way CleanRL does it:

https://github.com/vwxyzjn/cleanrl/blob/0b3f8eae7d07b90a0ee129ffe290bd82e5b57a14/cleanrl/ppo.py#L136-L152

Trinkle23897:
Yeah, I mean we can replace this functionality with a simple import.

vwxyzjn:
> Yeah, I mean we can replace this functionality with a simple import.

XD didn't complete my message. I was writing but it can't be easily applied here because the logger has other utilities like save and resume data. 

How about something like

```python
    if args.logger == "wandb":
        logger = WandbLogger(
            save_interval=1,
            name=log_name,
            run_id=args.resume_id,
            config=args,
        )
    writer = SummaryWriter(log_path)
    writer.add_text("args", str(args))
    if args.logger == "wandb":
        logger.load(writer)
```

and in the `logger.load` we basically load the `TensoboardLogger`

vwxyzjn:
Per conversation with @Trinkle23897, the latest code adopts the following style.

```python
    if args.logger == "wandb":
        logger = WandbLogger(
            save_interval=1,
            name=f"{args.task}__{log_name}__{args.seed}__{int(time.time())}",
            run_id=args.resume_id,
            config=args,
        )
    writer = SummaryWriter(log_path)
    writer.add_text("args", str(args))
    if args.logger == "tensorboard":
        logger = TensorboardLogger(writer)
    if args.logger == "wandb":
        logger.load(writer)
```


https://wandb.ai/costa-huang/tianshou/runs/uktkei7h?workspace=user-costa-huang tracks this run.

Trinkle23897 (dismissed):

examples/atari/atari_ppo.py
```diff
@@ -184,37 +186,44 @@ def dist(p):
...
-    log_name = 'ppo_icm' if args.icm_lr_scale > 0 else 'ppo'
-    log_path = os.path.join(args.logdir, args.task, log_name)
-    if args.logger == "tensorboard":
-        writer = SummaryWriter(log_path)
-        writer.add_text("args", str(args))
-        logger = TensorboardLogger(writer)
-    else:
+    now = datetime.datetime.now().strftime("%y%m%d-%H%M%S")
+    algo_name = "ppo_icm" if args.icm_lr_scale > 0 else "ppo"
+    log_name = os.path.join(args.task, algo_name, str(args.seed), now)
```
vwxyzjn: Could you add the `algo_name` to `args` so that we can filter all experiments whose `algo_name="dqn"` or group by `algo_name`?

vwxyzjn (commented):
Everything looks good except I think we should add `algo_name` to the args variable. Also, do you have a tracked run?

examples/atari/atari_ppo.py
Trinkle23897: https://wandb.ai/tianshou/atari.benchmark/runs/mh6zb1zf?workspace=user-tianshou

vwxyzjn:
image

It works now.

vwxyzjn:
All good on my end

Trinkle23897 (approved):

PR #613, opened 2022-04-27
Upgrade gym

### LINKED ISSUE #605: Setting seed, return_info, options for reset
    + [ ] exception-raising bug
    + [ ] RL algorithm bug
    + [x] new feature request
  ```python
  ```

Gym environment API supports seed, return_info, and options in reset:

https://github.com/openai/gym/blob/dcae553204957a75db5e4169e33547ccac599609/gym/core.py#L93

But I find the information lost in tianshou:

https://github.com/thu-ml/tianshou/blob/e01385ea30d21b7efc3f9377d22d83e6afd39e2f/tianshou/env/worker/subproc.py#L90
https://github.com/thu-ml/tianshou/blob/e01385ea30d21b7efc3f9377d22d83e6afd39e2f/tianshou/env/worker/dummy.py#L34

Also the return type is `Union[ObsType, tuple[ObsType, dict]]` in gym, but only `ObsType` in tianshou.

ycheng517:
    + [ ] exception-raising fix
    + [ ] new feature
---
fixes some deprecation warnings due to new changes in gym version 0.23:
- use `env.np_random.integers` instead of `env.np_random.randint`
- support `seed` and `return_info` arguments for reset (addresses https://github.com/thu-ml/tianshou/issues/605)


Trinkle23897:
I'll take a look later today. Thanks for the contribution anyway!

cc @ultmaster

test/modelbased/test_ppo_icm.py
```diff
@@ -75,7 +75,7 @@ def test_ppo(args=get_args()):
     args.state_shape = env.observation_space.shape or env.observation_space.n
     args.action_shape = env.action_space.shape or env.action_space.n
     if args.reward_threshold is None:
-        default_reward_threshold = {"CartPole-v0": 195}
+        default_reward_threshold = {"CartPole-v1": 195}
```
Trinkle23897: Please change the corresponding reward threshold together. CartPole-v1 is 475 instead of 195
Also, gym doesn't remove CartPole-v0 because they are with different config. So I personally do not recommend changing v0 to v1.

setup.py
```diff
@@ -15,7 +15,7 @@ def get_version() -> str:
 
 def get_install_requires() -> str:
     return [
-        "gym>=0.15.4",
+        "gym>=0.23.1",
```
Trinkle23897, line 18: I personally suggest supporting both for compatibility. Otherwise this pull request must be blocked.

tianshou/env/pettingzoo_env.py
```diff
@@ -55,7 +55,9 @@ def __init__(self, env: BaseWrapper):
 
         self.reset()
 
-    def reset(self) -> dict:
+    def reset(self, seed: Optional[int] = None) -> dict:
+        if seed is not None:
+            self.env.seed(seed)
```
Trinkle23897: Has PettingZoo upgraded their API in the newest release?

Trinkle23897: Also, you didn't change the reset's return value to `Union[obs, Tuple[obs, info]]`. That's a big issue.

tianshou/env/worker/subproc.py
```diff
@@ -104,7 +104,7 @@ def _encode_obs(
             elif cmd == "render":
                 p.send(env.render(**data) if hasattr(env, "render") else None)
             elif cmd == "seed":
-                p.send(env.seed(data) if hasattr(env, "seed") else None)
+                env.reset(seed=data)
```
ultmaster: Why not support arbitrary `*args, **kwargs` at L90?

tianshou/env/pettingzoo_env.py
ycheng517: PettingZoo has not upgrade their API yet however this is will be done in the next release. Changing the return value to `Union[obs, Tuple[obs, info]]` everywhere in this repo is part of the additional work that needs to be done at some point. I can include it in this PR if desired.

test/modelbased/test_ppo_icm.py
ycheng517: Thanks for the input. I can revert back to v0 to keep the configs the same as before

setup.py
jkterry1: I don't think this quite makes sense? Right now, there have been no large breaking changes made to 0.23.1, all the major stuff is new APIs. All the other RL libraries (RLlib, SB3, etc.) have or are about to use a minimum version of the current version of Gym leading up to the 1.0 release. Environments have largely been cleaned up to this end as well.

jkterry1: I don't think this quite makes sense? Right now, there have been no large breaking changes made to 0.23.1, all the major stuff is new APIs. All the other RL libraries (RLlib, SB3, etc.) have or are about to use a minimum version of the current version of Gym leading up to the 1.0 release. Environments have largely been cleaned up to this end as well, and virtually no reasonable environment code has been broken thus far.

tianshou/env/worker/subproc.py
ycheng517: good suggestion, will support that as well.

tianshou/env/pettingzoo_env.py
Trinkle23897: > Changing the return value to Union[obs, Tuple[obs, info]] everywhere in this repo is part of the additional work that needs to be done at some point. I can include it in this PR if desired.

I fix this in the last PR (#623) since PettingZoo==0.18.1 introduces this API.

setup.py
ycheng517: this PR should now support both, although it would still be nice to bump the minimum version, as dependencies envpool and pettingzoo requires `>=0.18` and `>=0.21` respectively.

test/modelbased/test_ppo_icm.py
ycheng517: reverted back to v0

tianshou/env/pettingzoo_env.py
ycheng517: updated return value to `Union[obs, Tuple[obs, info]]`

ycheng517:
@Trinkle23897 can you please take another look at this PR? I've implemented your suggested changes and also made `reset` support `return_info` and `seed` in the venvs.

tianshou/env/venvs.py
```diff
@@ -195,15 +196,26 @@ def reset(
...
         for i in id:
-            self.workers[i].send(None)
-        obs_list = [self.workers[i].recv() for i in id]
+            self.workers[i].send(None, **kwargs)
+        ret_list = [self.workers[i].recv() for i in id]
+
+        if "return_info" in kwargs and kwargs["return_info"]:
+            obs_list = [r[0] for r in ret_list]
+        else:
+            obs_list = ret_list
```
Trinkle23897: Is it possible to check it by type? Or do you have any other better approach?
```python
if isinstance(ret_list[0], (tuple, list)) and len(ret_list[0]) == 2 and isinstance(ret_list[0][1], dict):
  # return obs, info
# return obs
```

I know it's a little bit confusing since the observation may also be a tuple. However, I personally don't like this gym's API. Usually the user won't change the environment return data type during the whole process, so the `return_info` argument should be placed in `__init__` instead of `reset` function.

From this point of view, envpool uses `gym_reset_return_info` option in initialization. However, this method cannot support envpool. If possible, would you please add a test case for envpool? Though it only affects the venv wrapper part.

P.S. gym will finally remove this argument in reset function.

Trinkle23897: Also, we don't support tuple observation space and recommend using dict space. This is quite a strong assumption you can use here. If anyone uses tuple observation and meet exception, we can raise the corresponding hint for them by saying something like `please change your observation space from tuple to array or dict space`

ycheng517: I think checking by type is a workable approach, adopted it in 364b46e249f70b21dd8c3b46dcdcb24fe1162b88 . 
Also added test for envpool, and added exception for using tuple observation in the same commit

tianshou/env/venvs.py
```diff
@@ -202,16 +202,25 @@ def reset(
             self.workers[i].send(None, **kwargs)
         ret_list = [self.workers[i].recv() for i in id]
 
-        if "return_info" in kwargs and kwargs["return_info"]:
+        has_infos = isinstance(ret_list[0], (tuple, list)) and len(
+            ret_list[0]
+        ) == 2 and isinstance(ret_list[0][1], dict)
+        if has_infos:
```
Trinkle23897: Can we use `self.has_infos` to reduce checking overhead for all classes?
```python
if self.has_info is None:
  self.has_info = isinstance(ret_list[0], (tuple, list)) and len(
            ret_list[0]
        ) == 2 and isinstance(ret_list[0][1], dict)
if self.has_info:
  ...
else:
  ...
```

ycheng517: implemented in https://github.com/thu-ml/tianshou/pull/613/commits/662a68a05363db391b2a50717010f41ae10d7f1f. (although I did `if hasattr(self, "reset_returns_info")`, as mypy complains about typing in a bunch of places if I do `self.has_info is None`). A warning that with this change, users may experience a surprise if they call `reset` with `return_info` set to `True` and then `False`, but maybe this isn't likely to happen.

Trinkle23897: Have you ever tested with a full training pipeline?

There are some options:
- venv.reset only returns obs, no change to collector
- vecv.reset only returns (obs, info), change collector to adapt
- venv.reset can return either (obs) or (obs, info), collector needs to handle both cases

I personally favor the second approach.

ycheng517: > Have you ever tested with a full training pipeline?

You mean run it with one of the scripts in the examples folder? Haven't done that yet but can do.

I'm not sure if the second approach would work in all cases, since there could be some environments that don't return info along with obs right now. It would also be jumping ahead of the planned Gym API changes. If the 3rd option also sound good with you, I can implement that.

Trinkle23897: I mean if there's no info, we attach an empty dict. This can save a lot of code compared with 3.

Trinkle23897: > If the 3rd option also sound good with you, I can implement that.

I'm ok with it.

ycheng517: added support to collector for option 3 in https://github.com/thu-ml/tianshou/pull/613/commits/75ecd180f7f38076ad10e36301b367b2cf63268f

Also I ran `python examples/box2d/lunarlander_dqn.py` and it works fine

tianshou/data/collector.py
```diff
@@ -126,10 +131,20 @@ def reset_buffer(self, keep_statistics: bool = False) -> None:
 
     def reset_env(self) -> None:
         """Reset all of the environments."""
-        obs = self.env.reset()
-        if self.preprocess_fn:
-            obs = self.preprocess_fn(obs=obs,
-                                     env_id=np.arange(self.env_num)).get("obs", obs)
+        if self.gym_reset_return_info:
+            obs, info = self.env.reset(return_info=True)
```
Trinkle23897: I think you cannot do this, at least envpool will fail

tianshou/data/collector.py
```diff
@@ -143,6 +158,26 @@ def _reset_state(self, id: Union[int, List[int]]) -> None:
             elif isinstance(state, Batch):
                 state.empty_(id)
 
+    def _reset_env_with_ids(
+        self, local_ids: Union[List[int], np.ndarray], global_ids: Union[List[int],
```
Trinkle23897: Could you please add a test to check the correctness of info storage order?

tianshou/data/collector.py
```diff
@@ -34,13 +34,16 @@ class Collector(object):
         with corresponding policy's exploration noise. If so, "policy.
         exploration_noise(act, batch)" will be called automatically to add the
         exploration noise into action. Default to False.
+    :param bool gym_reset_return_info: if set to True, return the info dict when
```
Trinkle23897: Can we do the same things for collector? I mean dynamic checking instead of using user-defined flags (because the user's assumption is, `collector(policy, env, buffer)` will work in any way, no need to pass in other args in most use cases). If so, it will pop up tons of open issues when gym release 1.0, shift to return `(obs, info)` by default and delete the `return_info` flag.

ycheng517: In other places, doing dynamic checking made sense because we allow users to specify arbitrary kwargs when calling `reset`, so users can specify `return_info=True`. For `Collector.collect`, what do you think is the best way for users to specify `return_info`? I'm thinking we can have `gym_reset_kwargs` as an input arg in `Collector.collect` and `Collector.reset`. Then we can dynamically check the output.

tianshou/data/collector.py
ycheng517: can do

tianshou/data/collector.py
ycheng517: see my other comment just now https://github.com/thu-ml/tianshou/pull/613#discussion_r886324409 . With my proposal we shouldn't run into this issue. Open to other ideas as well.

tianshou/data/collector.py
Trinkle23897: I mean you can add an end-to-end training test with `envpool.make_gym(gym_reset_return_info=True)`...

ycheng517: supporting `envpool.make_gym(gym_reset_return_info=True)` through dynamic checking sounds good and I can also add an end-to-end test. My question was more geared towards supporting `return_info` for non-envpool environments where you pass this argument in through the reset function, should the collector support this? (i.e. `collector.collect(gym_reset_kwargs={return_info=True})`)

Trinkle23897: Sure, I think we should do that.

ycheng517: done in https://github.com/thu-ml/tianshou/pull/613/commits/c2ff71fb5ae1a5dae99dd677e3e72a274b28566a
Although one issue I ran into is that I had to do away with your earlier [suggestion](https://github.com/thu-ml/tianshou/pull/613#discussion_r878699042) of caching the `has_info` result, because [collector's __init__ function](https://github.com/thu-ml/tianshou/blob/master/tianshou/data/collector.py#L77) calls `self.reset` without any kwargs, so we end up always caching "not returning info". Wondering if you have any better ideas. I think maybe it's ok to check on every reset since it only takes a couple of microseconds on my computer and after gym version 0.5, the API is changing to always returning info.

ycheng517: Instead of an end-to-end test with `envpool.make_gym(gym_reset_return_info=True)`, I added a collector unit test that uses envpool, see: `test_collector_envpool_gym_reset_return_info`

tianshou/data/collector.py
ycheng517: added some checks to info storage in `test_collector` in https://github.com/thu-ml/tianshou/pull/613/commits/c2ff71fb5ae1a5dae99dd677e3e72a274b28566a , hopefully that's what you were looking for.

tianshou/data/collector.py
ycheng517: should be fixed now

tianshou/data/collector.py
Trinkle23897: > Wondering if you have any better ideas.

I think the current solution is fine, but if I were you, again I'll implement option 2 in https://github.com/thu-ml/tianshou/pull/613#discussion_r878787085 because that's cleaner.

ycheng517: One other problem that I forsee with option 2 is: envpool and gym environments return (obs, info) through different ways (envpool through `gym_reset_return_info` when initializing the env, and gym through `return_info` when calling reset), so the venvs needs to be smart about whether `return_info` can be included as a kwarg when calling `reset`. we might have to do something like:
```
if not hasattr(self.env, `gym_reset_return_info`):
    obs, info = self.env.reset(return_info=True, **kwargs)
elif not self.env.gym_reset_return_info:
    raise ValueError("gym_reset_return_info must be True for envpool environments")
else:
    obs, info = self.env.reset(**kwargs)
```
if you think this is still the way to go, I can implement it instead.

ycheng517: And let me know if there are other issues or if we are good to merge

Trinkle23897: Just returned from vacation. I'll review it as soon as possible. Sorry for the delay.

tianshou/data/collector.py
```diff
@@ -143,13 +164,41 @@ def _reset_state(self, id: Union[int, List[int]]) -> None:
...
+        self.data.obs_next[local_ids] = obs_reset
+
     def collect(
         self,
         n_step: Optional[int] = None,
         n_episode: Optional[int] = None,
         random: bool = False,
         render: Optional[float] = None,
         no_grad: bool = True,
+        gym_reset_kwargs: Optional[Dict[str, Any]] = None,
```
ultmaster, line 201: Any chance this can be a callable? Sometimes I want different reset kwargs for different environments, e.g., different seeds.

tianshou/env/worker/subproc.py
```diff
@@ -86,25 +87,40 @@ def _encode_obs(
...
-                if data is None:  # reset
-                    obs = env.reset()
+                obs, reward, done, info = env.step(data)
+                if obs_bufs is not None:
+                    _encode_obs(obs, obs_bufs)
+                    obs = None
+                p.send((obs, reward, done, info))
+            elif cmd == "reset":
+                retval = env.reset(**data)
+                if reset_returns_info is None:
```
ultmaster: This seems a little ad-hoc. Can we do this:

- If a tuple, encode the first element as obs, and keep the rest as is.
- Otherwise, encode the whole object as obs.

tianshou/data/collector.py
ycheng517: I'd like to just get this PR wrapped up, and introduce additional functionality in future PRs

tianshou/data/collector.py
ycheng517: @Trinkle23897 can you give this PR another review?

tianshou/env/worker/subproc.py
ycheng517: addressed in https://github.com/thu-ml/tianshou/pull/613/commits/edfcbcf3017d71132f235c0a5a4681811d7eeb6b. Earlier it was suggested to cache the result of `reset_returns_info`, but due to other difficulties we didn't end up doing it in other places.

Trinkle23897 (approved):
LGTM

Trinkle23897:
BTW it's better to update the docs to show how to use this `return_info` feature, but we can do it in the following PR.

PR #494, opened 2021-12-16
Pettingzoo

mahi97:
- [X] I have marked all applicable categories:
    + [X] exception-raising fix
    + [ ] algorithm implementation fix
    + [X] documentation modification
    + [X] new feature
- [X] I have reformatted the code using `make format` (**required**)
- [X] I have checked the code using `make commit-checks` (**required**)
- [X] If applicable, I have mentioned the relevant/related issue(s)
- [X]If applicable, I have listed every items in this Pull Request below

Tested with tic-tac-toe, piston ball (discrete and continuous)

Issue #399, opened 2021-07-14
MultiAgentPolicyManager misses rewards


benblack769:
- [ ] I have marked all applicable categories:
    + [ ] exception-raising bug
    + [x] RL algorithm bug
    + [ ] documentation request (i.e. "X is missing from the documentation.")
    + [ ] new feature request
- [x] I have visited the [source website](https://github.com/thu-ml/tianshou/)
- [x] I have searched through the [issue tracker](https://github.com/thu-ml/tianshou/issues) for duplicates
- [x] I have mentioned version numbers, operating system and environment, where applicable:

So Tianshou's multi-agent policy manager has an issue where it ignores certain rewards coming in from the environment. 

In particular, the only reward an agent sees is the reward it receives immediately after taking an action. It does not notice reward it receives when an opponent takes an action. 
To demonstrate this, consider the following extremely simple environment. 

Agent's action should be their opponent's action + 5. If they choose the correct action, they are rewarded +1. If not, then +0. This is a trivial environment to learn, and tianshou's multi-agent policy manager learns it successfully. 

But what if the reward is delayed? In the ordinary case, where the reward is not delayed, you would expect that

start action = 0

agent | action | reward_1 | reward_2
---   | -----  | -------- | --------
1     | 2      | 0        | 0
2     | 7      | 0        | 1
1     | 12     | 1        | 0
2     | 1      | 0        | 1

And in this setup, the agent learns fine.

However, it should be perfectly valid to delay the reward one turn:

agent | action | reward_1 | reward_2
---   | -----  | -------- | --------
1     | 2      | 0        | 0
2     | 7      | 0        | 0
1     | 12     | 0        | 1
2     | 1      | 1        | 0

And the agent should still learn. However, Tianshou's MultiAgentPolicyManager
is unable to learn this environment at all. The reason is that it only considers rewards on the same step the agent takes. 

Fully reproducible example here: https://gist.github.com/benblack769/80fe3ea5637108bf4e63d94de53e28b1

The problematic bit of code is this line here: https://github.com/thu-ml/tianshou/blob/master/tianshou/policy/multiagent/mapolicy.py#L121
Note that this line only fetches a subset of the rewards returned by the environment. 

Note that the tictactoe example embedded in tianshou also suffers from this problem, as only the last player is rewarded, the player who did not step during the last turn will not receive any reward. 

[...]

PR #619, opened 2022-04-28
Support Basic MARL

xihuai18:
    + [ ] exception-raising fix
    + [x] new feature

As I mentioned in #494, to support CTDE scheme or other schemes that need global information,  the returned `info` should contain more information.

By the way, I am trying to implement some basic MARL-communication algorithms based on tianshou. Inspired by https://github.com/thu-ml/tianshou/issues/399#issuecomment-881993755, I have the following design:
- [x] An multi-agent venv have `agent_num` agents and `env_num` envs should act as a venv with `agent_num` x `env_num` envs. `env_id` = `agent_num` x `env_num` + `env_num` should be contained in `info`.
- [x] Similar to AysncCollector, the `env_id` in `info` indicates which agents are taking actions.
- [ ] The MABuffer has `agent_num` x `env_num` buffers, the indices used for each agent should be the same.
- [ ] The MApolicy contains `agent_num` agents and the centralized module.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -46,6 +46,7 @@ def get_extras_require() -> str:
             "doc8",
             "scipy",
             "pillow",
+            "supersuit",  # for padding pettingzoo
             "pettingzoo>=1.17",
             "pygame>=2.1.0",  # pettingzoo test cases pistonball
             "pymunk>=6.2.1",  # pettingzoo test cases pistonball
--- a/tianshou/env/pettingzoo_env.py
+++ b/tianshou/env/pettingzoo_env.py
@@ -2,6 +2,7 @@ from abc import ABC
 from typing import Any, Dict, List, Tuple
 
 import gym.spaces
+import supersuit as ss
 from pettingzoo.utils.env import AECEnv
 from pettingzoo.utils.wrappers import BaseWrapper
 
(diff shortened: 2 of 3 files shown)


Trinkle23897:
>  An multi-agent venv have agent_num agents and env_num envs should act as a venv with agent_num x env_num envs. env_id = agent_num x env_num + env_num should be contained in info.

This is not a good assumption. The environment's input per step can be (action, player_id) and many players share the same environment. My previous solution to this issue is to extend the API from action array to action dict, for example, https://github.com/sail-sg/envpool/blob/c2b5b6c679304c976d821e2519e17cf3e9f4f98e/envpool/python/envpool.py#L101-L109
where action format is (4 envs, 8 actions)
```python
{
  "env_id": [0, 1, 2, 3],
  "players.env_id": [1, 2, 0, 2, 0, 1, 1, 2]
  "players.action: [x, x, x, x, x, x, x, x],
  "players.id": [a, b, c, a, b, c, d, e],
}
```

It shows 4 environments, 5 agents (a-e), and generates a batch of data with size 8. Policy a generates 0th and 3rd action for 1st and 2nd environment respectively.

The full test script is https://github.com/sail-sg/envpool/blob/c2b5b6c679304c976d821e2519e17cf3e9f4f98e/envpool/dummy/dummy_envpool_test.cc#L26-L92

The above is the solution provided by envpool. And It may not be consistent with PettingZoo. I put it here just for a reference.

The other 3 items look good!

PR #621, opened 2022-04-28
Convert RL Unplugged Atari datasets to tianshou ReplayBuffer

### LINKED ISSUE #612: Improve discrete control offline RL benchmark
    + [ ] exception-raising bug
    + [ ] RL algorithm bug
    + [x] new feature request
  ```python
  ```

I wonder if anyone is actively working on improving the benchmark on discrete control offline RL policies. As I noted in `examples/offline/README.md`, we should use a publicly available dataset to benchmark our policies.

Currently the best discrete control offline datasets seem to be the Atari portion of [RL Unplugged](https://github.com/deepmind/deepmind-research/tree/master/rl_unplugged). I tried to convert them into Tianshou `Batch` but couldn't find out how to get the `done` flag.

https://github.com/deepmind/deepmind-research/blob/1642ae3499c8d1135ec6fe620a68911091dd25ef/rl_unplugged/atari.py#L26-L37

If I assume the data points are in order, then I might be able to find the point where the next episode id is not the same as the current episode id and mark it to be episode end. But I don't know if this assumption holds.

@Trinkle23897 What do you think? Have you worked with [Reverb](https://github.com/deepmind/reverb) data before?

nuance1979:
    + [ ] exception-raising fix
    + [x] new feature

Add a script to convert RL Unplugged Atari datasets to tianshou ReplayBuffer for better benchmarks. #612

### DIFF
--- a/examples/offline/convert_rl_unplugged_atari.py
+++ b/examples/offline/convert_rl_unplugged_atari.py
@@ -1,0 +1,277 @@
+#!/usr/bin/env python3
+#
+# Adapted from https://github.com/deepmind/deepmind-research/blob/master/rl_unplugged/atari.py
+#
+"""Convert Atari RL Unplugged datasets to Tianshou replay buffers.
+
+Examples in the dataset represent SARSA transitions stored during a
+DQN training run as described in https://arxiv.org/pdf/1907.04543.
+
+For every training run we have recorded all 50 million transitions corresponding
+to 200 million environment steps (4x factor because of frame skipping). There
+are 5 separate datasets for each of the 45 games.
+
+Every transition in the dataset is a tuple containing the following features:
+
+* o_t: Observation at time t. Observations have been processed using the
+    canonical Atari frame processing, including 4x frame stacking. The shape
+    of a single observation is [84, 84, 4].
+* a_t: Action taken at time t.
+* r_t: Reward after a_t.
+* d_t: Discount after a_t.
+* o_tp1: Observation at time t+1.
+* a_tp1: Action at time t+1.
+* extras:
+  * episode_id: Episode identifier.
+  * episode_return: Total episode return computed using per-step [-1, 1]
+      clipping.
+"""
+import os
+from argparse import ArgumentParser
+
+import requests
+import tensorflow as tf
+from tqdm import tqdm
+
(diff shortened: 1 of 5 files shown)


tianshou/data/buffer/base.py
```diff
@@ -86,10 +86,10 @@ def __setattr__(self, key: str, value: Any) -> None:
                 ), "key '{}' is reserved and cannot be assigned".format(key)
         super().__setattr__(key, value)
 
-    def save_hdf5(self, path: str) -> None:
+    def save_hdf5(self, path: str, compression: Optional[str] = None) -> None:
         """Save replay buffer to HDF5 file."""
         with h5py.File(path, "w") as f:
-            to_hdf5(self.__dict__, f)
+            to_hdf5(self.__dict__, f, compression=compression)
```
Trinkle23897, line 92: How about from_hdf5? Does it also need compression to decode?

nuance1979: It will decompress automatically so no need to specify.

Trinkle23897:
Also, could you please write something in readme to demonstrate the step-by-step conversion process of rl-unplugged dataset?

Trinkle23897 (approved):

examples/offline/convert_rl_unplugged_atari.py
```diff
@@ -0,0 +1,278 @@
...
+
+def process_shard(url: str, fname: str, ofname: str) -> None:
+    download(url, fname)
+    file_ds = tf.data.TFRecordDataset(fname, compression_type="GZIP")
+    buffer = ReplayBuffer(500000)
+    cnt = 0
+    for example in file_ds:
+        batch = _tf_example_to_tianshou_batch(example)
+        buffer.add(batch)
+        cnt += 1
```
Trinkle23897, line 204: It's better to use a batch-style process instead of a for-loop, e.g., directly set buffer._meta.obs = ... (or maybe we should create another API to support this thing?)

For example,
```python
buffer = ReplayBuffer(500000)
# add first batch to initialize memory
batch = ...
buffer.add(batch)
# then directly set, I know it's ugly but this is the general idea
meta = buffer._meta
meta.obs = ...
meta.act = ...
...
buffer._meta = meta
```

PR #633, opened 2022-05-06
fixed problem with tuple obs being confused with return tuple

guillefix:
    + [x] exception-raising fix
    + [ ] new feature

Envs with tuple observations failed when reset because the code thought that reset was returning a tuple of obs, return, done, info, so I added a better type check. I also made the continuous Critic work with structured obs (like tuples)

### DIFF
--- a/tianshou/utils/net/continuous.py
+++ b/tianshou/utils/net/continuous.py
@@ -100,6 +100,8 @@ class Critic(nn.Module):
         preprocess_net_output_dim: Optional[int] = None,
         linear_layer: Type[nn.Linear] = nn.Linear,
         flatten_input: bool = True,
+        flatten_obs: bool = True,
+        convert_to_torch: bool = True,
     ) -> None:
         super().__init__()
         self.device = device
@@ -114,6 +116,8 @@ class Critic(nn.Module):
             linear_layer=linear_layer,
             flatten_input=flatten_input,
         )
+        self.flatten_obs = flatten_obs
+        self.convert_to_torch = convert_to_torch
 
     def forward(
         self,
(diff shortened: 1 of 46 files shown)


Trinkle23897:
Have you read https://github.com/thu-ml/tianshou/pull/147#issuecomment-660956151?
Tuple obs space is not supported because of our design choice. It will lead to many undefined behaviors in `Batch`, and further slow down the entire pipeline (because of the memory layout of np.array with dtype=object). We suggest using either vectorized env wrapper or `collector.preprocess_fn` to convert tuple space to dict space.

guillefix:
Ah, I hadn't seen that!
Oki, I will change to my environment to a Dict obs wrapped one.

I think the allowing the Critic to have structured obs (Dict) is still a good thing though no?

Trinkle23897:
> I think the allowing the Critic to have structured obs (Dict) is still a good thing though no?

Yes, but we cannot assume which key the users want. Thus the network module is under `tianshou.utils.net` instead of `tianshou.net` to encourage the users to overwrite the forward function to fulfill their requirements.

So maybe you can change this PR to a documentation update to make things more clear for the incoming users?

guillefix:
What I did is to have an option in the Critic network that just passes the obs as is to the preprocess_net, like the Actor network does as well, allowing the preprocess_net to be custom and deal with the Dict obs as desired by the user

Trinkle23897:
I wrote something here: https://tianshou.readthedocs.io/en/master/tutorials/cheatsheet.html#user-defined-environment-and-different-state-representation

tianshou/utils/net/continuous.py
```diff
@@ -122,18 +126,24 @@ def forward(
...
-                act,
+        if self.convert_to_torch:
+            obs = torch.as_tensor(
+                obs,
                 device=self.device,  # type: ignore
                 dtype=torch.float32,
-            ).flatten(1)
-            obs = torch.cat([obs, act], dim=1)
+            )
+            if self.flatten_obs: obs = obs.flatten(1)
```
Trinkle23897, line 135: If the `obs` from env is a dict, its type here is `Batch` and there's no `flatten` method.

MischaPanch:
I think it's better to have as simple interfaces as possible. Gymnasium envs return `obs, info` on reset, and I don't think other signatures should be part of tianshou. Adding more input options to the source code (like Critic) also creates additional complexity, which can rather easily be delegated to user code (by preparing the env correctly). Can this PR be closed? @Trinkle23897 @guillefix

MischaPanch:
Closing due to inactivity

PR #641, opened 2022-05-17
Show progress

michalgregor:
Hi, I added support for disabling the tqdm progress bar in the base trainer by replacing it with a dummy class that just keeps track of the total and n stats, but doesn't display progress.

I wonder whether you would like to merge this.

* A dummy_tqdm class added to utils: it replicates the interface used
  by trainers, but does not show the progress bar;
* Added a show_progress argument to the base trainer: when
  show_progress == True,  dummy_tqdm is used in place of tqdm;

Btw. I wonder whether requiring `make format` and then not actually doing it on every commit  is a good idea – I think it is going to make the changes I made more difficult to find – but at least I did that step in a separate commit, so that should help.

    + [ ] exception-raising fix
    + [X] new feature

### DIFF
--- a/examples/atari/atari_c51.py
+++ b/examples/atari/atari_c51.py
@@ -5,10 +5,10 @@ import pprint
 
 import numpy as np
 import torch
-from atari_network import C51
-from atari_wrapper import make_atari_env
 from torch.utils.tensorboard import SummaryWriter
 
+from atari_network import C51
+from atari_wrapper import make_atari_env
 from tianshou.data import Collector, VectorReplayBuffer
 from tianshou.policy import C51Policy
 from tianshou.trainer import offpolicy_trainer
--- a/examples/atari/atari_c51.py
+++ b/examples/atari/atari_c51.py
@@ -5,10 +5,10 @@ import pprint
 
 import numpy as np
 import torch
-from torch.utils.tensorboard import SummaryWriter
-
 from atari_network import C51
 from atari_wrapper import make_atari_env
+from torch.utils.tensorboard import SummaryWriter
+
 from tianshou.data import Collector, VectorReplayBuffer
 from tianshou.policy import C51Policy
 from tianshou.trainer import offpolicy_trainer
--- a/examples/atari/atari_dqn.py
+++ b/examples/atari/atari_dqn.py
@@ -5,10 +5,10 @@ import pprint
 
 import numpy as np
 import torch
-from atari_network import DQN
-from atari_wrapper import make_atari_env
 from torch.utils.tensorboard import SummaryWriter
 
+from atari_network import DQN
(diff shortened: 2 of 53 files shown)


Trinkle23897:
Did you use the old version isort?
I use isort 5.9.3, yapf 0.31.0, python3.8, ubuntu 2004

Trinkle23897:
Why not use `disable`? https://stackoverflow.com/a/61109243/8773993

michalgregor:
Disable does not seem to work – not the way we are using the interface here. With disabled, n does not get incremented when doing update:

```
if self.disable:
            return

        if n < 0:
            self.last_print_n += n  # for auto-refresh logic to work
        self.n += n
```

https://github.com/tqdm/tqdm/blob/master/tqdm/std.py

michalgregor:
> Did you use the old version isort? I use isort 5.9.3, yapf 0.31.0, python3.8, ubuntu 2004

I think you are right! I assumed it was among the packages updated automatically by the script. I will make sure my versions are as close to yours as possible. What should I do then? Do I need to rerun ``make format`` and ``make commit-checks`` with them?

Trinkle23897:
`pip3 install isort yapf --upgrade`

See the version here: https://github.com/thu-ml/tianshou/runs/6470112919?check_suite_focus=true (Install dependencies)

```
Collecting yapf
  Downloading yapf-0.32.0-py2.py3-none-any.whl (190 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 190.2/190.2 kB 37.8 MB/s eta 0:00:00
Collecting isort
  Downloading isort-5.10.1-py3-none-any.whl (103 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 103.4/103.4 kB 22.3 MB/s eta 0:00:00
```

Trinkle23897:
> Disable does not seem to work – not the way we are using the interface here. With disabled, n does not get incremented when doing update:

```diff
         # perform n step_per_epoch
         with progress(
-            total=self.step_per_epoch, desc=f"Epoch #{self.epoch}", **tqdm_config
+            total=self.step_per_epoch, desc=f"Epoch #{self.epoch}", **tqdm_config, disable=True
         ) as t:
```

I manually tested with `disable=True` and `dummy_tqdm`. I didn't find any difference between these two approaches.

michalgregor:
Are you sure you are in fact running the disable=True on the tqdm version and not the dummy version? Because for me, with disable=True, episodes never terminate – which is what I would expect with t.n not getting updated.

Trinkle23897:
Yes you're right. I'll take it.

Trinkle23897 (approved):

PR #642, opened 2022-05-17
Added support for clipping to DQNPolicy.

michalgregor:
Added support for clipping to DQNPolicy.
    * When clip_delta=True is passed, smooth_l1_loss is used
    instead of the MSE loss.

    + [ ] exception-raising fix
    + [X] new feature
- [ ] I have reformatted the code using `make format` (**required**)

### DIFF
--- a/tianshou/policy/modelfree/dqn.py
+++ b/tianshou/policy/modelfree/dqn.py
@@ -26,6 +26,7 @@ class DQNPolicy(BasePolicy):
     :param bool reward_normalization: normalize the reward to Normal(0, 1).
         Default to False.
     :param bool is_double: use double dqn. Default to True.
+    :param bool clip_delta: use clipping. Default to True.
     :param lr_scheduler: a learning rate scheduler that adjusts the learning rate in
         optimizer in each policy.update(). Default to None (no lr_scheduler).
 
@@ -44,6 +45,7 @@ class DQNPolicy(BasePolicy):
         target_update_freq: int = 0,
         reward_normalization: bool = False,
         is_double: bool = True,
+        clip_delta: bool = False,
         **kwargs: Any,
     ) -> None:
         super().__init__(**kwargs)
@@ -62,6 +64,7 @@ class DQNPolicy(BasePolicy):
             self.model_old.eval()
         self._rew_norm = reward_normalization
         self._is_double = is_double
+        self._clip_delta = clip_delta
 
     def set_eps(self, eps: float) -> None:
         """Set the eps for epsilon-greedy exploration."""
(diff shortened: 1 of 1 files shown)


Trinkle23897:
Is it possible to add this feature for all DQN family?

michalgregor:
> Is it possible to add this feature for all DQN family?

Yes, I can go through them and make the changes. Do you have any idea which ones I will need to look into? E.g. there is a component in SAC that is trained DQN-style – should I also look there, for example?

Also, I am currently using smooth L1 (I saw it being used in another implementation), but PyTorch also has Huber loss...

Finally, the implementation is a bit redundant – the loss expects the output and the target – but since we also need td_error for the prio buffer, the way I implemented this, we do subtraction twice. But I am not sure how best to avoid that...

Trinkle23897:
> Do you have any idea which ones I will need to look into? E.g. there is a component in SAC that is trained DQN-style – should I also look there, for example?

After a second check of the current codebase, I think it's fine for this PR to only modify DQN loss. It's because other methods, e.g., IQN / FQF / QRDQN, they use a distribution to calculate loss, and it's non-trivial to directly apply this smooth_l1_loss. (BTW QRDQN has already use smooth_l1_loss to some extend)

> Also, I am currently using smooth L1 (I saw it being used in another implementation), but PyTorch also has Huber loss...

Could you please elaborate more on it?

> Finally, the implementation is a bit redundant – the loss expects the output and the target – but since we also need td_error for the prio buffer, the way I implemented this, we do subtraction twice. But I am not sure how best to avoid that...

That's fine.

michalgregor:
> > Also, I am currently using smooth L1 (I saw it being used in another implementation), but PyTorch also has Huber loss...
> 
> Could you please elaborate more on it?

I have seen DQN code using the smooth L1 loss, but as I recalled, the DQN paper uses the Huber loss – so I figured someone here might know why one would choose one over the other. But looking into it properly just now, I see that with the default parameters (beta=1 for the smooth L1 and delta=1 for the Huber loss), they are actually exactly the same, so there is no difference. Sorry for the confusion.

tianshou/policy/modelfree/dqn.py
```diff
@@ -26,6 +26,7 @@ class DQNPolicy(BasePolicy):
     :param bool reward_normalization: normalize the reward to Normal(0, 1).
         Default to False.
     :param bool is_double: use double dqn. Default to True.
+    :param bool clip_delta: use clipping. Default to True.
```
Trinkle23897: `clip_delta` is quite confusing at the first glance...

michalgregor: Should we change to just clipping then? The idea is that what is effectively getting clipped are the gradients, not the values in the forward run. I am not sure what the best name would be.

Trinkle23897: So at least you should state these in the docstring right?

michalgregor: > So at least you should state these in the docstring right?

Yes, that's a fair point, I will take care of it.

Trinkle23897 (approved):

PR #732, opened 2022-08-18
Fix clip when action_space is provided

liuliu:
Thie clip method in BasePolicy fixed to -1.0, 1.0. For Humanoid, its
control range is between -0.4, 0.4, it won't clip correctly. This PR
fixed that problem.

    + [ ] exception-raising fix
    + [ ] new feature


Trinkle23897 (approved):

Trinkle23897:
Oops, I guess it's not an easy fix, since the raw action from nn could be mostly [-1, 1] and the action_space range can be various.

liuliu:
> Oops, I guess it's not an easy fix, since the raw action from nn could be mostly [-1, 1] and the action_space range can be various.

Yeah, I didn't look at other experiments, for MuJoCo related ones, seems Humanoid-PPO run fine with the new range. Others all have range between [-1, 1].

Trinkle23897:
How about InvertedPendulum `[-3, 3]` and Pusher `[-2, 2]`?

liuliu:
> How about InvertedPendulum `[-3, 3]` and Pusher `[-2, 2]`?

Sorry, I think that I made a mistake when reading your code. It doesn't impact PPO at all because after clip, we did `action_scaling`. Hence, with this PR, InvertedPendulum and Pusher will fail on assertion at line 180. I am under the impression this is not the right fix. Do you think we should care about cases where `action_scaling = False`?

PR #748, opened 2022-09-22
Changes to support Gym 0.26.0

### LINKED ISSUE #726: env.step API updated in gym 0.25.0
    + [ ] exception-raising bug
    + [ ] RL algorithm bug
    + [X] new feature request
- [] I have mentioned version numbers, operating system and environment, where applicable:
  ```python
  ```
Gym 0.25.0 introduced a new [step API](https://github.com/openai/gym/blob/ddce4a56c85ebc7d8a07711231822eefaaa1d6ee/gym/core.py#L133) that returns a tuple of length 5(observation, reward, terminated, truncated, info) instead of 4(observation, reward, done, info).
I think tianshou's setup.py should restrict the gym version to <= 0.24.1 until it supports the new API.

Markus28:
Generally, I am trying to keep compatibility with Gym < 0.26.0 and envpool. 
- Vector environments will accept environments complying either with the old or the new step API and behave analogously. 
- Replay buffers now contain two additional fields: `terminated` and `truncated`. The `done` field does no longer have to be specified when data is added to the buffer.
- Collectors will work with environments complying either with the old or the new step API. If the environment uses the old API, the collector will infer `truncated` from `info` and set `terminated = done and not truncated` before adding the transition to the replay buffer.
- The `preprocess_fn` accepts either `done` or `terminated` and `truncated`. 
- Everything should work exactly as before with old-style environments, due to the way we infer `terminated`

This should not break any user code, unless the user is directly accessing the replay buffer. In particular, I didn't need to modify any of the training tests.


Markus28:
- I believe the PEP8 check is failing due to an update of flake8 (add exception or fix the problem?)
- Tests are failing due to changes in PettingZoo (additional PR for this?)
- I disabled wandb for now because they aren't supporting Gym 0.26.0 yet. This should be probably be turned on once wandb fixes it
- It may be prudent to add a CI run with Gym==0.23.0.

Trinkle23897 (dismissed):

Trinkle23897:
Hey @Markus28 , I adjust GPU test to gym==0.25.2, but it seems some of the tests are failed to pass. Could you please fix those?

Markus28:
Thanks! I think the issue was caused by `_SetAttrWrapper`. In Gym 0.25, wrappers automatically convert environments to the old step API. Thus, `RayVectorEnv` will always follow the old step API in Gym 0.25, even if the underlying environment uses the new API. I changed `_SetAttrWrapper` to pass the right settings to the super constructor, depending on the API of the underlying environment. It's somewhat inconvenient, but I don't see a better way without getting rid of `_SetAttrWrapper` entirely. The test in question is now passing locally for me

Trinkle23897 (approved):

PR #751, opened 2022-09-26
Added support for new PettingZoo API

Markus28:
Maybe someone who really knows PZ should do a sanity check here. I simply changed `step` to be compatible with both the old and the new step API. Removal of seeding was seemingly already anticipated in the existing code.


Trinkle23897 (dismissed):

tianshou/env/pettingzoo_env.py
```diff
@@ -57,7 +57,8 @@ def __init__(self, env: BaseWrapper):
 
     def reset(self, *args: Any, **kwargs: Any) -> Union[dict, Tuple[dict, dict]]:
         self.env.reset(*args, **kwargs)
-        observation, _, _, info = self.env.last(self)
+        last_return = self.env.last(self)
+        observation, info = last_return[0], last_return[-1]
```
WillDudley, line 83: Why have you done this rather than `observation, _, _, _, info = self.env.last(self)`  ? This would keep the style of Tianshou, Gym and PZ.

tianshou/env/pettingzoo_env.py
```diff
@@ -83,9 +84,13 @@ def reset(self, *args: Any, **kwargs: Any) -> Union[dict, Tuple[dict, dict]]:
...
 
-    def step(self, action: Any) -> Tuple[Dict, List[int], bool, Dict]:
+    def step(
+        self, action: Any
+    ) -> Union[Tuple[Dict, List[int], bool, Dict], Tuple[Dict, List[int], bool, bool,
+                                                         Dict]]:
         self.env.step(action)
-        observation, rew, done, info = self.env.last()
+        last_return = self.env.last()
+        observation = last_return[0]
```
WillDudley, line 118: Why have you done this rather than `observation, _, _, truncation, termination, info = self.env.last(self)`  ? This would keep the style of Tianshou, Gym and PZ.

tianshou/env/pettingzoo_env.py
```diff
@@ -105,15 +110,15 @@ def step(self, action: Any) -> Tuple[Dict, List[int], bool, Dict]:
 
         for agent_id, reward in self.env.rewards.items():
             self.rewards[self.agent_idx[agent_id]] = reward
-        return obs, self.rewards, done, info
+        return (obs, self.rewards, *last_return[2:])  # type: ignore
```
WillDudley, line 138: `(obs, self.rewards, truncation, termination, info)` may be clearer.

WillDudley (changes requested):
Thanks for the PR. I contend your choice to change the syntax, but I'm open to being convinced.

Markus28:
Yes, I completely agree with your remarks. The code was more readable before, due to the explicit labeling of the return values. 

I chose to do it this way to keep backward compatibility with older PZ versions (which use the old step API) at the expense of readability. 
However, I'd also be open to imposing pettingzoo>=1.21.0 and using your proposed syntax. Unfortunately, I would expect this to break some user code.

We have a similar dilemma for Gym, but I don't believe that we should impose gym>=0.26.0 since this would probably break a lot of user code and (afaik) envpool hasn't caught up with the API changes.

tianshou/env/pettingzoo_env.py
Markus28: See my comment below

WillDudley:
@Markus28 

Very good points. Indeed, some users may wish to use a new Tianshou feature with an old PZ env. However, I believe we do have wrappers to convert old PZ envs to new PZ envs pretty easily.

But the question is will the rest of Tianshou's update preserve backwards compat? Are you saying it will?

If it will, fab, but perhaps there should be warnings making users aware that:
a) the env they are using is old and may be incompatible with future updates, features & algos
b) it's easy to update the env by adding one extra line

Markus28:
@WillDudley my changes to tianshou should be backward compatible with respect to *Gym and PettingZoo* in the sense that you should be able to use old-style environments without any issues. However, it is not entirely compatible with respect to *existing user code* since my changes may (slightly) break existing code that directly interacts with the replay buffer. I would expect very few users to actually be affected, given the abstraction offered by collectors and the `BasePolicy`.

So for all intents and purposes, I think one can call the changes backward compatible.

I agree that deprecation warnings may be reasonable. What do you think @Trinkle23897 @pseudo-rnd-thoughts?

WillDudley:
Would personally vote for a depreciation warning encouraging users to upgrade their envs with a simple wrapper (for maintenance reasons), as well as code comments explaining that the unconventional code logic is due to backwards compatibility with PettingZoo<1.21.

@Markus28

WillDudley:
LGTM!

Trinkle23897 (approved):

PR #752, opened 2022-09-26
Added pre-commit

Markus28:
- This PR adds the checks that are defined in the Makefile as pre-commit hooks.
- Hopefully, the checks are equivalent to those from the Makefile, but I can't guarantee it. 
- CI remains as it is.
- As I pointed out on discord, I experienced some conflicts between flake8 and yapf, so it might be better to transition to some other combination (e.g. black).


Markus28:
Sorry, I ran the wrong python version. It appears mypy isn't happy on Python 3.7. I added the ignore hints back.

Markus28:
I needed to add some new `type: ignore` because the mypy version changed for CI. This should also fix the CI for other PRs (e.g. the PettingZoo one). It might be a good idea to pin the mypy version.

Trinkle23897:
I'd like to remove mypy in pre-commit because it has inconsistent behaviors with different python version and os, which will block contributors living in non py38+ubuntu2004

Trinkle23897 (approved):

nuance1979:
I support the transition from `yapf` to `black`. I personally find `black` more consistent with code formatting.

Trinkle23897:
agree

PR #753, opened 2022-10-02
Hindsight Experience Replay as a replay buffer

Juno-T:
    + [ ] exception-raising fix
    + [x] new feature
- [x] I have checked the code using ~~`make commit-checks`~~ `make check-codestyle`(**required**)

Related thread: [pr#510](https://github.com/thu-ml/tianshou/pull/510)

Hi, I've implemented Hindsight Experience Replay (HER) as a replay buffer and would like to contribute to tianshou.
I saw the discussion in the mentioned [pr#510](https://github.com/thu-ml/tianshou/pull/510) and it seems to be inactive for quite a while so I decided to make a new, independent pull request.

## implementation
I implemented HER solely as a replay buffer. It is done by temporarily directly re-writing transitions storage (`self._meta`) during the `sample_indices()` call. The original transitions are cached and will be restored at the beginning of the next sampling or when other methods is called. This will make sure that. for example, n-step return calculation can be done without altering the policy.

There is also a problem with the original indices sampling. The sampled indices are not guaranteed to be from different episodes. So I decided to perform re-writing based on the episode. This guarantees that the sampled transitions from the same episode will have the same re-written goal. This also make the re-writing ratio calculation slightly differ from the paper, but it won't be too different if there are many episodes in the buffer.

In the current commit, HER replay buffer only support 'future' strategy and online sampling. This is the best of HER in term of performance and memory efficiency.

I also add a few more convenient replay buffers (`HERVectorReplayBuffer`, `HERReplayBufferManager`), test env (`MyGoalEnv`), gym wrapper (`TruncatedAsTerminated`), unit tests, and a simple example (examples/offline/fetch_her_ddpg.py).

## verification
I have added unit tests for almost everything I have implemented.
HER replay buffer was also tested using DDPG on [`FetchReach-v3` env](https://github.com/Farama-Foundation/Gymnasium-Robotics). I used default DDPG parameters from mujoco example and didn't tune anything further to get this good result! (train script: examples/offline/fetch_her_ddpg.py).

Screen Shot 2022-10-02 at 19 22 53

### Todo
- [ ] Compile doc with `make doc`


Trinkle23897:
The ubuntu GPU test is run on gym==0.25.2. I'll give some feedback later today.

Juno-T:
I see. I'll fix it after the feedback then.

README.md
```diff
@@ -39,6 +39,7 @@
 - [Generalized Advantage Estimator (GAE)](https://arxiv.org/pdf/1506.02438.pdf)
 - [Posterior Sampling Reinforcement Learning (PSRL)](https://www.ece.uvic.ca/~bctill/papers/learning/Strens_2000.pdf)
 - [Intrinsic Curiosity Module (ICM)](https://arxiv.org/pdf/1705.05363.pdf)
+- [Hindsight Experience Replay (HER)](https://arxiv.org/pdf/1707.01495.pdf)
```
Trinkle23897, line 42: add in docs/index.rst

tianshou/data/buffer/her.py
```diff
@@ -0,0 +1,180 @@
...
+class HERReplayBuffer(ReplayBuffer):
+    """Implementation of Hindsight Experience Replay. arXiv:1707.01495.
+
+    Currently support only 'future' strategy of HER.
+
+    :param int size: the size of the replay buffer.
+    :param compute_reward_fn: a function that takes 3 arguments: \
+        `acheived_goal`, `desired_goal`, `info` and returns the reward(s).
+        Note that the goal arguments can have extra batch_size dimension and in that \
+        case, the rewards of size batch_size should be returned
```
Trinkle23897: can mention the size/shape here, is it for single or batch input?

tianshou/data/buffer/her.py
```diff
@@ -0,0 +1,180 @@
...
+
+        Currently applies only HER's 'future' strategy. The new goals will be written \
+        directly to the internal batch data temporarily and will be restored right \
+        before the next sampling or when using some of the buffer's method (e.g. \
+        `add`, `save_hdf5`, etc.). This is to make sure that n-step returns \
+        calculation etc., performs correctly without additional alteration.
+        """
+        if indices.size == 0:
+            return
+        # Construct episode trajectories
```
Trinkle23897, line 114: I do think you should sort indices here. An example:
```
indices = [9, 7, 8, 6, 4, 5, 3, 1, 2, 0]
done = [0, 0, 1, 0, 0, 1, 0, 0, 1, 1]
```
after some processing with horizon=8, `unique_ep_open_indices = [7, 4, 1, 0]` and `unique_ep_close_indices=[ 3,  0, -1,  9]`, which makes no sense

tianshou/data/buffer/her.py
```diff
@@ -0,0 +1,180 @@
...
+        unique_ep_indices = indices[:, unique_ep_open_indices]
+        #   close indices are used to find max future_t among presented episodes
+        unique_ep_close_indices = np.hstack(
+            [(unique_ep_open_indices - 1)[1:],
+             len(terminal) - 1]
+        )
+        #   episode indices that will be altered
+        her_ep_indices = np.random.choice(
+            len(unique_ep_open_indices),
+            size=int(len(unique_ep_open_indices) * self.future_p),
```
Trinkle23897, line 140: Could you please briefly explain why we need `self.future_p` here, instead of all indices?

If I understand correctly, these few lines do some random HER update on the part of trajectories instead of all trajectories.

tianshou/data/buffer/her.py
```diff
@@ -0,0 +1,180 @@
...
+        It's called everytime before 'writing', 'sampling' or 'saving' the buffer.
+        """
+        if not hasattr(self, '_altered_indices'):
+            return
+
+        if self._altered_indices.size == 0:
+            return
+        self._meta[self._altered_indices] = self._original_meta
+        # Clean
+        del self._original_meta, self._altered_indices
```
Trinkle23897: no need to manually delete them because it has been overwritten in the next line?

Juno-T: fixed: https://github.com/thu-ml/tianshou/pull/753/commits/63e65c087fa4c070861f12ba4d5fea130fc56740

README.md
Juno-T: updated: https://github.com/thu-ml/tianshou/pull/753/commits/9d52936b6b368abbdd8736835300880fdeeda101

tianshou/data/buffer/her.py
Juno-T: updated: https://github.com/thu-ml/tianshou/commit/63e65c087fa4c070861f12ba4d5fea130fc56740

tianshou/data/buffer/her.py
Juno-T: Previously I sort indices in `sample_indices` method (since it is the only one that calls `rewrite_transitions`) but yeah, I agree that sort indices should be inside `rewrite_transitions`.

fixed: https://github.com/thu-ml/tianshou/commit/63e65c087fa4c070861f12ba4d5fea130fc56740

tianshou/data/buffer/her.py
Juno-T: Yes, you are correct. HER only performs goal re-writing on some random episodes (the amount is depending on the ratio parameter `future_k`, or in here, probablity `future_p`).

Juno-T:
Btw, I tried verifying documentation page by using `make doc` but I kept getting version mismatch error. The doc test seems to be ok so I think it shouldn't be a problem.

The error, just in case:
```
WARNING: while setting up extension sphinx.addnodes: node class 'meta' is already registered, its visitors will be overridden
/home/tianshou/venv/lib/python3.8/site-packages/sphinx/jinja2glue.py:106: DeprecationWarning: 'contextfunction' is renamed to 'pass_context', the old name will be removed in Jinja 3.1.
  def warning(context: Dict, message: str, *args: Any, **kwargs: Any) -> str:

Exception occurred:
  File "/home/tianshou/venv/lib/python3.8/site-packages/pkg_resources/__init__.py", line 791, in resolve
    raise VersionConflict(dist, req).with_context(dependent_req)
pkg_resources.ContextualVersionConflict: (docutils 0.19 (/home/tianshou/venv/lib/python3.8/site-packages), Requirement.parse('docutils<0.17,>=0.12'), {'Sphinx'})
The full traceback has been saved in /tmp/sphinx-err-_u9sq96y.log, if you want to report the issue to the developers.
Please also report this if it was a user error, so that a better error message can be provided next time.
A bug report can be filed in the tracker at <https://github.com/sphinx-doc/sphinx/issues>. Thanks!
make[1]: *** [Makefile:20: html] Error 2
make[1]: Leaving directory '/home/tianshou/docs'
make: *** [Makefile:47: doc] Error 2
```

tianshou/data/buffer/her.py
Juno-T: Here I also remove the third argument, `info`. Previously this argument was manually set to empty dict `{}` during the sampling because its shape was complicated. Originally I have it there because `Fetch` env's [`compute_reward`](https://github.com/Farama-Foundation/Gymnasium-Robotics/blob/main/gymnasium_robotics/envs/fetch_env.py#L67) function also takes `info` argument, although it was never used.

tianshou/data/buffer/her.py
Trinkle23897: My bad. I didn't see L92 previously, but it also has an issue: if the replay buffer start index is not 0 (i.e., overwrite previous data point in a circular buffer), the sort approach will cause some bugs at some point. It should start with x where x is the smallest index that is greater or equal to start index.

ref:https://github.com/thu-ml/tianshou/blob/0181fe79a5915791a8d22112dfedc905b6169cf8/tianshou/data/buffer/base.py#L285-L289

tianshou/data/buffer/her.py
```diff
@@ -0,0 +1,183 @@
...
+
+from tianshou.data import Batch, ReplayBuffer
+
+
+class HERReplayBuffer(ReplayBuffer):
+    """Implementation of Hindsight Experience Replay. arXiv:1707.01495.
+
+    `HERReplayBuffer` is to be used with goal-based environment where the \
+    observation is a dictionary with keys `observation`, `achieved_goal` and \
+    `desired_goal`. Currently support only HER's future strategy, online sampling.
```
Trinkle23897:
```suggestion
    HERReplayBuffer is to be used with goal-based environment where the
    observation is a dictionary with keys ``observation``, ``achieved_goal`` and
    ``desired_goal``. Currently support only HER's future strategy, online sampling.
```

tianshou/data/buffer/her.py
```diff
@@ -0,0 +1,183 @@
...
+    :param int size: the size of the replay buffer.
+    :param compute_reward_fn: a function that takes 2 `np.array` arguments, \
+        `acheived_goal` and `desired_goal`, and returns rewards as `np.array`.
+        The two arguments are of shape (batch_size, *original_shape) and the returned \
+        rewards must be of shape (batch_size,).
+    :param int horizon: the maximum number of steps in an episode.
+    :param int future_k: the 'k' parameter introduced in the paper. In short, there \
+        will be at most k episodes that are re-written for every 1 unaltered episode \
+        during the sampling.
+
```
Trinkle23897:
```suggestion
    :param compute_reward_fn: a function that takes 2 ``np.array`` arguments,
        ``acheived_goal`` and ``desired_goal``, and returns rewards as ``np.array``.
        The two arguments are of shape (batch_size, *original_shape) and the returned
        rewards must be of shape (batch_size,).
    :param int horizon: the maximum number of steps in an episode.
    :param int future_k: the 'k' parameter introduced in the paper. In short, there
        will be at most k episodes that are re-written for every 1 unaltered episode
        during the sampling.

```

Trinkle23897:
> Btw, I tried verifying documentation page by using make doc but I kept getting version mismatch error. The doc test seems to be ok so I think it shouldn't be a problem.

Have you tried to reinstall deps by `pip install -e ".[dev]"`?

tianshou/data/buffer/her.py
Juno-T: You are right. This case somehow slipped my mind.
In `rewrite_transitions`, sorting indices serves as a way to group indices from the same episode together and order them chronologically. So to apply with the circular property, I'm thinking of sorting the indices, then shifting them so that the oldest index is at the front.
 
Will fix this later maybe today or tomorrow.

Trinkle23897: Just few lines I think:
```py
mask = indice >= self._index
indice[~mask] += self._length
sort(indice)
indice[indice >= self._length] -= self._length

File: tianshou/data/buffer/her.py (hunk 7 above)
Comment (Juno-T) [reply tianshou/data/buffer/her.py]: updated: https://github.com/thu-ml/tianshou/pull/753/commits/7802451a0200bdf4e4c0399a204038d9dcf74925

File: tianshou/data/buffer/her.py (hunk 6 above)
Comment (Juno-T) [reply tianshou/data/buffer/her.py]: updated: https://github.com/thu-ml/tianshou/pull/753/commits/7802451a0200bdf4e4c0399a204038d9dcf74925

File: tianshou/data/buffer/her.py (hunk 3 above)
Comment (Juno-T) [reply tianshou/data/buffer/her.py:114]: Updated & add some test : https://github.com/thu-ml/tianshou/pull/753/commits/6ee7d08f414456ce2274e8ca0e6714e5cbe420c7

**Comment by Juno-T** [comment]
I have tried `pip install -e ".[dev]"` and manually downgraded to `docutils==0.16`. Both didn't work.
I wonder if the latest versions (installed during [`make doc`](https://github.com/thu-ml/tianshou/blob/master/Makefile)) of sphinx and docutils aren't compatible.
Could you share your sphinx and docutils version please?

**Comment by Trinkle23897** [comment]
`pip3 install "sphinx<4" sphinxcontrib-bibtex "jinja2<3.1"`

File: tianshou/data/buffer/her.py [hunk 8]
```diff
@@ -105,7 +105,12 @@ def rewrite_transitions(self, indices: np.ndarray) -> None:
         """
         if indices.size == 0:
             return
+
+        # Sort indices keeping chronological order
+        mask = indices >= self._index
+        indices[~mask] += self.maxsize
```
Trinkle23897: `indices[indices < self._index] += self.maxsize`

examples/offline/fetch_her_ddpg.py
```diff
@@ -0,0 +1,265 @@
...
+        logger = TensorboardLogger(writer)
+    else:  # wandb
+        logger.load(writer)
+
+    def save_best_fn(policy):
+        torch.save(policy.state_dict(), os.path.join(log_path, "policy.pth"))
+
+    if not args.watch:
+        # trainer
+        result = offpolicy_trainer(
```
nuance1979: Why is this file under `examples/offline` if `offpolicy_trainer()` is called? Should it be under `examples/mujoco` instead?

examples/offline/fetch_her_ddpg.py
```diff
@@ -0,0 +1,265 @@
...
+from tianshou.trainer import offpolicy_trainer
+from tianshou.utils import TensorboardLogger, WandbLogger
+from tianshou.utils.net.common import Net
+from tianshou.utils.net.continuous import Actor, Critic
+
+
+def get_args():
+    # python3 fetch_her_ddpg.py --task FetchReach-v3 --seed 0 --horizon 50
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--task", type=str, default="FetchReach-v3")
```
nuance1979: Is it possible to apply HER to other tasks, e.g., HalfCheetah-v3, Ant-v3, so that we can compare performances? If not, can you run DDPG on FetchReach-v3 with `mujoco_ddpg.py` to get a baseline number? Then you can update `examples/mujoco/README.md` with both numbers, i.e., DDPG with and without HER.

examples/offline/fetch_her_ddpg.py
```diff
@@ -0,0 +1,265 @@
...
+    train_envs = ShmemVectorEnv(
+        [lambda: TruncatedAsTerminated(gym.make(task)) for _ in range(training_num)]
+    )
+    test_envs = ShmemVectorEnv(
+        [lambda: TruncatedAsTerminated(gym.make(task)) for _ in range(test_num)]
+    )
+    return env, train_envs, test_envs
+
+
+class DictStateNet(Net):
```
nuance1979: I'm a bit concerned about the fact that the user need to follow the example in this file closely in order to apply HER to any other task. Can you rewrite `DictStateNet` as a [decorator](https://en.wikipedia.org/wiki/Decorator_pattern) of `Net` so that it can be applied to any `Net`? The way I see it, this decorator should reside in [common.py](https://github.com/thu-ml/tianshou/blob/master/tianshou/utils/net/common.py) or [continous.py](https://github.com/thu-ml/tianshou/blob/master/tianshou/utils/net/continuous.py) as a utility function to wrap any `Net` which wants to work with `HERReplayBuffer`.

examples/offline/fetch_her_ddpg.py
```diff
@@ -0,0 +1,265 @@
...
+        self,
+        obs: Union[Dict[str, Union[np.ndarray, torch.Tensor]], Union[np.ndarray,
+                                                                     torch.Tensor]],
+        state: Any = None,
+        info: Dict[str, Any] = {},
+    ) -> Tuple[torch.Tensor, Any]:
+        return super().forward(self.preprocess_obs(obs), state, info)
+
+
+class GoalStateCritic(Critic):
```
nuance1979: Same for `GoalStateCritic`.

Juno-T:
Sorry for missing for a few weeks, I had to clear up some work at hand.

In the latest update, https://github.com/thu-ml/tianshou/pull/753/commits/72b4ab446e4e6bd4f717a6d2669c96f4ac501595, I have:
- Debuged HER
- Updated documentation
- Updated the code as suggested 
(https://github.com/thu-ml/tianshou/pull/753#discussion_r988572521)
- Moved HER training example to mujoco folder 
(https://github.com/thu-ml/tianshou/pull/753#discussion_r989543191)
- Refactored the "dict state" code into a decorator according to the suggestion. I put it in `utils/net/common.py` 
(https://github.com/thu-ml/tianshou/pull/753#discussion_r989572583)
- Performed a simple evalutation against vanilla DDPG and reported the result in `examples/mujoco/README.md`. I'm not sure where to put it since HER is not a learning algorithm but a techqnique, so I created and put it in the "Others" section. 
(https://github.com/thu-ml/tianshou/pull/753#discussion_r989546459)
![W B Chart 10_27_2022, 11_48_23 AM] 
result_HER

nuance1979:
LGTM. Great work, @Juno-T !

nuance1979:
@Trinkle23897 I have fixed the test failures. Feel free to approve or suggest more changes.

Trinkle23897 (approved):
LGTM! Thanks for the great works!

PR #781, opened 2022-12-02
Updated atari wrappers, fixed pre-commit

### LINKED ISSUE #772: about the gym version
    + [ ] exception-raising bug
    + [ ] RL algorithm bug
    + [ ] new feature request
  ```python
  ```

gym 0.26.0 does not work, but 
pip install gym==0.25.0 will work.

python examples/atari/atari_dqn.py
will get the results.

Markus28:
This PR addresses #772 (updates Atari wrappers to work with new Gym API) and some additional issues:

- Pre-commit was using gitlab for flake8, which as of recently requires authentication -> Replaced with GitHub
- Yapf was quietly failing in pre-commit. Changed it such that it fixes formatting in-place
- There is an incompatibility between flake8 and yapf where yapf puts binary operators after the line break and flake8 wants it before the break. I added an exception for flake8.
- Also require `packaging` in setup.py

My changes shouldn't change the behaviour of  the wrappers for older versions, but please double check.
Idk whether it's just me, but there are always some incompatibilities between yapf and flake8 that need to resolved manually. It might make sense to try black instead.


nuance1979:
> Idk whether it's just me, but there are always some incompatibilities between yapf and flake8 that need to resolved manually. It might make sense to try black instead.

I concur. `black` seems to be more consistent than `yapf`.

Trinkle23897:
I agree we should change to black at some point.

Trinkle23897 (approved):

PR #789, opened 2022-12-29
Gymnasium Integration

Markus28:
Changes:
- Disclaimer in README
- Replaced all occurences of Gym with Gymnasium
- Removed code that is now dead since we no longer need to support the old step API
- Updated type hints to only allow new step API
- Increased required version of envpool to support Gymnasium
- Increased required version of PettingZoo to support Gymnasium
- Updated `PettingZooEnv` to only use the new step API, removed hack to also support old API
- I had to add some `# type: ignore` comments, due to new type hinting in Gymnasium. I'm not that familiar with type hinting but I believe that the issue is on the Gymnasium side and we are looking into it.
- Had to update `MyTestEnv` to support `options` kwarg
- Skip NNI tests because they still use OpenAI Gym
- Also allow `PettingZooEnv` in vector environment
- Updated doc page about ReplayBuffer to also talk about terminated and truncated flags.

Still need to do: 
- Update the Jupyter notebooks in docs
- Check the entire code base for more dead code (from compatibility stuff)
- Check the reset functions of all environments/wrappers in code base to make sure they use the `options` kwarg
- Someone might want to check test_env_finite.py
- Is it okay to allow `PettingZooEnv` in vector environments? Might need to update docs?


Trinkle23897:
Hey @Markus28, I'm actually thinking about a non-breaking change. If a user uses gym env, tianshou's VectorEnv itself will first detect the class of the env, and if it's gym env then we adapt shimmy wrapper on top of that. This is an automated process and won't degrade user-experience.

The current changes looks fine for me but it would be nice to add the above feature request!

Also, have you added shimmy usage in docs or readme so that people can know how to convert it to a gymnasium env without seeking other resources for help?

Markus28:
I can do that :) 
Currently, I only have a disclaimer in the README, but I can add a section to the docs

Markus28:
So it turns out that there is unfortunately no "universal" shimmy wrapper that will take any Gym environment and give us an up-to-date Gymnasium environment. Shimmy implements different wrappers for the different API versions. I think we have the following options: 

- Just let the user do the conversion
- Use different Shimmy wrappers depending on `gym.__version__`. That could lead to problems for people who are using environments with APIs that don't match their gym version
- Implement a universal wrapper ourselves

Mark seems to believe that doing a universal wrapper would be quite complicated, but I don't see why it shouldn't be doable. I'm not sure whether it would then be part of Shimmy or Tianshou though.

Trinkle23897:
I personally in favor of the second option. This will reduce most of (90%) the issue user hit and will reduce the maintenance effort quite a lot.

tianshou/env/venvs.py
```diff
@@ -1,21 +1,87 @@
...
+        gym_version = packaging.version.parse(old_gym.__version__)
+        if gym_version >= packaging.version.parse("0.26.0"):
+            return shimmy.GymV26CompatibilityV0(env=env)
+        elif gym_version >= packaging.version.parse("0.22.0"):
+            return shimmy.GymV22CompatibilityV0(env=env)
+        else:
+            raise Exception(
+                f"Found OpenAI Gym version {gym.__version__}. "
+                f"Tianshou only supports OpenAI Gym environments of version>=0.22.0"
+            )
```
Markus28, line 82: @pseudo-rnd-thoughts Can you confirm that this is how the shimmy wrappers are intended to be used? I.e. for any environment `0.26.0>gym.__version__>=0.22.0` we use `GymV22CompatibilityV0"` and for `gym.__version__ >= 0.26.0` we use `GymV26CompatibilityV0`?

Markus28:
I implemented the second version. I will still need to update some documentation and the tutorials :) There are currently no tests for OpenAI Gym environments, let me know if I should add some.

nuance1979:
> I implemented the second version. I will still need to update some documentation and the tutorials :) There are currently no tests for OpenAI Gym environments, let me know if I should add some.

Yes, I think we definitely need a test which imports `gym` and triggers the use of `shimmy`. I'd suggest making a copy of `test_dqn.py` to `test_dqn_gym.py` then only replace `import gymnasium as gym` with `import gym`.

nuance1979:
If we decide to use `gymnasium` as the preferred environment package, then this line should be updated to reflect that:

https://github.com/thu-ml/tianshou/blob/774d3d8e833a1b1c1ed320e0971ab125161f4264/setup.py#L18

setup.py
```diff
@@ -48,18 +48,19 @@ def get_extras_require() -> str:
             "doc8",
             "scipy",
             "pillow",
-            "pettingzoo>=1.17",
+            "pettingzoo>=1.22",
             "pygame>=2.1.0",  # pettingzoo test cases pistonball
             "pymunk>=6.2.1",  # pettingzoo test cases pistonball
             "nni>=2.3,<3.0",  # expect breaking changes at next major version
             "pytorch_lightning",
+            "gym>=0.22.0",
```
nuance1979, line 56: Probably add `shimmy` here?

Markus28: Yes, thanks for spotting that :) Also I should increase the required gymnasium version to >=0.26.0 I don't think there are any earlier releases but just to be sure

nuance1979:
@Trinkle23897 This PR is failing the doc spelling check but the error message is quite cryptic:

```
../../../../../../opt/hostedtoolcache/Python/3.8.16/x64/lib/python3.8/site-packages/tianshou/env/gym_wrappers.py:docstring of tianshou.env.gym_wrappers.ContinuousToDiscrete.action:4: : Spell check: Args: Args:.
```

I have no idea how to fix it. Can you take a look?

Other than that, I think this PR is ready to merge.

Trinkle23897:
will take a look tonight

Trinkle23897:
hmm on my local:
```
>>> gym.__version__, shimmy.__version__, gymnasium.__version__
('0.25.0', '0.2.0', '0.27.1')

$ python test_dqn_gym.py
  File "/Users/jiayi/github/tianshou/test/discrete/test_dqn_gym.py", line 172, in <module> 
    test_dqn(get_args())
  File "/Users/jiayi/github/tianshou/test/discrete/test_dqn_gym.py", line 64, in test_dqn
    train_envs = DummyVectorEnv(
  File "/Users/jiayi/github/tianshou/tianshou/env/venvs.py", line 428, in __init__
    super().__init__(env_fns, DummyEnvWorker, **kwargs)
  File "/Users/jiayi/github/tianshou/tianshou/env/venvs.py", line 148, in __init__
    self.workers = [worker_fn(_patch_env_generator(fn)) for fn in env_fns]
  File "/Users/jiayi/github/tianshou/tianshou/env/venvs.py", line 148, in <listcomp>
    self.workers = [worker_fn(_patch_env_generator(fn)) for fn in env_fns]
  File "/Users/jiayi/github/tianshou/tianshou/env/worker/dummy.py", line 13, in __init__
    self.env = env_fn()
  File "/Users/jiayi/github/tianshou/tianshou/env/venvs.py", line 77, in patched
    return shimmy.GymV22CompatibilityV0(env=env)
  File "/Users/jiayi/.virtualenvs/openai/lib/python3.9/site-packages/shimmy/openai_gym_compatibility.py", line 199, in __init__
    gym_env = _strip_default_wrappers(gym_env)
  File "/Users/jiayi/.virtualenvs/openai/lib/python3.9/site-packages/shimmy/openai_gym_compatibility.py", line 280, in _strip_default_wrappers
    gym.wrappers.render_collection.RenderCollection,
AttributeError: module 'gym.wrappers' has no attribute 'render_collection'
```

pseudo-rnd-thoughts:
You need to update your gym version, but we will update shimmy to avoid this issue

Trinkle23897 (approved):
leave `render_collection` as-is, will educate user to upgrade the gym version as temporary solution

PR #817, opened 2023-03-01
fix bug (#811)

### LINKED ISSUE #811: A Bug in HERReplayBuffer.
    + [ ] exception-raising bug
    + [x] RL algorithm bug
    + [ ] new feature request
  ```python
  ```

Hello, I have some questions in the code of ```rewrite_transitions``` in ```HERReplayBuffer```:
```python
    def rewrite_transitions(self, indices: np.ndarray) -> None:
        """Re-write the goal of some sampled transitions' episodes according to HER.
        Currently applies only HER's 'future' strategy. The new goals will be written \
        directly to the internal batch data temporarily and will be restored right \
        before the next sampling or when using some of the buffer's method (e.g. \
        `add`, `save_hdf5`, etc.). This is to make sure that n-step returns \
        calculation etc., performs correctly without additional alteration.
        """
        if indices.size == 0:
            return

        # Sort indices keeping chronological order
        indices[indices < self._index] += self.maxsize
        indices = np.sort(indices)
        indices[indices >= self.maxsize] -= self.maxsize

        # Construct episode trajectories
        indices = [indices]
        for _ in range(self.horizon - 1):
            indices.append(self.next(indices[-1]))
        indices = np.stack(indices)

        # Calculate future timestep to use
        current = indices[0]
        terminal = indices[-1]
        future_offset = np.random.uniform(size=len(indices[0])) * (terminal - current)
        future_offset = future_offset.astype(int)
        future_t = (current + future_offset)
        ...
```
As ReplayBuffer is implemented as a circular queue, the indices in ```terminal``` may be less than the corresponding indices in ```current```. So that some elements in ```future_offset``` may be negative, which will make the states in ```future_t``` not desired future states.
@Juno-T

sunkafei:
    + [ ] exception-raising fix
    + [ ] new feature

Fix the bug#811.

I run the command `make commit-checks` but the get some errors which are not relevent to my code:
```
tianshou/policy/modelfree/trpo.py:131:29: B028 No explicit stacklevel keyword argument found. The warn method from the warnings module uses a stacklevel of 1 by default. This will only show a stack trace for the line on which the warn method is called. It is therefore recommended to use a stacklevel of 2 or greater to provide more information to the user.
                            warnings.warn(
                            ^
```
Just ignore it?

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -8,7 +8,7 @@ exclude =
 	dist
 	*.egg-info
 max-line-length = 87
-ignore = B305,W504,B006,B008,B024,W503
+ignore = B305,W504,B006,B008,B024,W503,B028
 
 [yapf]
 based_on_style = pep8
--- a/setup.py
+++ b/setup.py
@@ -22,7 +22,6 @@ def get_install_requires() -> str:
         "torch>=1.4.0",
         "numba>=0.51.0",
         "h5py>=2.10.0",  # to match tensorflow's minimal requirements
-        "protobuf~=3.19.0",  # breaking change, sphinx fail
         "packaging",
     ]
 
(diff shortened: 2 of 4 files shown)


Trinkle23897:
I’ll take it, don’t worry

Trinkle23897 (dismissed):

Trinkle23897 (approved):

PR #920, opened 2023-08-23
Poetry and poe instead of setuptools and make

### LINKED ISSUE #914: Modernize build, dependencies and types
This issue is addressing previous comments and discussions about tianshou and adds some new ones. A preliminary list of tasks is:

- [ ] Deprecate python 3.8, use inbuilt types for annotations wherever possible (i.e. `list[str]` instead of `List[str]`, and so on)
- [ ] Use ruff and black instead of current formatter
- [ ] Move build to poetry, if possible
- [ ] If moved to poetry: replace `make` by [poethepoet](https://github.com/nat-n/poethepoet).
- [ ] Adjust pre-commit and CI to handle all these changes

Optional:
- [ ] Improve how documentation is built (not sure whether it's necessary)

@Trinkle23897 We havent't discussed using poe instead of make yet. I suggest doing it because:

1. No need for an external tool, nor an external config file
2. Works with poetry out of the box. Currently `make` will install dependencies into whichever virtualenv it's called from, which is not very safe or clean (in my opinion)
3. In fact, no additional dependencies or check for install will be needed, as one can use poetry's dependency groups for that

MischaPanch:
Closes #914 

I had to make one "functional" change: added `__iter__` to `Batch`. This was needed because ruff kept removing the `.keys()` in code like `for i in batch.keys()`. Here ruff is probably right and `__iter__` should exist in `Batch`. It's backwards compatible.

I bumped some dev dependencies (like ray) and removed some b/c I didn't understand why they were needed (like pytorch lightning).

The remaining changes in the code are just formatting. Ruff took care of replacing `List` by `list` and removing the imports


MischaPanch:
@Trinkle23897 here tests fail b/c poetry installs a gpu-based torch into a cpu env. How do you want to handle this? Should we declare the default dependency to be torch for cpu and the users can take care themselves of installing the correct torch afterwards? How did you do it so far?

examples/mujoco/mujoco_trpo.py
```diff
@@ -8,17 +8,16 @@
...
 from tianshou.data import Collector, ReplayBuffer, VectorReplayBuffer
 from tianshou.policy import TRPOPolicy
 from tianshou.trainer import OnpolicyTrainer
 from tianshou.utils import TensorboardLogger, WandbLogger
 from tianshou.utils.net.common import Net
 from tianshou.utils.net.continuous import ActorProb, Critic
+from torch import nn
+from torch.distributions import Independent, Normal
+from torch.optim.lr_scheduler import LambdaLR
+from torch.utils.tensorboard import SummaryWriter
```
Trinkle23897, line 20: could you add `torch` to `.isort.cfg`'s known_firstparty to keep the original order?

pyproject.toml
```diff
@@ -0,0 +1,144 @@
...
+src = ["tianshou", "test"]
+target-version = "py39"
+
+[tool.ruff.pydocstyle]
+ignore-decorators = ["D100" ,"D102", "D104", "D105", "D107", "D203", "D213", "D401", "D402"]
+# exclude = "^(test/)|(docs/)|(examples/)|(setup.py)"
+
+[tool.poetry_bumpversion.file."tianshou/__init__.py"]
+
+[tool.poetry-sort]
```
Trinkle23897, line 122: is there a command for package building? Similar to `python setup.py sdist bdist_wheel`

pyproject.toml
```diff
@@ -0,0 +1,144 @@
...
+gymnasium = ">=0.29.0"
+h5py = ">=3.9.0"
+numba = ">=0.57.1"
+numpy = ">=1"
+packaging = "*"
+pettingzoo = ">=1.22"
+ray = ">=2"
+tensorboard = ">=2.5.0"
+# Torch 2.0.1 causes problems, see https://github.com/pytorch/pytorch/issues/100974
+torch = "==2.0.0"
```
Trinkle23897, line 39: could you change to `>=2.0,!=2.0.1` if possible?

pyproject.toml
```diff
@@ -0,0 +1,144 @@
+[build-system]
+requires = ["poetry-core>=1.0.0"]
+build-backend = "poetry.core.masonry.api"
+
+[tool.poetry]
+name = "tianshou"
+version = "0.5.1"
```
Trinkle23897, line 7: bump version to 0.6.0?

Trinkle23897:
> here tests fail b/c poetry installs a gpu-based torch into a cpu env. How do you want to handle this? Should we declare the default dependency to be torch for cpu and the users can take care themselves of installing the correct torch afterwards? How did you do it so far?

Not sure tbh... The current solution is just `pip install torch` which is compatible with both cpu and gpu environments. Doesn't poetry have the same functionality? Could you check how CleanRL solves this issue?

Trinkle23897:
I'd suggest only doing ruff/black/python3.9 in this PR and split the poetry modification in another PR

pyproject.toml
MischaPanch: I generally wanted to discuss the release mechanism with you. How do you do it currently?

I added support for bumpversion here, so the versioning can be handled automatically. I have a script and some CI workflows for version bumping and releasing packages in other projects, do you think it makes sense to do that in tianshou as well? You could then release by just calling the script.

Here an example of the [script](https://github.com/appliedAI-Initiative/accsr/blob/develop/build_scripts/release-version.sh) and [CI workflow](https://github.com/appliedAI-Initiative/accsr/blob/develop/.github/workflows/release.yaml) in action.

MischaPanch:
> I'd suggest only doing ruff/black/python3.9 in this PR and split the poetry modification in another PR

Created a new PR for that, #921 . Once it's merged, I will adjust this one. The diff should be more manageable then :)

examples/mujoco/mujoco_trpo.py
MischaPanch: will do

pyproject.toml
MischaPanch: yeah, `poetry build` should do the job. You can also publish with `poetry publish`. Let's maybe discuss the release process at some point, as mentioned in a thread below

pyproject.toml
MischaPanch: Will do. Hopefully pytorch people will indeed fix this in `2.0.2`

MischaPanch:
Closing this for now until #921 is merged

pyproject.toml
MischaPanch: Done in the other PR

examples/mujoco/mujoco_trpo.py
MischaPanch: Done in the other PR

PR #921, opened 2023-08-24
Python 3.9, black + ruff formatting

MischaPanch:
Preparation for #914 and #920

Changes formatting to ruff and black. Remove python 3.8

## Additional Changes

- Removed flake8 dependencies
- Adjusted pre-commit. Now CI and Make use pre-commit, reducing the duplication of linting calls
- Removed check-docstyle option (ruff is doing that)
- Merged format and lint. In CI the format-lint step fails if any changes are done, so it fulfills the lint functionality.


MischaPanch:
@Trinkle23897 feel free to directly adjust the ruff config on this branch to your liking - I think that's easier than going back and forth with comments :).

I will not push on it from now on unless you make comments, so there won't be any conflicts

Trinkle23897:
I'm done

Trinkle23897:
hm let me fix more

Trinkle23897:
my bad, I would like to move to [73e49c6](https://github.com/thu-ml/tianshou/pull/921/commits/73e49c65ff71738f13c3625d7069cab8ac668a3f)'s `pyproject.toml` and `.flake8`, but introduced some error :(

will fix it tomorrow or weekend, or if you have time feel free to fix that

MischaPanch:
> my bad, I would like to move to [73e49c6](https://github.com/thu-ml/tianshou/pull/921/commits/73e49c65ff71738f13c3625d7069cab8ac668a3f)'s `pyproject.toml` and `.flake8`, but introduced some error :(
> 
> will fix it tomorrow or weekend, or if you have time feel free to fix that

I'll have a look now. Is there a reason you don't want to use local installations in pre-commit? It's faster and naturally integrates with poetry.

Also, ruff supersedes flake8, so there should be no .flake8 file anymore, I believe

MischaPanch:
The error you introduced is related to #922. You removed SIM118 from the ignored rules, so ruff replaced all `for key in batch.keys()` by `for key in batch`. I will undo this

MischaPanch:
There is some problem with pip install on windows now, probably some unversioned dependency got bumped and now causes encoding issues. This could in principle be overcome by setting the python encoding as described [here](https://stackoverflow.com/questions/49640513/unicodedecodeerror-charmap-codec-cant-decode-byte-0x9d-in-position-x-charac), but then one needs an `if-else` depending on the OS in the CI, and it really gets messy.

I think there are good chances that this problem disappears when moving the build to poetry in the next PR. So I propose to ignore it for the moment, merge this, and reopen the previous PR. Problems like this one are also one more reason to switch to poetry :). The CI will likely become quite a bit faster after that as well

Trinkle23897:
> Is there a reason you don't want to use local installations in pre-commit? It's faster and naturally integrates with poetry.

nope, feel free to revert it back

> Also, ruff supersedes flake8, so there should be no .flake8 file anymore, I believe

it's not related to ruff check. it's because I currently use vscode with flake8 plugin, and it's really annoying that flake8 doesn't respect ruff's config

MischaPanch:
> > Is there a reason you don't want to use local installations in pre-commit? It's faster and naturally integrates with poetry.
> 
> nope, feel free to revert it back

Will do in the next PR

> > Also, ruff supersedes flake8, so there should be no .flake8 file anymore, I believe
> 
> it's not related to ruff check. it's because I currently use vscode with flake8 plugin, and it's really annoying that flake8 doesn't respect ruff's config

There is also a ruff plugin ;). We could add `.flake8` to gitignore, then everyone can have it in their local repo. I think it's confusing to check it in when it's in fact unused.

Trinkle23897:
> There is also a ruff plugin ;). We could add .flake8 to gitignore, then everyone can have it in their local repo. I think it's confusing to check it in when it's in fact unused.

sgtm

> There is some problem with pip install on windows now

I don't quite get [93f0815](https://github.com/thu-ml/tianshou/pull/921/commits/93f0815af0288cb6c22f0dcef1b183fcf8ba4932) is succeed but [73e49c6](https://github.com/thu-ml/tianshou/pull/921/commits/73e49c65ff71738f13c3625d7069cab8ac668a3f) is fail; let me rerun 93f0815's windows ci again to double check

Trinkle23897:
I fairly believe it's setup.py issue: https://github.com/thu-ml/tianshou/actions/runs/5969625005/job/16221279168

MischaPanch:
> I fairly believe it's setup.py issue: https://github.com/thu-ml/tianshou/actions/runs/5969625005/job/16221279168

Ah, I see now. I guess ruff added these lines without asking... Or did you?

```
def get_readme() -> str:
    with open("README.md") as f:
        return f.read()
```

The open needs a utf8 encoding. I'm on the move, wanna add it?

Alternatively, just merge since setup.py will be deleted in next PR anyway

Trinkle23897:
yes that's my change, sorry about that, please add it back and many thanks!

Trinkle23897 (approved):

PR #925, opened 2023-08-27
Poetry install, remove gym, bump python

### LINKED ISSUE #914: Modernize build, dependencies and types

MischaPanch:
Closes #914 

Additional changes:

- Deprecate python below 11
- Remove 3rd party and throughput tests. This simplifies install and test pipeline
- Remove gym compatibility and shimmy
- Format with 3.11 conventions. In particular, add `zip(..., strict=True/False)` where possible

Since the additional tests and gym were complicating the CI pipeline (flaky and dist-dependent), it didn't make sense to work on fixing the current tests in this PR to then just delete them in the next one. So this PR changes the build and removes these tests at the same time.


MischaPanch:
@Trinkle23897 should be ready for review. I just hope the ray tests don't fail on windows...

MischaPanch:
Note that CI is way faster now, as the dependencies are cached and also python 3.11 is just faster. Should get even better once torch 2.0.2 is out

Trinkle23897:
Oh I think you need to test all instead of only test/continuous

MischaPanch:
> Oh I think you need to test all instead of only test/continuous

I took the command from the current version on master, where it was declared in the CI workflow as
`pytest test/base test/continuous --cov=tianshou --durations=0 -v --color=yes`

I was wondering myself why only part was tested on non-ubuntu

Trinkle23897:
No this is only for non-Ubuntu, for Ubuntu we use another command

MischaPanch:
> No this is only for non-Ubuntu, for Ubuntu we use another command

Ah, sorry, that was a typo, pushed the fix.

Now something is failing on windows in an async test and I have no idea how to debug it and what causes the error :( 

Gonna try to find a windows machine next week. Unless you know what's going on or want to fix it?

EDIT: my bet is on ray causing issues

Trinkle23897:
oh actually we can have fast ci and slow ci together, and mark fast ci as required, slow-ci as non-required, so that a single ci run can only take 10min

MischaPanch:
> oh actually we can have fast ci and slow ci together, and mark fast ci as required, slow-ci as non-required, so that a single ci run can only take 10min

Gonna address this in a follow-up PR.

Meanwhile, I couldn't stop myself from solving the issue with ray and found that pip, for whatever reason, **uninstalled it on windows after successfully installing it!**

See [here](https://github.com/thu-ml/tianshou/actions/runs/5980673629/job/16227258702), at the end of install dependencies

```
Found existing installation: ray 2.6.3
Uninstalling ray-2.6.3:
  Successfully uninstalled ray-2.6.3
```

I have no idea why it behaves this way... I will now remove installation of ray for windows in the pyproject, should fix the test and finally turn CI green

MischaPanch:
That did the trick

Trinkle23897:
> Successfully uninstalled ray-2.6.3

So which version did it finally install? My best guess would be some other projects on windows specified ray version for their own dependency

MischaPanch:
> > Successfully uninstalled ray-2.6.3
> 
> So which version did it finally install? My best guess would be some other projects on windows specified ray version for their own dependency

None, seems like it never installed ray in CI in windows, so the error never manifested itself. When I switched to poetry, ray was actually installed, so the ray-based env was added to tests and failed on windows.

Ray is flaky to install on windows, there are many issues about it. I guess users figured it out depending on their particular setting. Maybe there weren't many ray users on windows at all. Probably none on windows server 22 anyway

MischaPanch:
@Trinkle23897 from my side this is ready to merge. Do you have more comments on the changes?

Trinkle23897:
I'll take a look, thanks!

pyproject.toml
```diff
@@ -58,4 +147,27 @@ max-complexity = 20
...
+_ruff_format = "ruff --fix ."
+lint = ["_black_check", "_ruff_check"]
+_poetry_install_sort_plugin = "poetry self add poetry-plugin-sort"
+_poery_sort = "poetry sort"
+format = ["_black_format", "_ruff_format", "_poetry_install_sort_plugin", "_poery_sort"]
+_spellcheck = "sphinx-build -W -b spelling docs docs/_build"
+_doc_build = "sphinx-build -W -b html docs docs/_build"
+doc-clean = "rm -rf docs/_build"
+doc-build = ["_spellcheck", "_doc_build"]
+type-check = "mypy tianshou"
```
Trinkle23897, line 173: could you also update the corresponding parts in `docs/`? For example https://tianshou.readthedocs.io/en/master/contributing.html

Trinkle23897 (dismissed):

Trinkle23897:
> Format with 3.11 conventions. In particular, add zip(..., strict=True/False) where possible

Is it handled by ruff, or did you do this upgrade manually?

MischaPanch:
> > Format with 3.11 conventions. In particular, add zip(..., strict=True/False) where possible
> 
> Is it handled by ruff, or did you do this upgrade manually?

Had to do it manually, I guess ruff doesn't want to take the risk of breaking previous code. I also had to remove the check in some places b/c of an open issue in numba

Trinkle23897:
@MischaPanch should I wait until fixing https://github.com/thu-ml/tianshou/pull/925#discussion_r1315337450 or merge directly and create a new issue/pr for fixing doc issue?

MischaPanch:
> @MischaPanch should I wait until fixing [#925 (comment)](https://github.com/thu-ml/tianshou/pull/925#discussion_r1315337450) or merge directly and create a new issue/pr for fixing doc issue?

Let me fix the docs first, it's within the scope of this PR. I'll do it today

pyproject.toml
MischaPanch: Done. I haven't found any other part of docs that needs updating

MischaPanch:
The contributing.rst was updated. Any further updates can be done in subsequent PRs. Since there will generally be an overhaul of the docs, turning the rst files into executable notebooks (using something like nbconvert or jupyterbook), additional changes to docs can be done then.

@Trinkle23897 feel free to merge :)

Trinkle23897 (approved):

MischaPanch:
@Trinkle23897 I don't have the permissions to merge (or to review, for that matter). Do you wanna extend me to Maintainer? I could then also help reviewing community PRs

Trinkle23897:
granted

Issue #938, opened 2023-09-07
Create high level interfaces for config and experiments


MischaPanch:
Users should be able to quickly create a trainable policy without having to go through details of instantiating trainers, collectors, and so on. Once an env factory is written, starting the training should be easy. Loading the env and policy from a checkpoint should also be simple. Some inspiration can be taken from the way rllib or SB3 do this - we should just take the best parts/ideas from there though (if any).

This issue has probably one of the highest impact/effort ratios among the existing ones, as it will extend the target audience of tianshou. At least the following points should be fulfilled:
 

- [x] Config can be easily read and saved
- [x] Config can easily be manipulated from the command line
- [x] Config is saved with each experiment and agent **and env** are recreated from it at loading. Note that the recreation of custom envs from config is crucial. This means that custom env factories should be registered or at least passed at loading time.

The implementation will require at least

- [ ] Factoring out a bunch of utils from the example scripts
- [x] Removing the whole args stuff. In my current implementation (not in this repo), config, defaults and explanations are read off from the dataclasses using jsonargparse
- [x] Changing at least some of the example scripts to use the new structure. In fact, I believe that after this there might remain just a few script that are started using different yaml files.
- [x] Write at least some small documentation

A tangential, but I believe very important and useful feature would be: **compatibility with SB3 and rllib**.

What I mean by that is the following: the config determines the training entirely. SB3 and rllib also have high level interfaces, with their own way to configure stuff, their own names and ways to start the training. This complicates comparisons in benchmarking, and also makes users' lives more difficult, in case they want to switch from one lib to another. Note that many "industry" users will never want to go beyond the high level interfaces.

Just by mapping names to each other we could easily implement methods like `to_sb3_config` or `to_rllib_config` and even simple functions like `train_with_rllib(config)` and `train_with_sb3(config)`, in addition to the standard `train`. This will make reliable performance comparisons very simple, and is thus related to #935. For algorithms or options not implemented in other frameworks, one raises a `NotImplementedError`, but at least for the standard algorithms the overlap of what is implemented in all libraries is substantial.

@opcode81 Do you want to take a look at it? I will give you access to our fork and internal repositories where I have implemented parts of this with jsonarparse. I also included nni support for these configs, which should also make user's life easy.

**NOTE:**
We don't have to fully nail it at the beginning, and it also doesn't need to be exhaustive. Would be nice to cover the most commonly used algos in the beginning, say SAC, PPO and TD3.

[...]

PR #950, opened 2023-09-24
Remove kwargs in policy init

### LINKED ISSUE #947: Remove kwargs in policies init

MischaPanch:
Closes #947 

This removes all kwargs from all policy constructors. While doing that, I also improved several names and added a whole lot of TODOs.

## Functional changes:

1. Added possibility to pass None as `critic2` and `critic2_optim`. In fact, the default behavior then should cover the absolute majority of cases
2. Added a function called `clone_optimizer` as a temporary measure to support passing `critic2_optim=None` 

## Breaking changes:

1. `action_space` is no longer optional. In fact, it already was non-optional, as there was a ValueError in BasePolicy.init. So now several examples were fixed to reflect that
2. `reward_normalization` removed from DDPG and children. It was never allowed to pass it as `True` there, an error would have been raised in `compute_n_step_reward`. Now I removed it from the interface
3. renamed `critic1` and similar to `critic`, in order to have uniform interfaces. Note that the `critic` in DDPG was optional for the sole reason that child classes used `critic1`. I removed this optionality (DDPG can't do anything with `critic=None`)
4. Several renamings of fields (mostly private to public, so backwards compatible)

## Additional changes: 
1. Removed type and default declaration from docstring. This kind of duplication is really not necessary
2. Policy constructors are now only called using named arguments, not a fragile mixture of positional and named as before
5. Minor beautifications in typing and code 
6. Generally shortened docstrings and made them uniform across all policies (hopefully)

## Comment:

With these changes, several problems in tianshou's inheritance hierarchy become more apparent. I tried highlighting them for future work.


MischaPanch:
@Trinkle23897 this is a next big step towards more robust and maintainable code, it will require your attention.

These changes required some serious, soul-crushing grinding on the whole code base :cry: . Kwargs are such an insidious antipattern, once it's there, it's really hard to remove.

I described all changes that could be somehow controversial above, the rest is just pushing args around and adjusting everything after.

@opcode81 The changes will probably slightly affect the high level interfaces, e.g., the removal of `reward_normalization` for DDPG and children. Should make it easier for you to directly see which configuration a policy can take

MischaPanch:
Ah, one more thing @Trinkle23897 : I think it would be better not to squash the commits on this PR, so that eventual mistakes can be caught more easily. I tried to structure the commits in a sensible way, which should be much better for transparency than a single, huge commit over 2k lines of code

Trinkle23897:
Actually, could you do the following two things before this PR?
1. Support mypy for all files, including examples and tests
2. Change function signature to 
```
def func(
  self,
  *,  # add this `*` to force use kwargs
  ...
) -> ...:
  pass
```

MischaPanch:
Good point about the forcing of kwargs! Wish it'd been done earlier ^^

I'll do it in this PR and not in a separate one since it would create lot's of conflicts and duplicated work otherwise

> Actually, could you do the following two things before this PR?
> 
> 1. Support mypy for all files, including examples and tests
> 2. Change function signature to
> 
> ```
> def func(
>   self,
>   *,  # add this `*` to force use kwargs
>   ...
> ) -> ...:
>   pass
> ```

MischaPanch:
@Trinkle23897 mypy on examples and tests finds a total of over 1000 typing violations. I would strongly prefer to handle that in a separate PR, maybe multiple ones.

I created #952 for that

MischaPanch:
@Trinkle23897 this is ready for review/merge. Since it will create conflicts with almost any other PR, would be great to merge it soon :)

Trinkle23897:
I'll take a carefully view

examples/atari/atari_c51.py
```diff
@@ -88,13 +88,14 @@ def test_c51(args=get_args()):
...
     # define policy
     policy = C51Policy(
-        net,
-        optim,
-        args.gamma,
-        args.num_atoms,
-        args.v_min,
-        args.v_max,
-        args.n_step,
+        model=net,
```
Trinkle23897, line 91: need to modify the following parts together:
https://github.com/thu-ml/tianshou/blob/6449a43261eeea82185e07dd659a79afff9556ab/README.md?plain=1#L227
https://github.com/thu-ml/tianshou/blob/6449a43261eeea82185e07dd659a79afff9556ab/docs/tutorials/dqn.rst?plain=1#L153
https://github.com/thu-ml/tianshou/blob/6449a43261eeea82185e07dd659a79afff9556ab/docs/tutorials/tictactoe.rst?plain=1#L314
https://github.com/thu-ml/tianshou/blob/6449a43261eeea82185e07dd659a79afff9556ab/docs/tutorials/tictactoe.rst?plain=1#L329

examples/offline/d4rl_cql.py
```diff
@@ -141,8 +141,8 @@ def test_cql():
     policy = CQLPolicy(
         actor,
         actor_optim,
-        critic1,
-        critic1_optim,
+        critic,
```
Trinkle23897: missing kwargs and action_space?

pyproject.toml
```diff
@@ -101,7 +102,7 @@ warn_redundant_casts = true
 warn_unreachable = true
 warn_unused_configs = true
 warn_unused_ignores = true
-exclude = "^build/|^docs/|^test/|^examples/"
```
Trinkle23897, line 105: oh wow does it work now?

test/base/test_collector.py
```diff
@@ -27,13 +27,13 @@
 
 
 class MyPolicy(BasePolicy):
-    def __init__(self, dict_state=False, need_state=True, action_shape=None):
+    def __init__(self, action_space=None, dict_state=False, need_state=True, action_shape=None):
```
Trinkle23897: Won't action_space=None raise ValueError in BasePolicy? How come it currently passed all tests?

test/base/test_env_finite.py
```diff
@@ -164,6 +164,9 @@ class FiniteSubprocVectorEnv(FiniteVectorEnv, SubprocVectorEnv):
 
 
 class AnyPolicy(BasePolicy):
+    def __init__(self):
+        super().__init__(action_space=None)
```
Trinkle23897: same here as above

test/offline/gather_pendulum_data.py
```diff
@@ -118,16 +116,15 @@ def gather_data():
...
     policy = SACPolicy(
-        actor,
-        actor_optim,
-        critic1,
-        critic1_optim,
-        critic2,
-        critic2_optim,
+        actor=actor,
+        actor_optim=actor_optim,
+        critic=critic1,
```
Trinkle23897: rename critic1 to critic?

test/offline/test_bcq.py
```diff
@@ -149,14 +149,15 @@ def test_bcq(args=get_args()):
...
-        actor_optim,
-        critic1,
-        critic1_optim,
-        critic2,
-        critic2_optim,
-        vae,
-        vae_optim,
+        actor_perturbation=actor,
+        actor_perturbation_optim=actor_optim,
+        critic=critic1,
```
Trinkle23897: same as here?

test/offline/test_cql.py
```diff
@@ -139,12 +139,12 @@ def test_cql(args=get_args()):
...
     policy = CQLPolicy(
-        actor,
-        actor_optim,
-        critic1,
-        critic1_optim,
-        critic2,
-        critic2_optim,
+        actor=actor,
+        actor_optim=actor_optim,
+        critic=critic1,
```
Trinkle23897: same?

tianshou/data/buffer/her.py
```diff
@@ -15,13 +15,13 @@ class HERReplayBuffer(ReplayBuffer):
     observation is a dictionary with keys ``observation``, ``achieved_goal`` and
     ``desired_goal``. Currently support only HER's future strategy, online sampling.
 
-    :param int size: the size of the replay buffer.
+    :param size: the size of the replay buffer.
```
Trinkle23897, line 18: would you mind changing the docstring style to google one next time?

tianshou/policy/imitation/discrete_crr.py
```diff
@@ -1,34 +1,36 @@
...
+    :param beta: when policy_improvement_mode is "exp", this is the denominator
         of the exp function. Default to 1.
-    :param float min_q_weight: weight for CQL loss/regularizer. Default to 10.
-    :param int target_update_freq: the target network update frequency (0 if
+    :param min_q_weight: weight for CQL loss/regularizer. Default to 10.
+    :param target_update_freq: the target network update frequency (0 if
         you do not use the target network). Default to 0.
-    :param bool reward_normalization: normalize the reward to Normal(0, 1).
+    :param reward_normalization: normalize the reward to Normal(0, 1).
         Default to False.
```
Trinkle23897: missing docstring? action_bound_method for example

tianshou/policy/imitation/td3_bc.py
```diff
@@ -1,49 +1,46 @@
...
-        Default to "clip".
-    :param Optional[gym.Space] action_space: env's action space, mandatory if you want
-        to use option "action_scaling" or "action_bound_method". Default to None.
-    :param lr_scheduler: a learning rate scheduler that adjusts the learning rate in
-        optimizer in each policy.update(). Default to None (no lr_scheduler).
+    :param action_scaling: whether to map actions from range [-1, 1]
+        to range[action_spaces.low, action_spaces.high].
+    :param action_bound_method: method to bound action to range [-1, 1],
+        can be either "clip" (for simply clipping the action)
+        or empty string for no bounding. Only used if the action_space is continuous.
```
Trinkle23897: or `None` for no bounding

tianshou/policy/modelfree/a2c.py
```diff
@@ -9,42 +10,32 @@
...
+    :param max_grad_norm: clipping gradients in back propagation.
+    :param gae_lambda: in [0, 1], param for Generalized Advantage Estimation.
+    :param max_batchsize: the maximum size of the batch when computing GAE.
+    :param discount_factor: in [0, 1].
+    :param reward_normalization: normalize estimated values to have std close to 1.
+    :param deterministic_eval: if True, use deterministic evaluation.
+    :param observation_space: the space of the observation.
+    :param action_scaling: if True, scale the action from [-1, 1] to the range of
+        action_space. Only used if the action_space is continuous.
+    :param action_bound_method: method to bound action to range [-1, 1].
```
Trinkle23897, line 37: missing `Only used if the action_space is continuous.`?

tianshou/policy/imitation/td3_bc.py
Trinkle23897: missing obsevation_space?

tianshou/policy/modelbased/psrl.py
```diff
@@ -163,15 +172,26 @@ class PSRLPolicy(BasePolicy):
...
+        observation_space: gym.Space | None = None,
+        action_scaling: bool = False,
+        action_bound_method: Literal["clip", "tanh"] | None = "clip",
+        lr_scheduler: TLearningRateScheduler | None = None,
     ) -> None:
-        super().__init__(**kwargs)
+        super().__init__(
+            action_space=action_space,
+            observation_space=observation_space,
+            action_scaling=action_scaling,
```
Trinkle23897: can actually set action_scaling=False and action_bound_method=None here since psrl can be only applied on discrete case

tianshou/policy/modelfree/discrete_sac.py
```diff
@@ -1,37 +1,38 @@
...
+    :param critic2: the second critic network. (s, a -> Q(s, a)).
+        If None, use the same network as critic (via deepcopy).
+    :param critic2_optim: the optimizer for the second critic network.
+        If None, clone critic_optim to use for critic2.parameters().
+    :param tau: param for soft update of the target network.
+    :param gamma: discount factor, in [0, 1].
+    :param alpha: entropy regularization coefficient.
+        If a tuple (target_entropy, log_alpha, alpha_optim) is provided,
+        then alpha is automatically tuned.
+    :param lr_scheduler: a learning rate scheduler that adjusts the learning rate
```
Trinkle23897, line 36: missing observation_space?

tianshou/policy/modelfree/pg.py
```diff
@@ -20,69 +23,75 @@
...
+        of action_space. Only used if the action_space is continuous.
+    :param action_bound_method: method to bound action to range [-1, 1].
+        Only used if the action_space is continuous.
+    :param lr_scheduler: if not None, will be called in `policy.update()`.
 
     .. seealso::
 
-        Please refer to :class:`~tianshou.policy.BasePolicy` for more detailed
-        explanation.
+        Please refer to :class:`~tianshou.policy.BasePolicy` for more detailed explanation.
```
Trinkle23897, line 51: could you double-check if the rendered html doc has any formatting issues locally?

tianshou/policy/modelfree/redq.py
```diff
@@ -55,72 +47,86 @@ class REDQPolicy(DDPGPolicy):
...
+            self.target_entropy, self.log_alpha, self.alpha_optim = alpha
+            self.alpha = self.log_alpha.detach().exp()
+        else:
+            # TODO: make mypy undestand this, or switch to something like pyright...
+            alpha = cast(float, alpha)
+            self.alpha = alpha
+
+    @property
+    def is_auto_alpha(self) -> bool:
+        return self._is_auto_alpha
```
Trinkle23897, line 125: why not just replace `_is_auto_alpha` to `is_auto_alpha`?

tianshou/policy/modelfree/sac.py
```diff
@@ -9,42 +10,42 @@
...
+        "default" is equivalent to GaussianNoise(sigma=0.1).
+    :param deterministic_eval: whether to use deterministic action
+        (mean of Gaussian policy) in evaluation mode instead of stochastic
+        action sampled by the policy. Does not affect training.
+    :param action_scaling: whether to map actions from range [-1, 1]
+        to range[action_spaces.low, action_spaces.high].
+    :param action_bound_method: method to bound action to range [-1, 1],
+        can be either "clip" (for simply clipping the action)
+        or empty string for no bounding. Only used if the action_space is continuous.
+    :param lr_scheduler: a learning rate scheduler that adjusts the learning rate
```
Trinkle23897, line 48: missing observation_space?

tianshou/policy/modelfree/td3.py
```diff
@@ -1,48 +1,46 @@
...
+        This is useful when solving "hard exploration" problems.
+        "default" is equivalent to GaussianNoise(sigma=0.1).
+    :param policy_noise: the noise used in updating policy network.
+    :param update_actor_freq: the update frequency of actor network.
+    :param noise_clip: the clipping range used in updating policy network.
+    :param action_scaling: whether to map actions from range [-1, 1]
+        to range[action_spaces.low, action_spaces.high].
+    :param action_bound_method: method to bound action to range [-1, 1],
+        can be either "clip" (for simply clipping the action)
+        or empty string for no bounding. Only used if the action_space is continuous.
```
Trinkle23897: same missing observation_space?

tianshou/utils/optim.py
```diff
@@ -20,3 +23,42 @@ def optim_step(
...
+    torch.optim.Adam,
+    torch.optim.SGD,
+    torch.optim.RMSprop,
+    torch.optim.Adadelta,
+    torch.optim.AdamW,
+    torch.optim.Adamax,
+    torch.optim.NAdam,
+    torch.optim.SparseAdam,
+    torch.optim.LBFGS,
+]
```
Trinkle23897, line 38: Can we use their parent class instead of hardcoding things here? Otherwise, if pytorch has an upgrade that changes some of them, it would silently fail on our side.

tianshou/utils/optim.py
```diff
@@ -20,3 +23,42 @@ def optim_step(
...
+
+TOptim = TypeVar("TOptim", bound=torch.optim.Optimizer)
+
+
+def clone_optimizer(
```
Trinkle23897, line 43: why do we need that?

tianshou/utils/optim.py
```diff
@@ -20,3 +23,42 @@ def optim_step(
     if max_grad_norm:
```
Trinkle23897, line 23: this `optim_step` is not covered by any test

pyproject.toml
MischaPanch: Nope, but at least you can run `mypy tests` and `mypy examples` to find the 1k+ typing issues. `poe type-check`, CI and pre-commit still run `mypy tianshou`

test/base/test_collector.py
MischaPanch: Yeah, I didn't understand that either, but also didn't bother going through the tests. Maybe this class is never instantiated?

tianshou/data/buffer/her.py
MischaPanch: If there is a reliable docstring converter, it could be done. But why? Is there a large benefit in it?

This would need to adjust the sphinx build, and maybe also things like doc8 config

tianshou/policy/imitation/discrete_crr.py
MischaPanch: I tried my best to keep the docstrings correct, but there are probably still several incomplete ones... Gonna address the ones you found

tianshou/policy/modelfree/redq.py
MischaPanch: Because the user doesn't pass `is_auto_aplha`, it is inferred from the `alpha` input (if it's a tuple or a constant). So the user code should never be able to mutate it, hence the property

tianshou/utils/optim.py
MischaPanch: I purposedly didn't use the base class because the clone_optimizer method might not work for user-defined optimizers

tianshou/utils/optim.py
MischaPanch: Because I relaxed the interface to pass `critic2=None, critic2_optim=None`. The default behavior is then to copy `critic` and `critic_optim`. The latter cannot be copied, hence this function. Note that this default behavior should cover the majority of use cases, and I believe it covers all examples and tests, even though it's not used there atm.

Generally, I think we should bundle the nets and optimizers in a single abstraction, or let users pass optimizer factories instead. Currently, we always rely on the user to have a correct optimizer, i.e. we hope that `critic_optim` was indeed created with `critic.parameters()`. This is not verified at runtime, and if it doesn't hold, the algorithms won't work, of course.

tianshou/utils/optim.py
MischaPanch: Yes, it slipped in with a previous merge, where I have forgotten to delete it. I use it on other branches in my fork, and plan to include tests for it and use it across policies (this code is duplicated many times over). Can we keep it for the time being? It is currently unused

test/base/test_collector.py
Trinkle23897: could you add a breakpoint when executing test?

tianshou/utils/optim.py
Trinkle23897: sure

Trinkle23897:
Oh the dumb thing: https://github.com/thu-ml/tianshou/blob/6449a43261eeea82185e07dd659a79afff9556ab/tianshou/policy/base.py#L107
missing `raise`

MischaPanch:
> Oh the dumb thing:
> 
> https://github.com/thu-ml/tianshou/blob/6449a43261eeea82185e07dd659a79afff9556ab/tianshou/policy/base.py#L107
> 
> 
> missing `raise`

How did me and my IDE miss this :( .

I was busy lately but will continue on this PR tomorrow and address your comments then

examples/offline/d4rl_cql.py
MischaPanch: Done

test/base/test_env_finite.py
MischaPanch: Done

test/base/test_collector.py
MischaPanch: Done (added the raise and passed an action space here)

test/offline/gather_pendulum_data.py
MischaPanch: Done, also removed explicit construction of second critic and optim, so the new default behavior is covered by tests

test/offline/test_bcq.py
MischaPanch: Done, as above

test/offline/test_cql.py
MischaPanch: Done

tianshou/utils/optim.py
opcode81: Indeed, I don't think this is needed (or elegant). The algorithm requires both critics and if the user wants to use the same type of critic twice, he can create two identical instances. This is more appropriate for a low-level interface than a perhaps contrived cloning mechanism.
In the high-level interface, which is currently being developed (#938), the user will be offered more convenient interfaces anyway (involving factories). 
For the more low-level interface, I would prefer to keep it simple (and true to the actual requirements) and rather remove this.

tianshou/policy/imitation/td3_bc.py
MischaPanch: Made docstrings for these params the same as in DDPG

tianshou/policy/modelbased/psrl.py
MischaPanch: Done

tianshou/policy/modelfree/discrete_sac.py
MischaPanch: Done

tianshou/policy/modelfree/sac.py
MischaPanch: Done

tianshou/policy/modelfree/td3.py
MischaPanch: Done

tianshou/utils/optim.py
MischaPanch: I disagree, AFAIK in virtually all cases in tianshou the `net_2`, `critic2`, and `critic_optim2` are covered by the default I introduced here. This will also be true for the absolute majority of users, I don't see the point in forcing them to create a second net/critic/optim if there is such a sensible default behavior. Of course, they are still able to do so if desired.

In my opinion, the main problem manifested here is that the nets and their optimizers are passed separately, as I outlined above, requiring this hacky and non-elegant solution. I view it as temporary step, until the "bundling" of optimizers and nets is improved - either by new abstractions of by switching to passing optim factories (like already done in many other torch-based libraries)

MischaPanch: I am also not aware of any recommendation from research about using different model architectures or learning rates for the dual-critic algorithms. There might be some exotic cases, of course

opcode81: IMO, clarity and simplicity should be prioritized over ease of use in a low-level interface. The high-level interface will solve the usability issues, so there is not really a good reason to solve some of them twice (in different ways).
It is just clearer to require exactly what the algorithm actually requires in the interface than to have a default of `None` which needs to be explained (and requires additional code to be handled).

opcode81: I would be more inclined to agree with your notion if there was no high-level interface in development, mind you.
If you continue down this path, the low-level interface will, more and more, become a high-level interface which focuses on usability rather than simplicity and clarity, introducing unnecessary complexity.

examples/atari/atari_c51.py
MischaPanch: Done

tianshou/utils/optim.py
opcode81: There is value in maintaining a high degree of simplicity in the core (low-level) part of the library ([KISS](https://www.baeldung.com/cs/kiss-software-design-principle)).

MischaPanch: I don't think this one change leads to a slippery slope ;). The policy interface in a way is a high-level interface, and I suspect that researchers will continue using it as main entry point for a while.

However, I do see how a bundling of nets and optimizers (or use of factories) in the future might increase complexity if not done with much care, so I'm not 100% sure either.

I think it's a general design decision here - should we put some weight on increasing the usability of the current interfaces, despite the coming addition of high level interfaces. My feeling is that researches will mainly rely on the current interfaces, while the new ones will be useful for newcomers and in applications. This is b/c researchers who develop algorithms will usually create a new policy by inheriting from existing ones. So in my opinion, potentially sacrificing a bit of explicitness for covering this stark default behavior is a positive development. I also view the fact that optimizers are passed explicitly (instead of through factories or abstractions) as technical debt, and not necessarily improving on simplicity of the core library. But I understand that it's not clear-cut

@Trinkle23897 your call to give the decisive vote.

tianshou/utils/optim.py
opcode81: This already indicates that cloning the optimizer might not be the best approach in the first place.

tianshou/policy/modelfree/pg.py
MischaPanch: I went over it, looks good to me, links are resolved

tianshou/policy/modelfree/a2c.py
MischaPanch: Done

tianshou/policy/imitation/discrete_crr.py
MischaPanch: Done

tianshou/utils/optim.py
Trinkle23897: I'm more inclined toward @opcode81. I agree with @MischaPanch that we should make a better interface, and I think 
> I view it as temporary step, until the "bundling" of optimizers and nets is improved - either by new abstractions of by switching to passing optim factories (like already done in many other torch-based libraries)

is the right direction we should push forward. As an intermediate step, if we don't care about clean/elegant, then the current stuff is totally fine; however, if we do care about it in this intermediate step, I tend to revert the optimizer change in this file, for two reasons:
- we will refactor this whole interface anyway in the future
- it's confusing people if the algorithm uses two critics/optimizers but the entrypoint only passes one critic and optimizer (that's why at the beginning I made the entrypoint to explicitly pass two critics instead of one)

Trinkle23897 (approved):
Could you trigger another ci run to make sure everything looks good? If gpu test still fails, I'll take it anyway.

GPU test is flaky because it's in mainland China, which stays inside the GFW and needs "help" connecting to github stablely.

MischaPanch:
> Could you trigger another ci run to make sure everything looks good? If gpu test still fails, I'll take it anyway.
> 
> GPU test is flaky because it's in mainland China, which stays inside the GFW and needs "help" connecting to github stablely.

CI tests passed on penultimate commit (except the GPU runner), and the last one was just an additional line in a docstring. Didn't seem worth retriggering the long tests over it.

@Trinkle23897 feel free to merge at your convenience, but I really recommend to add all commits and not squash them here. I enabled that option in the repo's settings

tianshou/utils/optim.py
MischaPanch: Since atm it's clearly marked as intermediate step and not to be used directly, I think it's fine to keep it for the time being. We should continue the discussion in a separate setting, probably only after the first version of the high-level interfaces has been merged. @opcode81 is preparing a Wiki entry for it. I will create an issue linking to this thread, so we keep it in mind.

tianshou/utils/optim.py
MischaPanch: See convo below, resolving

Trinkle23897:
> I really recommend to add all commits and not squash them here

It’s hard to do bisect in that way. You can always first do bisect on master branch, and if you find this PR is causing the issue, you can checkout this branch and do another round of bisect.

opcode81:
> > I really recommend to add all commits and not squash them here
> 
> It’s hard to do bisect in that way. You can always first do bisect on master branch, and if you find this PR is causing the issue, you can checkout this branch and do another round of bisect.

@Trinkle23897 it's not just about bisecting. But bisecting is rendered rather difficult if the original branch no longer exists (deleted upon merge) and isn't even in the same repository.

Importantly, with the original commits intact, 
 * you get much more information on the individual changes and can better trace why certain changes were made (e.g. when using *blame* on a file) - provided that each commit has a meaningful commit message.
 * you can choose to revert only individual commits later.

Also, my branch for the high-level API actually built on this branch, because it provided important foundations, so I will now need to rebuild my branch based on the new, rewritten history. It is not uncommon for branches to build upon each other, and rewriting the history along the way causes problems that could otherwise be avoided. If it had just been merged, I would only have needed to merge the updated master.

For these reasons, I generally advise against squashing. The only thing squashing has going for it is that it makes the history shorter/simpler. This is, in most cases, far less important than retaining more information and a seamless workflow when branches depend upon each other.

Trinkle23897:
I'm sorry that caused some inconvenience for the development experience. In the future if I give a stamp, feel free to merge PR by yourself @MischaPanch (I believe you have the merge permission) in your own way.

MischaPanch:
> I'm sorry that caused some inconvenience for the development experience. In the future if I give a stamp, feel free to merge PR by yourself @MischaPanch (I believe you have the merge permission) in your own way.

@Trinkle23897 I have fewer permissions than one might expect for a maintainer status. E.g., I'm not allowed to merge when the pipeline fails (as is the case most of the time b/c of the GPU runner), I also cannot change the way in which merge is performed (i.e., not squash) for forks. Tbh, I'm not even sure I can push to branches in tianshou, since so far GH rejected such pushes - but maybe something is wrong in my ssh settings.

I'm happy to take more responsibility for how things are merged, but I'd need the permissions for that. Maybe there are some settings that should be adjusted?

Trinkle23897:
done, please check (previously you have "Maintain")
Screenshot 2023-10-09 at 12 06 38 PM

MischaPanch:
Thank you! I promise to be responsible with the new powers :)

I will make a mini-PR now, supporting slightly older versions of gymnasium (currently tianshou requirements are in conflict with rllib, that's why), and a minor change to the readme. Since you're online, gonna ask you for a quick glance, but in future I might merge such minor changes directly without review, if you don't mind

Trinkle23897:
Thanks! I'll make sure in the future I can give review for small-enough PR (diff < 200 lines) within 24h.

PR #961, opened 2023-10-09
Minor dep update

MischaPanch:
Support gymnasium >=0.28

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,7 +27,7 @@ exclude = ["test/*", "examples/*", "docs/*"]
 
 [tool.poetry.dependencies]
 python = "^3.11"
-gymnasium = "^0.29.0"
+gymnasium = "^0.28.0"
 h5py = "^3.9.0"
 numba = "^0.57.1"
 numpy = "^1"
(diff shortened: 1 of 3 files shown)


MischaPanch:
@Trinkle23897 The support of older gymnasium should be uncontroversial, so it's mainly the readme which needs a look. I modified the top-level alert, added a second one, and also added appliedAI to the acknowledgments section. Maybe the statement about long-term support is attractive for some users :)

MischaPanch:
Torch and poetry still don't like each other, even in torch 2.1.0 :( 

I added a section in the readme on how to install torch until they fix this

pyproject.toml
```diff
@@ -36,7 +36,7 @@ packaging = "*"
 pettingzoo = "^1.22"
 tensorboard = "^2.5.0"
 # Torch 2.0.1 causes problems, see https://github.com/pytorch/pytorch/issues/100974
-torch = "^2.0.0, !=2.0.1"
+torch = "^2.0.0, !=2.0.1, !=2.1.0"
```
Trinkle23897, line 39: may i know why?

Trinkle23897 (approved):

pyproject.toml
MischaPanch: See my message above and the failing CI tests

MischaPanch: And the new link in the readme

MischaPanch:
Woops, I clicked the button for auto-merge after checks pass, but it merged immediately for some reason. Sorry for that!

If CI does fail for some unexpected reason, I'll fix it immediately

Trinkle23897:
np

PR #962, opened 2023-10-11
Update mujoco_sac.py

bitsandscraps:
    + [X] exception-raising fix
    + [ ] new feature

Fixed the example file to make it compatible with the new parameter names of SACPolicy.


Trinkle23897:
We change it recently: https://github.com/thu-ml/tianshou/blob/66b7fc542b496090e83d2df4a846fc02f3f3167b/tianshou/policy/modelfree/sac.py#L62-L63

If you installed via `pip install -e .` it should work fine, however, if you installed via `pip install tianshou`, that will give you version 0.5.1, which is not exactly the current master version. Please check out `v0.5.1` to use the corresponding examples.

Issue #944, opened 2023-09-20
Remove symlinks in repo


opcode81:
Two files in the repo are symlinks:
* examples/vizdoom/network.py -> ../atari/atari_network.py
* examples/modelbased/psrl.py -> ../../test/modelbased/test_psrl.py

The use of symlinks can cause problems on Windows (git setting core.symlinks must be enabled and Windows Developer Mode must be used).
Unless there is a very good reason to have symlinks, they should be removed.

MischaPanch:
The symlinks are there for executing examples inside tests. I have the feeling that this mechanism should generally be adjusted, after which this issue may become obsolete. I'm putting it on hold for now, as it doesn't seem to be very urgent (given the limited number of devs on windows)

PR #970, opened 2023-10-17
High-Level API

### LINKED ISSUE #938: Create high level interfaces for config and experiments

opcode81:
    + [ ] exception-raising fix
    + [X] new feature

This PR closes #938. It introduces all the fundamental concepts and abstractions, and it already covers the majority of the algorithms. It is not a complete and finalised product, however, and we recommend that the high-level API remain in alpha stadium for some time, as already suggested in the issue.

The changes in this PR are described on a [wiki page](https://github.com/aai-institute/tianshou/wiki/High-Level-API), a copy of which is provided below. (The original page is perhaps more readable, because it does not render line breaks verbatim.)

# Introducing the Tianshou High-Level API

The new high-level library was created based on object-oriented design 
principles with two primary design goals:

 * **ease of use** for the end user (without sacrificing generality)

   This is achieved through:
     * a single, well-defined point of interaction (`ExperimentBuilder`)
       which uses declarative semantics, allowing the user to focus on 
       what to do rather than how to do it.

     * easily injectible parametrisation.
       
       For complex parametrisation involving objects, the respective 
       library classes are easily discoverable, keeping the need to 
       browse reference documentation - or, even worse, inspect code or class
       hierarchies - to an absolute minimium.

     * reduced points of failure. 
     
       Because the high-level API is at a higher level of abstraction, where
       more knowledge is available, we can centrally define reasonable 
       defaults and apply consistency checks in order to ensure that
       illegal configurations result in meaningful errors (and are completely
       avoided as long as the users does not modify default behaviour).
       For example, we can consider interactions between the nature of the
       action space and the neural networks being used.

 * **maintainability** for developers
   
   This is achieved through:
     * a modular design with strong separation of concerns
     * a high level of factorisation, which largely avoids duplication, 
       partly through the use of mixins and multiple inheritance. 
       This invariably makes the code slightly more complex, yet it greatly 
       reduces the lines of code to be written/updated, so it is a reasonable 
       compromise in this case.

## Changeset

The entire high-level library is in its own subpackage `tianshou.highlevel` 
and **almost no changes were made to the original library** in order to 
support the new APIs. 
For the most part, only typing-related changes were made, which have
aligned type annotations with existing example applications or have made 
explicit interfaces that were previously implicit.

Furthermore, some helper modules were added to the the `tianshou.util` package 
(all of which were copied from the [sensAI library](https://github.com/jambit/sensAI)).

Many example applications were added, based on the existing MuJoCo and Atari 
examples (see below).

### User Experience Example

To illustrate the UX, consider this video recording (IntelliJ IDEA):

UX

Observe how conveniently relevant classes can be discovered via the IDE's
auto-completion function.
Discoverability is markedly enhanced by using a prefix-based naming convention, 
where classes that can be used as parameters use the base class name as a prefix, 
allowing all potentially relevant subclasses to be straightforwardly 
auto-completed.

### Declarative Semantics

A key design principle for the user-facing interface was to achieve 
*declarative semantics*, where the user
is no longer concerned with generating a lengthy procedure that sequentially
constructs components that build upon each other. 
Instead, the user focuses purely on
*declaring* the properties of the learning task he would like to run.

  * This essentially reduces boiler-plate code to zero, as every part of the 
    code is defining essential, experiment-specific configuration.
  * This makes it possible to centrally handle interdependent configuration 
    and detect/avoid misspecification.

In order to enable the configuration of interdependent objects without 
requiring the user to instantiate the respective objects sequentially, we 
heavily employ the *factory pattern*.

### Experiment Builders

The end user's primary entry point is an `ExperimentBuilder`, which is
specialised for each algorithm.
As the name suggests, it uses the builder pattern in order to create
an `Experiment` object, which is then used to run the learning task.

 * At builder construction, the user is required to provide only essential 
   configuration, particularly the environment factory.
 * The bulk of the algorithm-specific parameters can be provided 
   via an algorithm-specific parameter object. 
   For instance, `PPOExperimentBuilder` has the method `with_ppo_params`,
   which expects an object of type `PPOParams`.
 * Parametrisation that requires the provision of more complex interfaces
   (e.g. were multiple specification variants exist) are handled via 
   dedicated builder methods. 
   For example, for the specification of the critic component in an
   actor-critic algorithm, the following group of functions is provided:
     * `with_critic_factory` (where the user can provide any (user-defined) 
       factory for the critic component)
     * `with_critic_factory_default` (with which the user specifies that
       the default, `Net`-based critic architecture shall be used and has the 
       option to parametrise it)
     * `with_critic_factory_use_actor` (with which the user indicates that the
       critic component shall reuse the preprocessing network from the actor 
       component)

##### Minimal Example

In the simplest of cases, where the user wants to use the default 
parametrisation for everything, a user could run a PPO learning task
as follows,

```python
experiment = PPOExperimentBuilder(MyEnvFactory()).build()
experiment.run()
```

where `MyEnvFactory` is a factory for the agent's environment.
The default behaviour will adapt depending on whether the factory 
creates environments with discrete or continuous action spaces.

##### Fully Parametrised MuJoCo Example 

Importantly, the user still has the option to configure all the details.
Consider this example, which is from the high-level version of the 
`mujoco_ppo` example:

```python
log_name = os.path.join(task, "ppo", str(experiment_config.seed), datetime_tag())

sampling_config = SamplingConfig(
    num_epochs=epoch,
    step_per_epoch=step_per_epoch,
    batch_size=batch_size,
    num_train_envs=training_num,
    num_test_envs=test_num,
    buffer_size=buffer_size,
    step_per_collect=step_per_collect,
    repeat_per_collect=repeat_per_collect,
)

env_factory = MujocoEnvFactory(task, experiment_config.seed, obs_norm=True)

experiment = (
    PPOExperimentBuilder(env_factory, experiment_config, sampling_config)
    .with_ppo_params(
        PPOParams(
            discount_factor=gamma,
            gae_lambda=gae_lambda,
            action_bound_method=bound_action_method,
            reward_normalization=rew_norm,
            ent_coef=ent_coef,
            vf_coef=vf_coef,
            max_grad_norm=max_grad_norm,
            value_clip=value_clip,
            advantage_normalization=norm_adv,
            eps_clip=eps_clip,
            dual_clip=dual_clip,
            recompute_advantage=recompute_adv,
            lr=lr,
            lr_scheduler_factory=LRSchedulerFactoryLinear(sampling_config)
            if lr_decay
            else None,
            dist_fn=DistributionFunctionFactoryIndependentGaussians(),
        ),
    )
    .with_actor_factory_default(hidden_sizes, torch.nn.Tanh, continuous_unbounded=True)
    .with_critic_factory_default(hidden_sizes, torch.nn.Tanh)
    .build()
)
experiment.run(log_name)
```

This is functionally equivalent to the procedural, low-level example.
Compare the scripts here:
 * [original low-level example](https://github.com/aai-institute/tianshou/blob/feat/high-level-api/examples/mujoco/mujoco_ppo.py)
 * [new high-level example](https://github.com/aai-institute/tianshou/blob/feat/high-level-api/examples/mujoco/mujoco_ppo_hl.py)

In general, find example applications of the high-level API in the `examples/`
folder in scripts using the `_hl.py` suffix:

 * [MuJoCo examples](https://github.com/aai-institute/tianshou/tree/feat/high-level-api/examples/mujoco)
 * [Atari examples](https://github.com/aai-institute/tianshou/tree/feat/high-level-api/examples/atari)

### Experiments

The `Experiment` representation contains
 * the agent factory ,
 * the environment factory,
 * further definitions pertaining to storage & logging.

An exeriment may be run several times, assigning a name (and corresponding
storage location) to each run.

#### Persistence and Logging

Experiments can be serialized and later be reloaded.

```python
    experiment = Experiment.from_directory("log/my_experiment")
```

Because the experiment representation is composed purely of configuration
and factories, which themselves are composed purely of configuration and
factories, persisted objects are compact and do not contain state.

Every experiment run produces the following artifacts:
 * the serialized experiment
 * the serialized best policy found during training
 * a log file
 * (optionally) user-defined data, as the persistence
   handlers are modular

Running a reloaded experiment can optionally resume training of the serialized
policy.

All relevant objects have meaningful string representations that can appear
in logs, which is conveniently achieved through the use of 
`ToStringMixin` (from sensAI). 
Its use furthermore prevents string representations of recurring objects 
from being printed more than once.
For example, consider this string representation, which was generated for
the fully parametrised PPO experiment from the example above:

```
Experiment[
    config=ExperimentConfig(
        seed=42, 
        device='cuda', 
        policy_restore_directory=None, 
        train=True, 
        watch=True, 
        watch_render=0.0, 
        persistence_base_dir='log', 
        persistence_enabled=True), 
    sampling_config=SamplingConfig[
        num_epochs=100, 
        step_per_epoch=30000, 
        batch_size=64, 
        num_train_envs=64, 
        num_test_envs=10, 
        buffer_size=4096, 
        step_per_collect=2048, 
        repeat_per_collect=10, 
        update_per_step=1.0, 
        start_timesteps=0, 
        start_timesteps_random=False, 
        replay_buffer_ignore_obs_next=False, 
        replay_buffer_save_only_last_obs=False, 
        replay_buffer_stack_num=1], 
    env_factory=MujocoEnvFactory[
        task=Ant-v4, 
        seed=42, 
        obs_norm=True], 
    agent_factory=PPOAgentFactory[
        sampling_config=SamplingConfig[<<], 
        optim_factory=OptimizerFactoryAdam[
            weight_decay=0, 
            eps=1e-08, 
            betas=(0.9, 0.999)], 
        policy_wrapper_factory=None, 
        trainer_callbacks=TrainerCallbacks(
            epoch_callback_train=None, 
            epoch_callback_test=None, 
            stop_callback=None), 
        params=PPOParams[
            gae_lambda=0.95, 
            max_batchsize=256, 
            lr=0.0003, 
            lr_scheduler_factory=LRSchedulerFactoryLinear[sampling_config=SamplingConfig[<<]], 
            action_scaling=default, 
            action_bound_method=clip, 
            discount_factor=0.99, 
            reward_normalization=True, 
            deterministic_eval=False, 
            dist_fn=DistributionFunctionFactoryIndependentGaussians[], 
            vf_coef=0.25, 
            ent_coef=0.0, 
            max_grad_norm=0.5, 
            eps_clip=0.2, 
            dual_clip=None, 
            value_clip=False, 
            advantage_normalization=False, 
            recompute_advantage=True], 
        actor_factory=ActorFactoryTransientStorageDecorator[
            actor_factory=ActorFactoryDefault[
                continuous_actor_type=ContinuousActorType.GAUSSIAN, 
                continuous_unbounded=True, 
                continuous_conditioned_sigma=False, 
                hidden_sizes=[64, 64], 
                hidden_activation=<class 'torch.nn.modules.activation.Tanh'>, 
                discrete_softmax=True]], 
        critic_factory=CriticFactoryDefault[
            hidden_sizes=[64, 64], 
            hidden_activation=<class 'torch.nn.modules.activation.Tanh'>], 
        critic_use_action=False], 
    logger_factory=LoggerFactoryDefault[
        logger_type=tensorboard, 
        wandb_project=None], 
    env_config=None]
```

## Library Developer Perspective

The presentation thus far has focussed on the user's perspective.
From the perspective of a Tianshou developer, it is important that the 
high-level API be clearly structured and maintainable.
Here are the most relevant representations:

 * **Policy parameters** are represented as dataclasses (base class `Params`).

   The goal is for the parameters to be ultimately passed to the corresponding
   policy class (e.g. `PPOParams` contains parameters for `PPOPolicy`).

   * **Parameter transformation**:
     In part, the parameter dataclass attributes already correspond directly to 
     policy class parameters.
     However, because the high-level interface must, in many cases, abstract away 
     from the low-level interface,
     we establish the notion of a `ParamTransformer`, which transforms
     one or more parameters into the form that is required by the policy class:
     The idea is that the dictionary representation of the dataclass is 
     successively transformed via `ParamTransformer`s such that the resulting
     dictionary can ultimately be used as keyword arguments for the policy.
     To achieve maintainability, the declaration of parameter transformations 
     is colocated with the parameters they affect.
     Tests ensure that naming issues are detected.

   * **Composition and inheritance**:
     We use inheritance and mixins to reduce duplication.

 * **Factories** are an essential principle of the library.
   Because the creation of objects may depend on objects that are not 
   yet created, a declarative approach necessitates that we transition from 
   the objects themselves to factories.

     * The `EnvFactory` was already mentioned above, as it is a user-facing
       abstraction.
       Its purpose is to create the (vectorized) `Environments` that will be 
       used in the experiments.
     * An `AgentFactory` is the central component that creates the policy,
       the trainer as well as the necessary collectors.
       To support a new type of policy, a subclass that handles the policy
       creation is required.
       In turn, the main task when implementing a new algorithm-specific 
       `ExperimentBuilder` is the creation of the corresponding `AgentFactory`.
     * Several types of factories serve to parametrize policies and training
       processes, e.g.
         * `OptimizerFactory` for the creation of torch optimizers
         * `ActorFactory` for the creation of actor models
         * `CriticFactory` for the creation of critic models
         * `IntermediateModuleFactory` for the creation of models that produce
           intermediate/latent representations
         * `EnvParamFactory` for the creation of parameters based on properties
           of the environment
         * `NoiseFactory` for the creation of `BaseNoise` instances
         * `DistributionFunctionFactory` for the creation of functions that
           create torch distributions from tensors
         * `LRSchedulerFactory` for learning rate schedulers
         * `PolicyWrapperFactory` for policy wrappers that extend the 
           functionality of the regular policy (e.g. intrinsic curiosity)
         * `AutoAlphaFactory` for automatically tuned regularization 
           coefficients (as supported by SAC or REDQ)
     * A `LoggerFactory` handles the creation of the experiment logger, 
       but the default implementation already handles the cases that were
       used in the examples.

  * The `ExperimentBuilder` implementations make use of mixins to add common
    functionality. As mentioned above, the main task in an algorithm-specific
    specialization is to create the `AgentFactory`.


MischaPanch:
@Trinkle23897 this is a major extension of tianshou, so it will require a thorough review from your side before being merged. As you see, @opcode81 was very thorough in documenting the many aspects and perspectives on these new features.

We already did some internal rounds of reviews and discussions on this, but will continue doing so next week, when I have more time. As promised, the changes are a pure addition (so @spacegoing @nuance1979 it's not creating additional engineering complexity for powerusers - quite the contrary).

Together with this PR, the wiki mentioned above should be migrated to this repository and the documentation (both readme and on readthedocs) should be extended. We believe that the new high-level interfaces will provide the main way for the majority of users to interact with tianshou.

I will ping you once we are done with internal reviews, but of course, if you already want to start a discussion - feel free :)

MischaPanch:
One thing to consider, is whether the [sensAI](https://github.com/jambit/sensAI) project (also mainly done by @opcode81, and actively developed and used at appliedAI) should eventually become a dependency of tianshou. In terms of transitive dependencies, it would only add scikit-learn, if I'm not mistaken.

sensAI is a high level library for general non-RL AI things, and over the years we developed several utilities that are useful in other contexts - in particular in tianshou. For now, however, these utilities were simply copied over. So it's a question for the future (though for the near future). sensAI was never marketed, and we haven't written sufficient documentation for it, but this is about to change, as appliedAI is interested in making it available to broader audience.

**TL;DR**: we borrowed sufficiently many utilities from another OSS project of ours, that it might be worth thinking about including it as dependency, instead of copy-pasting stuff

Trinkle23897:
I'll try my best to take a look this weekend, thanks for the work here!

examples/mujoco/mujoco_a2c_hl.py
```diff
@@ -0,0 +1,85 @@
...
+        .with_optim_factory(OptimizerFactoryRMSprop(eps=1e-5, alpha=0.99))
+        .with_actor_factory_default(hidden_sizes, nn.Tanh, continuous_unbounded=True)
+        .with_critic_factory_default(hidden_sizes, nn.Tanh)
+        .build()
+    )
+    experiment.run(log_name)
+
+
+if __name__ == "__main__":
+    logging.run_cli(main)
```
nuance1979, line 85: May I suggest moving `run_cli()` and friends into a new file called `cli.py`? So it becomes something like:

```python
from tianshou.utils import cli

cli.run_cli(main)
```

Or
```python
from tianshou.utils.cli import run_cli

run_cli(main)
```

tianshou/highlevel/logger.py
```diff
@@ -0,0 +1,73 @@
...
+        :param run_id: a unique name, which, depending on the logging framework, may be used to identify the logger
+        :param config_dict: a dictionary with data that is to be logged
+        :return: the logger
+        """
+
+
+class LoggerFactoryDefault(LoggerFactory):
+    def __init__(
+        self,
+        logger_type: Literal["tensorboard", "wandb"] = "tensorboard",
```
nuance1979, line 33: Maybe an enum called `class LoggerType(StrEnum)` here?

nuance1979:
Really nice work! Thanks a lot @opcode81 !

examples/mujoco/mujoco_a2c_hl.py
opcode81: `run_cli` is really the only function that is strictly concerned with CLIs. It's in the `util.logging` module because it also configures logging, adding log messages at the beginning and end of the execution as well as ensuring that any exceptions that may occur are appropriately logged. It is but an extension of the function `util.logging.run_main`, adding the `jsonargparse.CLI` wrapper around the function that is passed and configuring `jsonargarse` for use with dataclasses.

docs/autogen_rst.py
```diff
@@ -0,0 +1,121 @@
+import logging
```
MischaPanch, line 1: This is the usual script we used in our projects, right? Do you think it makes sense to add at this stage already? You haven't modified the CI to do this and I generally haven't looked into the way tianshou docs are being built at the moment

examples/atari/atari_dqn_hl.py
```diff
@@ -0,0 +1,105 @@
...
+        .with_model_factory(IntermediateModuleFactoryAtariDQN())
+        .with_trainer_epoch_callback_train(
+            TrainEpochCallbackNatureDQNEpsLinearDecay(eps_train, eps_train_final),
+        )
+        .with_trainer_epoch_callback_test(TestEpochCallbackDQNSetEps(eps_test))
+        .with_trainer_stop_callback(AtariStopCallback(task))
+    )
+    if icm_lr_scale > 0:
+        builder.with_policy_wrapper_factory(
+            PolicyWrapperFactoryIntrinsicCuriosity(
```
MischaPanch, line 90: I think it would be better to use `kwargs` explicitly in most methods that have more than say 3-4 parameters. Makes it easier to understand the logic and prevents difficult-to-debug mistakes when the order is accidentally switched by the user. You could use  `(*, <args>)` in the constructor of `PolicyWrapperFactoryIntrinsicCuriosity` and other places where it's appropriate to ensure that this is happening - similar to how I've done it for the policies in the last PR

examples/atari/atari_network.py
```diff
@@ -220,3 +229,43 @@ def forward(
...
+class ActorFactoryAtariDQN(ActorFactory):
+    def __init__(self, hidden_size: int | Sequence[int], scale_obs: bool, features_only: bool):
+        self.hidden_size = hidden_size
+        self.scale_obs = scale_obs
+        self.features_only = features_only
+
+    def create_module(self, envs: Environments, device: TDevice) -> Actor:
+        net_cls = scale_obs(DQN) if self.scale_obs else DQN
+        net = net_cls(
+            *envs.get_observation_shape(),
```
MischaPanch, line 243: also here, I'd prefer using named args

examples/atari/atari_ppo_hl.py
```diff
@@ -0,0 +1,117 @@
...
+            ),
+        )
+        .with_actor_factory(ActorFactoryAtariDQN(hidden_sizes, scale_obs, features_only=True))
+        .with_critic_factory_use_actor()
+        .with_trainer_stop_callback(AtariStopCallback(task))
+    )
+    if icm_lr_scale > 0:
+        builder.with_policy_wrapper_factory(
+            PolicyWrapperFactoryIntrinsicCuriosity(
+                IntermediateModuleFactoryAtariDQNFeatures(),
```
MischaPanch: named args

examples/atari/atari_network.py
```diff
@@ -5,7 +5,16 @@
 import torch
 from torch import nn
 
-from tianshou.utils.net.discrete import NoisyLinear
+from tianshou.highlevel.env import Environments
```
MischaPanch, line 8: These modules are growing and are not included in the package code. Users who wan't to train on atari will need to copy-paste them to their repos. I wonder if now is the right time to move it to the package, maybe a module like `common_envs.atari`?

examples/atari/atari_sac_hl.py
```diff
@@ -0,0 +1,103 @@
+#!/usr/bin/env python3
+
+import os
+
+from examples.atari.atari_network import (
+    ActorFactoryAtariDQN,
+    IntermediateModuleFactoryAtariDQNFeatures,
+)
+from examples.atari.atari_wrapper import AtariEnvFactory, AtariStopCallback
+from tianshou.highlevel.config import SamplingConfig
```
MischaPanch, line 10: Shorten package name to `hl`? Makes all imports shorter :). Just a proposal, I don't care too much

examples/atari/atari_sac_hl.py
```diff
@@ -0,0 +1,103 @@
...
+                IntermediateModuleFactoryAtariDQNFeatures(),
+                [hidden_size],
+                actor_lr,
+                icm_lr_scale,
+                icm_reward_scale,
+                icm_forward_loss_weight,
+            ),
+        )
+    experiment = builder.build()
+    experiment.run(log_name)
```
MischaPanch, line 99: since run returns a result object, I think it would be good to include examples for what can be done with it in the scripts - like printing some basic statistics about the training. It might make sense to add a new method to `ExperimentResult` (and/or to the objects it contains) like `get/print_summary`

examples/atari/atari_wrapper.py
```diff
@@ -369,3 +371,35 @@ def make_atari_env(task, seed, training_num, test_num, **kwargs):
...
+
+class AtariStopCallback(TrainerStopCallback):
+    def __init__(self, task: str):
+        self.task = task
+
+    def should_stop(self, mean_rewards: float, context: TrainingContext) -> bool:
+        env = context.envs.env
+        if env.spec.reward_threshold:
+            return mean_rewards >= env.spec.reward_threshold
+        if "Pong" in self.task:
```
MischaPanch, line 403: Could you add a comment on why this is necessary?

examples/mujoco/mujoco_a2c_hl.py
```diff
@@ -0,0 +1,85 @@
...
+    task: str = "Ant-v3",
+    buffer_size: int = 4096,
+    hidden_sizes: Sequence[int] = (64, 64),
+    lr: float = 7e-4,
+    gamma: float = 0.99,
+    epoch: int = 100,
+    step_per_epoch: int = 30000,
+    step_per_collect: int = 80,
+    repeat_per_collect: int = 1,
+    batch_size: int = 99999,
```
MischaPanch, line 33: batch_size larger than buffer size (which is minuscule)? Might be like this in current example, but seems fishy to me. @Trinkle23897 

I have the suspicion that no mini-batching is happening here

examples/mujoco/mujoco_a2c_hl.py
```diff
@@ -0,0 +1,85 @@
...
+        buffer_size=buffer_size,
+        step_per_collect=step_per_collect,
+        repeat_per_collect=repeat_per_collect,
+    )
+
+    env_factory = MujocoEnvFactory(task, experiment_config.seed, obs_norm=True)
+
+    experiment = (
+        A2CExperimentBuilder(env_factory, experiment_config, sampling_config)
+        .with_a2c_params(
```
MischaPanch, line 61: General question - should this maybe just be `with_params` or `with_policy_params`? It would allow to use a generic interface for builders. It's clear (and would be annotated by generics) that the `A2CExperimentBuilder` can only take params of type `A2CParams`

examples/mujoco/mujoco_env.py
```diff
@@ -1,34 +1,40 @@
...
 try:
     import envpool
 except ImportError:
     envpool = None
 
+log = logging.getLogger(__name__)
 
-def make_mujoco_env(task, seed, training_num, test_num, obs_norm):
+
+def make_mujoco_env(task: str, seed: int, num_train_envs: int, num_test_envs: int, obs_norm: bool):
```
MischaPanch, line 20: General comment - feel free to add TODOs where you see fit. I think `num_test_envs` should be 0 by default, or maybe 1. Currently passing 0 will lead to an error, I think. The test envs should also be created with a different seed. in fact with multiple different seeds... See the references in #978 

I understand it's outside of the scope of your PR to fix it, but as many problems as possible should be made visible

examples/mujoco/mujoco_env.py
```diff
@@ -37,3 +43,46 @@ def make_mujoco_env(task, seed, training_num, test_num, obs_norm):
         test_envs = VectorEnvNormObs(test_envs, update_obs_rms=False)
         test_envs.set_obs_rms(train_envs.get_obs_rms())
     return env, train_envs, test_envs
+
+
+class MujocoEnvObsRmsPersistence(Persistence):
```
MischaPanch, line 48: It's a bit unfortunate that this is necessary, and it has nothing to do with Mujoco. What do you think will be the future of persistence for such stateful wrappers? Is this just a temporary measure? If yes, this should be highlighted

examples/mujoco/mujoco_env.py
```diff
@@ -37,3 +43,46 @@ def make_mujoco_env(task, seed, training_num, test_num, obs_norm):
...
+
+    def create_envs(self, num_training_envs: int, num_test_envs: int) -> ContinuousEnvironments:
+        env, train_envs, test_envs = make_mujoco_env(
+            task=self.task,
+            seed=self.seed,
+            num_train_envs=num_training_envs,
+            num_test_envs=num_test_envs,
+            obs_norm=self.obs_norm,
+        )
+        envs = ContinuousEnvironments(env=env, train_envs=train_envs, test_envs=test_envs)
```
MischaPanch, line 85: I thought a bit more, and in the spirit of discoverability with the IDE, maybe it does make sense to rename to `EnvironmentsContinuous`? I have no strong opinion on this though

examples/mujoco/mujoco_ppo_hl.py
```diff
@@ -0,0 +1,97 @@
...
+import torch
+
+from examples.mujoco.mujoco_env import MujocoEnvFactory
+from tianshou.highlevel.config import SamplingConfig
+from tianshou.highlevel.experiment import (
+    ExperimentConfig,
+    PPOExperimentBuilder,
+)
+from tianshou.highlevel.params.dist_fn import (
+    DistributionFunctionFactoryIndependentGaussians,
```
MischaPanch, line 16: We were talking about introducing shorter aliases to increase readability - here `IndependentGaussians` comes to mind.

examples/mujoco/mujoco_sac_hl.py
```diff
@@ -0,0 +1,83 @@
...
+    step_per_epoch: int = 5000,
+    step_per_collect: int = 1,
+    update_per_step: int = 1,
+    n_step: int = 1,
+    batch_size: int = 256,
+    training_num: int = 1,
+    test_num: int = 10,
+):
+    now = datetime.datetime.now().strftime("%y%m%d-%H%M%S")
+    log_name = os.path.join(task, "sac", str(experiment_config.seed), now)
```
MischaPanch: use `datetime_tag` as in other scripts. This may happen in other places as well - maybe a search for "now" would find them

tianshou/policy/modelfree/pg.py
```diff
@@ -17,7 +17,8 @@
 from tianshou.policy.base import TLearningRateScheduler
 from tianshou.utils import RunningMeanStd
 
-TDistParams = torch.Tensor | tuple[torch.Tensor]
+# TODO: Is there a better way to define this type? mypy doesn't like Callable[[torch.Tensor, ...], torch.distributions.Distribution]
```
MischaPanch, line 20: The better way is probably a callable prototype

tianshou/trainer/base.py
```diff
@@ -79,7 +82,9 @@ class BaseTrainer(ABC):
         e.g., the reward of agent 1 or the average reward over all agents.
     :param logger: A logger that logs statistics during
         training/testing/updating. Default to a logger that doesn't log anything.
-    :param verbose: whether to print the information. Default to True.
+    :param verbose: whether to print status information to stdout.
+        If set to False, status information will still be logged (provided that
+        logging is enabled via the `logging` module).
     :param show_progress: whether to display a progress bar when training.
         Default to True.
```
MischaPanch: Pls remove all these `Default to ...` statements when you encounter them. Then, step by step, we'll get rid of them all at some point in the future. I'm doing the same in my PRs

tianshou/utils/logging.py
```diff
@@ -0,0 +1,155 @@
...
+from collections.abc import Callable
+from datetime import datetime
+from io import StringIO
+from logging import *
+from typing import Any
+
+log = getLogger(__name__)
+
+LOG_DEFAULT_FORMAT = "%(levelname)-5s %(asctime)-15s %(name)s:%(funcName)s - %(message)s"
+_logFormat = LOG_DEFAULT_FORMAT
```
MischaPanch, line 20: Could you pls add a block comment explaining that this var will be accessed by several methods below, and may be overwritten by `configure`, in order to configure logging globally?

It's not entirely obvious and some readers may be unfamiliar with this pattern

tianshou/utils/logging.py
```diff
@@ -0,0 +1,155 @@
...
+    Using this function requires that `jsonargparse` and `docstring_parser` be available.
+    Like `run_main`, two additional log messages will be logged (at the beginning and end
+    of the execution), and it is ensured that all exceptions will be logged.
+
+    :param main_fn: the function to be executed
+    :param format: the log message format
+    :param level: the minimum log level
+    :return: the result of `main_fn`
+    """
+    from jsonargparse import set_docstring_parse_options, CLI
```
MischaPanch, line 94: Write a good message on import error?

tianshou/utils/logging.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+def datetime_tag() -> str:
+    """:return: a string tag for use in log file names which contains the current date and time (compact but readable)"""
+    return datetime.now().strftime("%Y%m%d-%H%M%S")
+
+
+_fileLoggerPaths: list[str] = []
+_isAtExitReportFileLoggerRegistered = False
+_memoryLogStream: StringIO | None = None
```
MischaPanch, line 107: use pep8?

tianshou/utils/net/common.py
```diff
@@ -595,3 +598,13 @@ def forward(self, obs: np.ndarray | torch.Tensor, *args, **kwargs) -> Any:
         return new_net_class
 
     return decorator_fn, new_state_shape
+
+
+class BaseActor(nn.Module, ABC):
```
MischaPanch, line 603: Maybe instead `ActorProtocol` with fields `preprocess` and `output_dim`? I don't feel like the new getter methods that return public attributes are adding value

MischaPanch (commented):
I reviewed the examples and changes in existing modules, but barely had a look at the new code in `tianshou.highlevel`. However, I believe it's not that important to review this new package right now, and would instead start using and incrementally improving the implementations of highlevel functionality after merge. So I don't plan on diving deeper in that part, if you agree

tianshou/highlevel/logger.py
MischaPanch: I lately switched to using `Literal` for such cases, as it just introduces less code. Especially for the logger, where a corresponding `Enum` would likely not be used in many places for annotation.

Even if one needs something for annotation in many places in the future, `LoggerName = Literal[...]` could be factored out then and would do the job

examples/atari/atari_dqn_hl.py
MischaPanch: Also, how about using an alias type `IntrinsicCuriousity` for better readability?

examples/atari/atari_ppo_hl.py
MischaPanch: And maybe type alias `AtariDQNFeatures`?

docs/autogen_rst.py
opcode81: I did modify the CI. The auto-generation of rst files for the `highlevel` subpackage was added to pyproject.toml and therefore will take place when running `poe doc-build`:

image

Yes, it's based on our script, but I adjusted it slightly to produce output that aligns with the existing docs.
You can run the poe task and take a look.

examples/atari/atari_dqn_hl.py
opcode81: Agreed, forced kwargs can improve readability (and potentially avoid errors) here. Added in commit da2194eff6c12d32d8ca0e717dedb480243a31ef.

As far as the shorter aliases are concerned, I am still not entirely convinced they are needed. I would never use them personally, because they're just not easily discovered via auto-completion. And you lose the semantics. What do the others think?

examples/atari/atari_network.py
opcode81: I agree, but this is not part of the high-level API. Should I change it anyway?

examples/atari/atari_ppo_hl.py
opcode81: Same as above.

examples/atari/atari_network.py
opcode81: I think it would make sense. What do you think, @Trinkle23897?
It would also remove the need for one of the two symlinks in the repo (#944).

tianshou/utils/logging.py
opcode81: Done

examples/atari/atari_wrapper.py
opcode81: This is just reproducing the functionality of `stop_fn` in the low-level examples. Not my idea.

examples/mujoco/mujoco_a2c_hl.py
opcode81: I had actually written it using `Generic` in the first place, but unfortunately, the type variable is not resolved when displaying type hints to the user in JetBrains IDEs:

image

It works in VS Code though.

Usability being of the utmost importance, I decided against using `Generic`.

Also, I can't really think of a polymorphic use case for experiment builders, as you'll usually want to call their specific methods, so using the same name for all such methods won't really improve anything.

examples/mujoco/mujoco_env.py
opcode81: As far as the high-level API is concerned, the relevant place to change is `SamplingConfig`. I have now changed the defaults there to 1 for testing, and os.cpu_count() for training. Commit 8dedea0f92b985fea067656db16b89160561397e

examples/mujoco/mujoco_env.py
opcode81: Yes, this could also be applied outside of the MuJoCo example. But it does have something to do with the MuJoCo environment implementation, because the envs can be explicitly configured to be wrapped in a `VectorEnvNormObs`, which makes use of the statistic that is persisted here. We could consider adding (an extensible list of) persistence handlers to the `BaseVectorEnv` representation and thus handle this directly in `VectorEnvNormObs`. The higher-level persistence can then retrieve these handlers and apply them.

examples/mujoco/mujoco_env.py
opcode81: I am a bit torn. There will probably never be many more implementations than the two we have now, and `ContinuousEnvironments` just rolls off the tongue better. But I don't mind either way. Final call?

examples/mujoco/mujoco_ppo_hl.py
opcode81: `IndependentGaussians` already exists, but it wasn't used here, because using it requires me to change my preferred workflow (see above).

examples/mujoco/mujoco_sac_hl.py
opcode81: Done

tianshou/policy/modelfree/pg.py
opcode81: I don't think it's possible with a prototype either. This just isn't a well-defined callable type, because some functions accept only one tensor, others more than one. The `Callable` definition I had initially seems to work (at least in IntelliJ), but MyPy didn't like it, as stated in the comment:

image

tianshou/utils/logging.py
opcode81: Done. cf2a3d56f9dcf69358061b45ad2b52ad2f0a606e

tianshou/utils/logging.py
opcode81: The regular import error is already good enough imo.
The error that is issued by `set_docstring_parse_options` by default is also excellent.

tianshou/utils/logging.py
opcode81: Actually, this is already PEP8-compliant. PEP8 does not require module or class attributes to use snake case.
(Also, changing existing modules that I copied from sensai in order for them to satisfy new requirements really wasn't a priority for me.)

tianshou/utils/net/common.py
opcode81: A Protocol is not the right concept here. There is no use case where all that matters is for the two methods to be present.  (And even if there was, an explicit type relationship would be better than an implicit one.)

The concept that is required here is that of a neural network module that is able to provide the functionality given by the two methods - and the abstraction represents exactly this.
The fact that the implementations of the methods mostly return public members is an irrelevant implementation detail.

General discussion: Protocols serve to provide type annotations for duck typing (which arguably has no place in well-typed libraries). The only case where I would prefer a protocol over an explicit subclass relationship is a case where the classes are truly unrelated and where I have no control over their implementation. In all other cases, the explicit relationship is preferable because 
 * you get a type bound that facilitates discovery of implementing types (and which you can always check against)
 * explicitness avoids errors, as the relationship can be made use of in order to
     * check that all implementations indeed satisfy the interface (even if it changes)
     * apply automated refactoring (eliminating any ambiguity as to which implementations need updating).

examples/atari/atari_sac_hl.py
opcode81: While it is shorter, it's also less clear. Especially new users won't immediately know what it stands for.

examples/atari/atari_sac_hl.py
opcode81: `run` already logs everything, so there is, at present, no need to use the result in the examples.
But once we add more documentation on the high-level API, we should definitely demonstrate it.

tianshou/trainer/base.py
opcode81: Done

docs/autogen_rst.py
MischaPanch: Ah, I see

examples/atari/atari_network.py
MischaPanch: Sure, it's fully uncontroversial

tianshou/utils/net/common.py
MischaPanch: You could inherit from the protocol, to make it explicit. My point was not to introduce duck typing but rather to not bloat the public interface by adding rather redundant methods. It can also be achieved in a different way:

1. Make the public field private. You could then also use abstract properties in the base class instead of getters
2. Let the abstract class have an init that creates these attributes

I would lean towards the second solution as it needs less refactoring, and I remember already doing something like this with the critics

opcode81: Inheriting from a protocol does not really work (and partly defeats the purpose of a Protocol tbh), as it does not enforce that the protocol is actually implemented - and no one (i.e. no dev tool I am aware of) will complain if it's not or offer you an auto-completion that will help you get it done.

I don't see how changing the base class to have state, necessitating a super call in init, improves anything. Without it, subclasses are free to choose any internal representation and simply implement the two methods. Importantly, using public properties rather than methods is not generally viewed as a preferable interface; this may be your personal preference, but it does not align with general conventions in OOP. From the point of view of the abstraction `BaseActor`, there is a clearly defined interface based on methods. The fact that the interface could also take another form when looking at the subclasses (all of which happen to use public attributes) is not relevant, nor is it a problem.

In general, the reason why methods are preferred is that internal representations often change, and many programmers carelessly use public attributes and later break the (supposedly public) interface by changing the attributes. A method-based interface can be more stable (and is perceived as reliable/permanent); For this reason, the convention to never use any public properties except in dataclass-style representations is frequently adhered to. (Python's property annotation can help provide stable interfaces when using properties, of course, but the fact remains that a public property just doesn't really mean much to most Python programmers.)

MischaPanch: Ok, it's fine for me leaving it as is. But does anything speak against making the corresponding attributes private in the implementations? Could be done later as well, course

MischaPanch (dismissed):
@Trinkle23897 For me this looks good, there are some unresolved discussions about minor things above. Pls have a look when you find time

I changed the merge strategy to no longer squash but instead to add all commits and a merge commit

Trinkle23897:
Sorry I just came back from vacation, will take a look in this weekend

tianshou/utils/net/common.py
opcode81: I wouldn't mind making them private, but it constitutes a breaking change and would perhaps seem odd, as virtually all attributes are public. The library, thus far, does not really consistently differentiate between private and public attributes. Making an isolated change won't do much to change that.
Does it bother you if they remain public? I think the key point here is to primarily look at the abstraction and whether it makes sense; The two ways of retrieving the information in the concrete classes are but a side-effect of implementing the abstraction (and the "public attributes" convention).

nuance1979:
> I changed the merge strategy to no longer squash but instead to add all commits and a merge commit

I don't think that's a good idea. A PR usually goes through many rounds of revisions and there is no point keeping those history in the main repo.

tianshou/utils/string.py
```diff
@@ -0,0 +1,536 @@
+"""Copy of sensai.util.string from sensAI commit d7b4afcc89b4d2e922a816cb07dffde27f297354."""
```
Trinkle23897, line 1: I think it's better to directly use sensAI utils instead of copy-paste here, otherwise you need to add extra tests and need to pay extra attention to sync the code. Same as logging.py

Trinkle23897: Also the typing is missing for some function

pyproject.toml
```diff
@@ -102,7 +104,7 @@ warn_redundant_casts = true
 warn_unreachable = true
 warn_unused_configs = true
 warn_unused_ignores = true
-exclude = "^build/|^docs/"
```
Trinkle23897, line 105: could we revert this change?

examples/mujoco/mujoco_a2c_hl.py
Trinkle23897: yes that's true, and it's a tuned hparam

examples/atari/atari_network.py
Trinkle23897: yeah sgtm, maybe in env/atari.py or utils/whatever.py?

Trinkle23897 (commented):
I only checked some conversations and did a quick skim, will do a full review later

I still recommend using squash instead of merge, to avoid screwing up the master branch and keep a linear history

tianshou/utils/string.py
opcode81: With commit fdb0eba93d81fa5e698770b4f7088c87fc1238da we now depend on sensAI instead of copying modules from it. Note that this does add some transitive dependencies though.

pyproject.toml
opcode81: It's reverted (save for the addition of `^temp/`) as part of the sensAI dependency commit.

examples/atari/atari_network.py
opcode81: Don't really have spare time to handle this right now, so given that it's not strictly related to this PR, resolving without action.

opcode81:
> > I changed the merge strategy to no longer squash but instead to add all commits and a merge commit
> 
> I don't think that's a good idea. A PR usually goes through many rounds of revisions and there is no point keeping those history in the main repo.

@nuance1979, actually, there are some very good reasons to keep the original history. I mention some of them [here](https://github.com/thu-ml/tianshou/pull/950#issuecomment-1752112379).

There is no inherent value in a linear history, yet there is value
 * in the information contained in the individual commit messages and 
 * in the flexibility gained by 
    * being able to revert individual commits
    * being able to build on a branch's history (which can't be done if it is rewritten)
    * being able to bissect the history at a lower level of granularity

Squashing is more reasonable for smaller PRs, but this one adds thousands of lines of code.

opcode81:
FYI I will be on holiday for the next 10 days. I will address your further comments when I return.

tianshou/utils/net/common.py
MischaPanch: Resolving, since it's out of scope for this PR and you don't have much time. At some point in the future we can think about improving the structure of the public interfaces

examples/mujoco/mujoco_a2c_hl.py
MischaPanch: Should that be addressed somehow? Maybe documented or set to be equal to buffer size, or to -1/None for more clarity? Feel free to open an issue. It's out of scope for this PR?

nuance1979:
> > > I changed the merge strategy to no longer squash but instead to add all commits and a merge commit
> > 
> > 
> > I don't think that's a good idea. A PR usually goes through many rounds of revisions and there is no point keeping those history in the main repo.
> 
> @nuance1979, actually, there are some very good reasons to keep the original history. I mention some of them [here](https://github.com/thu-ml/tianshou/pull/950#issuecomment-1752112379).
> 
> There is no inherent value in a linear history, yet there is value
> 
> * in the information contained in the individual commit messages and
> * in the flexibility gained by
>   
>   * being able to revert individual commits
>   * being able to build on a branch's history (which can't be done if it is rewritten)
>   * being able to bissect the history at a lower level of granularity
> 
> Squashing is more reasonable for smaller PRs, but this one adds thousands of lines of code.

I understand your points but would still argue for squashing because:

- The information contained in the individual commit messages are still preserved in the PR even after squashing.
- Individual commits inside a PR is rarely used to revert. In fact, I would argue that PR is the proper unit for reversion, i.e., PR should be a self-contained change that can be reverted without a broken tip. we don't usually care if a commit inside a PR passes full CI/CD tests so reverting to it would mean a broken main branch; for a PR, we do expect it to pass.
- The granularity is controlled by having smaller PRs. Basically, as long as it is self contained, a smaller PR is preferred to a larger one because the time it takes to review a large PR is an impediment to development velocity.

Anyway, just my personal opinions.

examples/atari/atari_network.py
Trinkle23897: could you create an issue?

examples/mujoco/mujoco_a2c_hl.py
Trinkle23897: Yeah, -1/None is a good idea since 99999 might be a footgun. Feel free to create an issue.

examples/mujoco/mujoco_env.py
Trinkle23897: could it be a parameter for `ContinuousEnvironments`?

examples/mujoco/mujoco_env.py
Trinkle23897: `EnvironmentsContinuous` is a little bit weird. In RL terminology, we always call continuous environments.

examples/mujoco/mujoco_npg_hl.py
```diff
@@ -0,0 +1,87 @@
...
+        step_per_epoch=step_per_epoch,
+        batch_size=batch_size,
+        num_train_envs=training_num,
+        num_test_envs=test_num,
+        buffer_size=buffer_size,
+        step_per_collect=step_per_collect,
+        repeat_per_collect=repeat_per_collect,
+    )
+
+    env_factory = MujocoEnvFactory(task, experiment_config.seed, obs_norm=True)
```
Trinkle23897, line 59: could you make it force kwargs?

examples/mujoco/mujoco_npg_hl.py
```diff
@@ -0,0 +1,87 @@
...
+
+    experiment = (
+        NPGExperimentBuilder(env_factory, experiment_config, sampling_config)
```
Trinkle23897, line 62: same as ExperimentBuilder

examples/mujoco/mujoco_npg_hl.py
```diff
@@ -0,0 +1,87 @@
...
+                actor_step_size=actor_step_size,
+                lr=lr,
+                lr_scheduler_factory=LRSchedulerFactoryLinear(sampling_config)
+                if lr_decay
+                else None,
+                dist_fn=DistributionFunctionFactoryIndependentGaussians(),
+            ),
+        )
+        .with_actor_factory_default(hidden_sizes, torch.nn.Tanh, continuous_unbounded=True)
+        .with_critic_factory_default(hidden_sizes, torch.nn.Tanh)
```
Trinkle23897, line 80: and kwargs for this

tianshou/highlevel/agent.py
```diff
@@ -0,0 +1,598 @@
...
+)
+TDiscreteCriticOnlyParams = TypeVar(
+    "TDiscreteCriticOnlyParams",
+    bound=Params | ParamsMixinLearningRateWithScheduler,
+)
+TPolicy = TypeVar("TPolicy", bound=BasePolicy)
+log = logging.getLogger(__name__)
+
+
+class AgentFactory(ABC, ToStringMixin):
```
Trinkle23897, line 83: force all function to use kwargs?

tianshou/highlevel/agent.py
```diff
@@ -0,0 +1,598 @@
...
+            actor_optim=actor.optim,
+            critic=critic.module,
+            critic_optim=critic.optim,
+            action_space=envs.get_action_space(),
+            observation_space=envs.get_observation_space(),
+            **kwargs,
+        )
+
+
+class REDQAgentFactory(OffPolicyAgentFactory):
```
Trinkle23897, line 452: Is there any documentation describing how to add an extra RL algorithm? I guess right now we need to add it to at least two places -- one is tianshou/policy/* and another one is tianshou/highlevel/agent.py

tianshou/highlevel/module/actor.py
```diff
@@ -0,0 +1,261 @@
...
+
+
+class ContinuousActorType(Enum):
+    GAUSSIAN = "gaussian"
+    DETERMINISTIC = "deterministic"
+    UNSUPPORTED = "unsupported"
+
+
+@dataclass
+class ActorFuture:
```
Trinkle23897, line 34: `Future` confused me -- I originally think it's a threadpool Future. Maybe another better naming?

tianshou/highlevel/module/actor.py
```diff
@@ -0,0 +1,261 @@
...
+        return discrete.Actor(
+            net_a,
+            envs.get_action_shape(),
+            hidden_sizes=(),
+            device=device,
+            softmax_output=self.softmax_output,
+        ).to(device)
+
+
+class ActorFactoryTransientStorageDecorator(ActorFactory):
```
Trinkle23897, line 231: docs? I don't quite get why do we need `ActorFactoryTransientStorageDecorator` and `IntermediateModuleFactoryFromActorFactory`. Could we move them into base class? Is it an over-engineering?

tianshou/highlevel/module/intermediate.py
```diff
@@ -0,0 +1,27 @@
...
+
+import torch
+from sensai.util.string import ToStringMixin
+
+from tianshou.highlevel.env import Environments
+from tianshou.highlevel.module.core import ModuleFactory, TDevice
+
+
+@dataclass
+class IntermediateModule:
```
Trinkle23897, line 12: It would be nice if you can have a figure that illustrates each Module's scope in docs/high_level_api. It's because at the first glance in actor.py I didn't understand what is Intermediate Module.

tianshou/highlevel/module/module_opt.py
```diff
@@ -0,0 +1,29 @@
...
+@dataclass
+class ModuleOpt:
+    """Container for a torch module along with its optimizer."""
+
+    module: torch.nn.Module
+    optim: torch.optim.Optimizer
+
+
+@dataclass
+class ActorCriticModuleOpt:
```
Trinkle23897: can `ModuleOpt` be a base class of `ActorCriticModuleOpt`? Otherwise I guess we need to deal with Union typing and different attr (.module and .actor_critic_module) all the time. Mypy has already caught the typing issue.

tianshou/utils/net/common.py
Trinkle23897: Yeah, output_dim is really hacky in the current implementation. I totally agree we should make it as a standard interface.

tianshou/trainer/base.py
```diff
@@ -19,6 +20,8 @@
     tqdm_config,
 )
 
+log = logging.getLogger(__name__)
```
Trinkle23897, line 23: can we rename it to logger?
```suggestion
logger = logging.getLogger(__name__)
```

tianshou/highlevel/params/noise.py
```diff
@@ -0,0 +1,34 @@
...
+
+    def __init__(self, std_fraction: float):
+        """:param std_fraction: fraction (between 0 and 1) of the maximum action value that shall
+        be used as the standard deviation
+        """
+        self.std_fraction = std_fraction
+
+    def create_noise(self, envs: Environments) -> GaussianNoise:
+        envs.get_type().assert_continuous(self)
+        envs: ContinuousEnvironments
```
Trinkle23897, line 28: use cast or assert?

tianshou/highlevel/persistence.py
```diff
@@ -0,0 +1,125 @@
...
+from typing import TYPE_CHECKING
+
+import torch
+
+from tianshou.highlevel.world import World
+
+if TYPE_CHECKING:
+    from tianshou.highlevel.module.core import TDevice
+
+log = logging.getLogger(__name__)
```
Trinkle23897, line 14:
```suggestion
logger = logging.getLogger(__name__)
```

tianshou/highlevel/experiment.py
```diff
@@ -0,0 +1,1099 @@
...
+        hidden_sizes: Sequence[int] = CriticFactoryDefault.DEFAULT_HIDDEN_SIZES,
+        hidden_activation: ModuleType = torch.nn.ReLU,
+    ) -> Self:
+        """Makes the second critic use the default, MLP-style architecture with the given parameters.
+
+        :param hidden_sizes: the sequence of dimensions to use in hidden layers of the network
+        :param hidden_activation: the activation function to use for hidden layers
+        :return: the builder
+        """
+        self._with_critic_factory_default(0, hidden_sizes, hidden_activation)
```
Trinkle23897:
```suggestion
        self._with_critic_factory_default(1, hidden_sizes, hidden_activation)
```

Trinkle23897: Please double check other cases, I just came across to this bug

tianshou/utils/string.py
opcode81: @Trinkle23897, so do you really prefer the added dependencies?

Trinkle23897: I think it’s fine, open to both options

opcode81: Just to clarify, this is what is added:

``````Package operations: 16 installs, 0 updates, 0 removals

  • Installing contourpy (1.2.0)
  • Installing cycler (0.12.1)
  • Installing fonttools (4.44.0)
  • Installing kiwisolver (1.4.5)
  • Installing pillow (10.0.1)
  • Installing pyparsing (3.1.1)
  • Installing python-dateutil (2.8.2)
  • Installing pytz (2023.3.post1)
  • Installing tzdata (2023.3)
  • Installing joblib (1.3.2)
  • Installing matplotlib (3.8.1)
  • Installing pandas (2.1.2)
  • Installing threadpoolctl (3.2.0)
  • Installing scikit-learn (1.3.2)
  • Installing seaborn (0.13.0)
  • Installing sensai (1.0.0b3)
```

tianshou/highlevel/params/noise.py
opcode81: Cast would make no difference and it's already asserted in the line above (albeit in a way that is not apparent to the type checker).

tianshou/highlevel/persistence.py
opcode81: `log` is a very commonly used name. In Java's logging frameworks, which Python's `logging` was inspired by (hence the camelCase), `log` is the recommended name; In Python code, both `log` and `logger` are common.
If you really prefer `logger`, I'll change it, but the use of a different identifier can also help to differentiate it from other loggers (like TensorBoard or wandb).
(Note that I can't really commit the suggestion, because the references need to be updated as well.)

tianshou/highlevel/agent.py
opcode81: I don't think it provides any benefit. Forcing kwargs makes sense only if there is a danger of getting parameters mixed up otherwise, which is the case if there is a large number of parameters or if there are several parameters of the same type (that don't have a natural ordering).
Python is not Objective-C and we shouldn't enforce such paradigms if there isn't an obvious benefit. It would likely irritate users.

examples/mujoco/mujoco_npg_hl.py
opcode81: What are you referring to?

tianshou/highlevel/module/module_opt.py
opcode81: Which typing issue are you referring to? Mypy reports no issues on my end.

An `ActorCriticModuleOpt` instance can never be used in place of a `ModuleOpt`. No Union typing is being used, the two classes are used in distinct cases. I realize that the name of the former class perhaps suggests a type relationship, but there isn't one. I think the best solution is to rename `ActorCriticModuleOpt` to `ActorCriticOpt` in order to avoid confusion (which I have now done).

tianshou/trainer/base.py
opcode81: See comment on respective suggestion below.

tianshou/highlevel/experiment.py
opcode81: Good catch. Found no further index-related issues.

examples/mujoco/mujoco_npg_hl.py
opcode81: See general kwargs discussion below.

examples/mujoco/mujoco_npg_hl.py
opcode81: See general kwargs discussion below.

examples/mujoco/mujoco_env.py
opcode81: It could be. But I don't think we need to address this in this PR.

tianshou/highlevel/module/intermediate.py
opcode81: Not sure what you mean. Please elaborate - perhaps in an issue description. I would suggest to postpone such documentation improvements.

tianshou/highlevel/module/actor.py
opcode81: I think it's a fitting name, because it's a realization of the same fundamental concept of a "future", i.e. an object that can provide a value in the future. In this case, it's just a simple container, because there is no asynchronicity involved.
Note that this is an internal class which library users will not be interacting with.

Do you have another suggestion?

tianshou/highlevel/module/actor.py
opcode81: The decorator serves to store the most recently created actor instance, such that it can be retrieved. This is necessary for the implementation of the concept of reusing the agent's preprocessing model for the critics, as we need to be able to access the object that was created by the `ActorFactory` in order to reuse it for the creation of the critics.

`IntermediateModuleFactoryFromActorFactory` is an unrelated class, which allows an `ActorFactory` to be used in contexts where an `IntermediateModuleFactory` is expected. I use it in DQN and IQN, for practical reasons only.

tianshou/highlevel/module/actor.py
Trinkle23897: No I don't

tianshou/highlevel/persistence.py
Trinkle23897: okay let's keep it as-is

tianshou/highlevel/module/intermediate.py
Trinkle23897: create an issue #985

tianshou/utils/string.py
Trinkle23897: For now I admit I hesitate to add this dependency. The drawback I can foresee is that in the future if we want to customize the utility function (e.g., redirect log to some other places) it would be much harder because we need to change sensai as well, and that might slow down the development speed. If you don't mind feel free to revert to the old version, I'm very sorry.

Trinkle23897 (commented):
I'll give @MischaPanch the final call

tianshou/utils/string.py
opcode81: I have reverted the commit that made the change. Should we change our minds later, we can always go back (and revert the reversion) - provided that we don't squash the PR.

opcode81: Note that lack of customizability would not have been an issue though, because the logging module is fully configurable and essentially makes no project-specific assumptions.

tianshou/highlevel/agent.py
opcode81: I have added documentation on this [here](https://github.com/aai-institute/tianshou/wiki/High-Level-API#supporting-a-new-algorithm)

MischaPanch: Will be addressed in #985

examples/mujoco/mujoco_a2c_hl.py
MischaPanch: Will be resolved in #986

MischaPanch:
I would merge this now, thanks to everybody for the vivid discussion and contributions! I feel like we're on a great path forward here :) 

I'm merging without squash because:

1. @opcode81 is strongly in favor of it
2. I am rather in favor
3. @Trinkle23897 and @nuance1979 are against it, but not very strongly, if I interpret it correctly.
4. The commit history here is kept clean through @opcode81 continuously rebasing and force-pushing
5. Importantly, we will have an option to revert to the sensai-dependent version without effort if we want that in the future

MischaPanch:
If no objections come within 30 mins, I'm merging.

@Trinkle23897 would you be ok to add @opcode81 as maintainer? We will continue working on tianshou together, and from this PR I think you can see that he's a very responsible contributor. Neither of us plans to take away control from you, but it would be more convenient if Dominik could help me with reviewing issues, PRs, making wiki entries, and so on.

MischaPanch (approved):

Trinkle23897:
yep, sg

Issue #948, opened 2023-09-24
Remove kwargs in policies' forward


MischaPanch:

MischaPanch:
Related to #949

PR #340, opened 2021-04-18
TRPO benchmark release

ChenDRAG:
See #338 for details

PR #991, opened 2023-11-11
Method to compute actions from observations

### LINKED ISSUE #981: Update PR Template
The template still mentions `make`.
Instead, contributors should be instructed to use `poe format`, `poe type-check` and `poe doc-build`.

MischaPanch:
This PR adds a new method for getting actions from an env's observation and info. This is useful for standard inference and stands in contrast to batch-based methods that are currently used in training and evaluation. Without this, users have to do some kind of gymnastics to actually perform inference with a trained policy. I have also added a test for the new method.

In future PRs, this method should be included in the examples (in the the "watch" section).

To add this required improving multiple typing things and, importantly, _simplifying the signature of `forward` in many policies!_ This is a **breaking change**, but it will likely affect no users. The `input` parameter of forward was a rather hacky mechanism, I believe it is good that it's gone now. It will also help with #948 .

The main functional change is the addition of `compute_action` to `BasePolicy`.

Other minor changes:
- improvements in typing
- updated PR and Issue templates
- Improved handling of `max_action_num`

Closes #981

### DIFF
--- a/tianshou/policy/base.py
+++ b/tianshou/policy/base.py
@@ -13,7 +13,11 @@ from torch import nn
 from tianshou.data import ReplayBuffer, to_numpy, to_torch_as
 from tianshou.data.batch import BatchProtocol
 from tianshou.data.buffer.base import TBuffer
-from tianshou.data.types import BatchWithReturnsProtocol, RolloutBatchProtocol
+from tianshou.data.types import (
+    BatchWithReturnsProtocol,
+    ObsBatchProtocol,
+    RolloutBatchProtocol,
+)
 from tianshou.utils import MultipleLRSchedulers
 
 logger = logging.getLogger(__name__)
@@ -152,7 +156,7 @@ class BasePolicy(ABC, nn.Module):
     @abstractmethod
     def forward(
         self,
-        batch: RolloutBatchProtocol,
+        batch: ObsBatchProtocol,
         state: dict | BatchProtocol | np.ndarray | None = None,
         **kwargs: Any,
     ) -> BatchProtocol:
--- a/tianshou/policy/base.py
+++ b/tianshou/policy/base.py
@@ -1,7 +1,7 @@
 import logging
 from abc import ABC, abstractmethod
 from collections.abc import Callable
-from typing import Any, Literal, TypeAlias, cast, overload
+from typing import Any, Literal, TypeAlias, cast
 
 import gymnasium as gym
 import numpy as np
@@ -11,9 +11,10 @@ from numba import njit
 from torch import nn
 
 from tianshou.data import ReplayBuffer, to_numpy, to_torch_as
-from tianshou.data.batch import BatchProtocol
+from tianshou.data.batch import Batch, BatchProtocol, arr_type
 from tianshou.data.buffer.base import TBuffer
 from tianshou.data.types import (
+    ActBatchProtocol,
     BatchWithReturnsProtocol,
     ObsBatchProtocol,
     RolloutBatchProtocol,
@@ -21,7 +22,6 @@ from tianshou.data.types import (
 from tianshou.utils import MultipleLRSchedulers
 
 logger = logging.getLogger(__name__)
-
 
 TLearningRateScheduler: TypeAlias = torch.optim.lr_scheduler.LRScheduler | MultipleLRSchedulers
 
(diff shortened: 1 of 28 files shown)


MischaPanch:
The cql integration test is failing, as it is on master currently. I don't know why, and it doesn't happen locally.. Any idea, @Trinkle23897 ?

```
FAILED test/offline/test_cql.py::test_cql - assert False
 +  where False = <function test_cql.<locals>.stop_fn at 0x7fc8a49911c0>(-1202.872077798278)
===== 1 failed, 109 passed, 1 skipped, 2800 warnings in 1210.42s (0:20:10) =====
```

nuance1979:
> The cql integration test is failing, as it is on master currently. I don't know why, and it doesn't happen locally.. Any idea, @Trinkle23897 ?
> 
> ```
> FAILED test/offline/test_cql.py::test_cql - assert False
>  +  where False = <function test_cql.<locals>.stop_fn at 0x7fc8a49911c0>(-1202.872077798278)
> ===== 1 failed, 109 passed, 1 skipped, 2800 warnings in 1210.42s (0:20:10) =====
> ```

We set an arbitrary threshold [here](https://github.com/thu-ml/tianshou/blob/962c6d1e11d033b6faaa2da37a89794563baebf6/test/offline/test_cql.py#L86) and check to see if after 5 iterations the model can beat it. The difference is most probably due to floating point math error but could be a hidden bug.

I'd suggest lowering the threshold to -1210 to let it pass for now and revisit it as part of the benchmarking work in the future.

Trinkle23897:
I'd suggest changing seed but I'm fine with lowering threshold

tianshou/data/buffer/base.py
```diff
@@ -290,20 +290,32 @@ def add(
             self._meta[ptr] = batch
         return ptr, ep_rew, ep_len, ep_idx
 
-    def sample_indices(self, batch_size: int) -> np.ndarray:
+    def sample_indices(self, batch_size: int | None) -> np.ndarray:
```
Trinkle23897: This change is actually wrong and unnecessary... I stepped back to take a look at #340, it uses https://github.com/thu-ml/tianshou/blob/a57503c0aa6d40a0c939b37f4ae3c3a83bb2e126/tianshou/trainer/onpolicy.py#L134-L136 which has different semantics -- sample size is 0 which will extract all data from buffer, but batch_size is 99999 which will presumably update all data in one forward-backward call. 

Now it's not consistent with above: https://github.com/thu-ml/tianshou/blob/962c6d1e11d033b6faaa2da37a89794563baebf6/tianshou/trainer/base.py#L496

You should revive that change in trainer and make some classes' policy.update function accept batch_size=None to allow such a feature, instead of changing here

tianshou/policy/base.py
```diff
@@ -149,13 +153,39 @@ def soft_update(self, tgt: nn.Module, src: nn.Module, tau: float) -> None:
...
+        # need to add empty batch dimension
+        obs = obs[None, :]
+        obs_batch = cast(ObsBatchProtocol, Batch(obs=obs, info=info))
+        act = self.forward(obs_batch, state=state).act.squeeze()
+        if isinstance(act, torch.Tensor):
+            act = act.detach().cpu().numpy()
+        act = self.map_action(act)
+        if isinstance(self.action_space, Discrete):
+            # could be an array of shape (), easier to just convert to int
+            act = int(act)  # type: ignore
```
Trinkle23897, line 179:
```suggestion
            return int(act)
```

tianshou/policy/base.py
```diff
@@ -149,13 +153,39 @@ def soft_update(self, tgt: nn.Module, src: nn.Module, tau: float) -> None:
...
+        """Get action as int (for discrete env's) or array (for continuous ones) from
+        an env's observation and info.
+
+        :param obs: observation from the gym's env.
+        :param info: information given by the gym's env.
+        :param state: the hidden state of RNN policy, used for recurrent policy.
+        :return: action as int (for discrete env's) or array (for continuous ones).
+        """
+        # need to add empty batch dimension
+        obs = obs[None, :]
```
Trinkle23897, line 171: nit
```suggestion
        obs = obs[None]
```

tianshou/policy/base.py
MischaPanch: I'd prefer to keep one return, makes it easier for me personally to read the code. If you have a strong preference though, I can add this change

tianshou/data/buffer/base.py
MischaPanch: This change seems less innocent than I anticipated. I'd like to address it in a separate issue, maybe at the same time addressing the TODOs I added to buffer. I am a bit afraid of touching this code since it's at the core of everything. Would you have time to work on it or do some pair programming with me in the near future?

Meanwhile, I'd drop this commit here and open a new draft branch

Trinkle23897: k will push here later

MischaPanch: Ok, then let's resolve it on this branch, will wait for your commits

MischaPanch:
@Trinkle23897 I removed the `batch_size=None` related changes (by force pushing) and moved them to a draft PR #993 . Feel free to take it over.

Would be nice to merge this one soon, as we need the `compute_action` method in downstream code in internal projects

Trinkle23897 (approved):

PR #993, opened 2023-11-16
Support batch_size=None and use it in various scripts

### LINKED ISSUE #986: Better default for batch_size in examples
In various example scripts one finds `batch_size: int = 99999`, which just means that there is no batching, since batch_size exceeds buffer capacity. This should be handled by a better default like `None` or `-1`.

MischaPanch:
Closes #986


Trinkle23897 (approved):
My bad, lgtm

PR #996, opened 2023-11-23
Feature/dataclasses

### LINKED ISSUE #933: Replace dicts by dataclasses in methods that return info
Affects things like `train_step`, `test_step`, various update methods and others.

Methods that consume these outputs have to be adjusted, and the new types should be used for annotations where appropriate

maxhuettenrauch:
This PR adds strict typing to the output of `update` and `learn` in all policies. This will likely be the last large refactoring PR before the next release (0.6.0, not 1.0.0), so it requires some attention. Several difficulties were encountered on the path to that goal:

1. The policy hierarchy is actually "broken" in the sense that the keys of dicts that were output by `learn` did not follow the same enhancement (inheritance) pattern as the policies. This is a real problem and should be addressed in the near future. Generally, several aspects of the policy design and hierarchy might deserve a dedicated discussion.
2. Each policy needs to be generic in the stats return type, because one might want to extend it at some point and then also extend the stats. Even within the source code base this pattern is necessary in many places.
3. The interaction between learn and update is a bit quirky, we currently handle it by having update modify special field inside TrainingStats, whereas all other fields are handled by learn.
4. The IQM module is a policy wrapper and required a TrainingStatsWrapper. The latter relies on a bunch of black magic.

They were addressed by:
1. Live with the broken hierarchy, which is now made visible by bounds in generics. We use type: ignore where appropriate.
2. Make all policies generic with bounds following the policy inheritance hierarchy (which is incorrect, see above). We experimented a bit with nested TrainingStats classes, but that seemed to add more complexity and be harder to understand. Unfortunately, mypy thinks that the code below is wrong, wherefore we have to add `type: ignore` to the return of each `learn`

```python

T = TypeVar("T", bound=int)


def f() -> T:
  return 3
```

3. See above
4. Write representative tests for the `TrainingStatsWrapper`. Still, the black magic might cause nasty surprises down the line (I am not proud of it)...

Closes #933


tianshou/data/collector.py
```diff
@@ -214,7 +215,7 @@ def collect(
             One and only one collection number specification is permitted, either
             ``n_step`` or ``n_episode``.
 
-        :return: A dict including the following keys
+        :return: A CollectStats object including the following keys
 
             * ``n/ep`` collected number of episodes.
             * ``n/st`` collected number of steps.
```
MischaPanch: The docstrings have to be adjusted, no need to tell which keys are contained since this information is in the type

tianshou/data/stats.py
```diff
@@ -0,0 +1,165 @@
+from abc import ABC
+from collections.abc import MutableMapping
+from dataclasses import asdict, dataclass, field, InitVar
+
+import numpy as np
+
+
+# adapted from https://stackoverflow.com/questions/6027558/flatten-nested-dictionaries-compressing-keys
+def flatten_dict(dictionary: dict, parent_key='', separator='/',
```
MischaPanch: Add a docstring pls

tianshou/data/stats.py
```diff
@@ -0,0 +1,165 @@
...
+
+@dataclass(kw_only=True)
+class BaseStats(ABC):
+    """Abstract base class for statistics."""
+
+    def update(self, stats: dict):
+        for k, v in stats.items():
+            if np.isscalar(v):
+                setattr(self, k, v)
+            elif isinstance(v, np.ndarray | list):
```
MischaPanch: This seems a bit fishy

tianshou/data/stats.py
```diff
@@ -0,0 +1,165 @@
...
+        if exclude is not None:
+            stat_dict = {k: v for k, v in stat_dict.items() if k not in exclude}
+        return stat_dict
+
+
+@dataclass(kw_only=True)
+class ArrayStats(BaseStats):
+    """A data structure for storing the statistics of an array."""
+
+    _array: np.ndarray | list[float] = field(repr=False)
```
MischaPanch: This should be adjusted

tianshou/data/collector.py
Trinkle23897: could you also change here?

tianshou/data/stats.py
```diff
@@ -0,0 +1,165 @@
...
+    def update(self, stats: dict):
+        for k, v in stats.items():
+            if np.isscalar(v):
+                setattr(self, k, v)
+            elif isinstance(v, np.ndarray | list):
+                setattr(self, k, ArrayStats(_array=v))
+            elif isinstance(v, BaseStats):
+                getattr(self, k).update(v.to_dict())
+
+    def to_dict(self, exclude: set[str] = None, exclude_arrays=True, exclude_none=True):
```
Trinkle23897: could you use pprint.pformat to override `__repr__`?

tianshou/data/stats.py
```diff
@@ -0,0 +1,165 @@
...
+
+    def __post_init__(self):
+        self.mean: float = np.mean(self._array)
+        """The mean of the array."""
+        self.std: float = np.std(self._array)
+        """The standard deviation of the array."""
+        self.max: float = np.max(self._array)
+        """The maximum of the array."""
+        self.min: float = np.min(self._array)
+        """The minimum of the array."""
```
Trinkle23897: This is pretty nice! Small nit: remove the duplicated docstring since it has already represented above.

tianshou/data/stats.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+@dataclass(kw_only=True)
+class UpdateStats(BaseStats):
```
Trinkle23897: better to move to `trainer/base.py` since it's only been used in trainer?

tianshou/data/stats.py
```diff
@@ -0,0 +1,165 @@
...
+    train_time: float = 0.
+    """The time for learning models."""
+    loss: BaseStats
+    """The loss statistics of the policy learn step."""
+    smoothed_loss: dict = field(default_factory=dict)
+    """The smoothed loss statistics of the policy learn step."""
+
+
+@dataclass(kw_only=True)
+class CollectStats(BaseStats):
```
Trinkle23897: better to move to `collector.py`?

tianshou/data/stats.py
```diff
@@ -0,0 +1,165 @@
...
+            if len(array_lens) > 0:
+                self.lens = ArrayStats(_array=array_lens)
+            else:
+                self.lens = array_lens
+        else:
+            self.lens = None
+
+
+@dataclass(kw_only=True)
+class InfoStats(BaseStats):
```
Trinkle23897: `trainer.py`?

tianshou/data/stats.py
```diff
@@ -0,0 +1,165 @@
...
+    """The total collected step of test collector."""
+    test_episode: int
+    """The total collected episode of test collector."""
+
+    timing: BaseStats
+    """The timing statistics."""
+
+
+@dataclass(kw_only=True)
+class EpochStats(BaseStats):
```
Trinkle23897: also `trainer.py`?

tianshou/policy/imitation/base.py
```diff
@@ -36,6 +37,12 @@ class ImitationPolicy(BasePolicy):
         explanation.
     """
 
+    @dataclass(kw_only=True)
+    class LossStats(BaseStats):
```
Trinkle23897: I think it's better to move all the `LossStats` classes out of the Policy class to make it more explicit instead of a nested one.

Trinkle23897 (commented):
Overall lgtm, thanks for making things better! Could you please update all docs (.rst) and README to apply this change as well?

MischaPanch:
Hey @Trinkle23897, thanks for the quick review! 

This is still in draft status, @maxhuettenrauch and I are doing some internal rounds and going to do some more improvements and simplifications. Once it's done and tests pass, we'll ping you for a second review if you don't mind :)

Trinkle23897:
Of course, no problem! Take your time

tianshou/data/collector.py
```diff
@@ -214,17 +216,14 @@ def collect(
...
-            * ``n/ep`` collected number of episodes.
-            * ``n/st`` collected number of steps.
-            * ``rews`` array of episode reward over collected episodes.
-            * ``lens`` array of episode length over collected episodes.
-            * ``idxs`` array of episode start index in buffer over collected episodes.
-            * ``rew`` mean of episodic rewards.
-            * ``len`` mean of episodic lengths.
-            * ``rew_std`` standard error of episodic rewards.
-            * ``len_std`` standard error of episodic lengths.
+        :return: A dataclass object including the following keys
```
MischaPanch: No need to write this, the type is self-documenting

tianshou/data/collector.py
```diff
@@ -379,26 +379,22 @@ def collect(
...
-        }
+            rews, lens = np.array([]), np.array([], int)
+
+        return CollectStats(
+            n_collected_episodes=episode_count,
+            n_collected_steps=step_count,
+            collect_time=collect_time,
+            collect_speed=step_count / collect_time,
+            rews=rews,
+            rews_stat=SequenceSummaryStats.from_sequence(rews) if episode_count > 0 else None,
```
MischaPanch: is the if-else here really necessary?

tianshou/data/collector.py
```diff
@@ -461,17 +457,14 @@ def collect(
...
-            * ``n/ep`` collected number of episodes.
-            * ``n/st`` collected number of steps.
-            * ``rews`` array of episode reward over collected episodes.
-            * ``lens`` array of episode length over collected episodes.
-            * ``idxs`` array of episode start index in buffer over collected episodes.
-            * ``rew`` mean of episodic rewards.
-            * ``len`` mean of episodic lengths.
-            * ``rew_std`` standard error of episodic rewards.
-            * ``len_std`` standard error of episodic lengths.
+        :return: A dataclass object including the following keys
```
MischaPanch: See above

tianshou/data/stats.py
```diff
@@ -0,0 +1,107 @@
+from collections.abc import Sequence
+from dataclasses import dataclass, field
+
+import numpy as np
+
+
+@dataclass(kw_only=True)
```
MischaPanch: remove annotation (there are no fields)

tianshou/data/stats.py
```diff
@@ -0,0 +1,107 @@
...
+class BaseStats:
+    """A base class for all statistics data structures."""
+
+    def update(self, stats: dict) -> None:
+        for k, v in stats.items():
+            assert hasattr(self, k), f"Unknown key {k} in stats dict {stats}."
```
MischaPanch: turn to ValueError

MischaPanch (changes requested):

MischaPanch (commented):
From my side this is ready

MischaPanch:
@maxhuettenrauch I addressed several TODOs and made mypy happy. Thank you for all the work on this!

@Trinkle23897 @opcode81 @nuance1979 I have updated the PR description and highlighted the design decisions and reasons for them. It's ready to review

I wish everyone a happy rollout into New-Year! :)

MischaPanch (approved):

tianshou/policy/imitation/cql.py
```diff
@@ -373,16 +390,12 @@ def learn(self, batch: RolloutBatchProtocol, *args: Any, **kwargs: Any) -> dict[
...
-        }
-        if self.is_auto_alpha:
-            self.alpha = cast(torch.Tensor, self.alpha)
-            result["loss/alpha"] = alpha_loss.item()
-            result["alpha"] = self.alpha.item()
-        if self.with_lagrange:
-            result["loss/cql_alpha"] = cql_alpha_loss.item()
-            result["cql_alpha"] = cql_alpha.item()
-        return result
+        return CQLTrainingStats(  # type: ignore[return-value]
```
Trinkle23897, line 393: could you explain why do we need `type: ignore` for all TrainingStats?

Trinkle23897 (approved):

Issue #1004, opened 2023-12-06
Use nbqa on notebooks


MischaPanch:
@carlocagnetta Could you look into this? Should be hopefully a low hanging fruit

https://nbqa.readthedocs.io/en/latest/readme.html#readme

carlocagnetta:
@MischaPanch Can you please assign this to me?

carlocagnetta:
Opened PR: #1041

PR #1017, opened 2024-01-03
Refactoring/mypy issues test

dantp-ai:
Improves typing in examples and tests, towards mypy passing there.

Introduces the `SpaceInfo` utility


MischaPanch:
Is this ready for review? If yes, then pls include the tests directory into the poe task (inside pyproject.toml) and make sure CI is green

dantp-ai:
Noted.

Re: status, not ready for review yet. I think I can finish resolving the issues in the test by the end of this weekend.

MischaPanch:
Great, I'm marking it as draft until then. Mark it as ready and assign me once you're done please

MischaPanch:
There's been some changes on master now. They likely won't affect the tests (they do affect examples though) but it would still be better to merge master into your branch before continuing.

dantp-ai:
### Issue 1:

Fixing [attr-defined] for different types of spaces. Here is one example:

```
    env = gym.make(args.task)
    args.state_shape = env.observation_space.shape or env.observation_space.n
    args.action_shape = env.action_space.shape or env.action_space.n
    args.max_action = env.action_space.high[0]  # float
```

that gives this error:
```
test/offline/test_td3_bc.py:77: error: "Space[Any]" has no attribute "n"  [attr-defined]
        args.state_shape = env.observation_space.shape or env.observation_space.n
                                                          ^~~~~~~~~~~~~~~~~~~~~~~
test/offline/test_td3_bc.py:78: error: "Space[Any]" has no attribute "n"  [attr-defined]
        args.action_shape = env.action_space.shape or env.action_space.n
                                                      ^~~~~~~~~~~~~~~~~~
test/offline/test_td3_bc.py:79: error: "Space[Any]" has no attribute "high"  [attr-defined]
        args.max_action = env.action_space.high[0]  # float
```

**Solution 1:**

Check for the `action_space`/`observation_space` instance:

  ```
  if isinstance(env.action_space, Box):
      args.action_shape = env.action_space.shape
      args.max_action = env.action_space.high[0]  # Access the 'high' attribute for the upper bound
      args.action_dim = args.action_shape[0]
  elif isinstance(env.action_space, Discrete):
      args.action_shape = env.action_space.n
      args.max_action = env.action_space.start + env.action_space.n - 1
      args.action_dim = args.n
  else:
      raise NotImplementedError("Action space is not of type `Box` or `Discrete`.")
  ```

I am only checking for `Box` and `Discrete` as an example, but if we choose this solution, we should also consider other spaces such as `Graph`, etc.


**Solution 2:**

Tell mypy to ignore the [attr-defined] error in these cases.


### Outcome

I prefer Solution 1 or something similar to it. What do you think?


### Issue 2:

The issue above repeats itself in different test files. It would be better to move the solution implementation in a  `utils` or `commons` module to be used by all the tests instead of applying the same changes in dozens of files. What do you think?

MischaPanch:
I also prefer solution 1, and feel free to write small util functions for extracting such info from an unknown space into the source code to reduce duplication.

I'm not supper happy that something called `action_shape` can hold an int, is there an easy way to not use the same var name for continuous and discrete cases?

dantp-ai:
I managed to knock off a bunch of issues from the test and examples.

Re: the above, this is my current refactored solution to adding type hints to the Gymnasium space info: https://github.com/thu-ml/tianshou/pull/1017/commits/5f5710f878244c3684708b9200782ea03d787d1a (it supports only the most common spaces for now, `Discrete` and `Box`, and I couldn't find a nice way yet to name the `action_shape` differently :( )

(Original solution is here: (`get_spaces_info()` implementation https://github.com/thu-ml/tianshou/pull/1017/commits/cadb8136bfdae879644561386851da4cb0392f10 and usage of `get_spaces_info()` https://github.com/thu-ml/tianshou/pull/1017/commits/9a5220e424417d0b541ebb77d3fe14fe4d12a7fc)

The issue is present in ~60 files and it leads to ~2-4 issues per file so fixing it will lower the total count substantially.

If you can take a look at it would be helpful, before applying it to the remaining files.

I think after fixing this issue we could maybe also close this PR once it has been reviewed, and open a new one for fixing the remaining issues. They are still quite a few and I don't want to overwhelm the reviewer with so many changes. But I can proceed with fixing all of them in this PR if it is no problem.

**Progress update**

`test`: Found **729** errors in 54 files (checked 56 source files) ==> Found **460** errors in 43 files (checked 58 source files)

`examples`: Found **560** errors in 65 files (checked 64 source files) ==> Found **378** errors in 61 files (checked 65 source files)

* As mentioned, the total count will decrease substantially with solving the space-info issue.

---
P.S.: This week I will spend even more time to get it done. I apologize for any delay caused.

test/utils.py
```diff
@@ -0,0 +1,51 @@
+from collections.abc import Sequence
```
MischaPanch: Would make sense to move this to the library code, right?

test/utils.py
```diff
@@ -0,0 +1,51 @@
...
+from dataclasses import dataclass
+
+import numpy as np
+from gymnasium import spaces
+
+from tianshou.data.collector import CollectStats
+
+
+@dataclass
```
MischaPanch: pls use `kw_only=True`

test/utils.py
```diff
@@ -0,0 +1,51 @@
...
+class SpaceInfo:
+    action_shape: int | Sequence[int]
+    state_shape: int | Sequence[int]
+    action_dim: int
+    state_dim: int
+    min_action: float
+    max_action: float
```
MischaPanch: I suggest that you add a constructor method here called `from_env` or something like that. Then you can delete the `get_space_info` function

test/utils.py
```diff
@@ -0,0 +1,51 @@
...
+            action_dim=int(space.n),
+            state_dim=int(space.n),
+            min_action=float(space.start),
+            max_action=float(space.start + space.n - 1),
+        )
+    else:
+        raise NotImplementedError("Unsupported space type")
+
+
+def print_final_stats(collect_stats: CollectStats) -> None:
```
MischaPanch: where is this used? print statements in tests are not really useful, I think, and it's not importable in the rest of the code

test/utils.py
```diff
@@ -0,0 +1,51 @@
...
+        return SpaceInfo(
+            action_shape=int(space.n),
+            state_shape=int(space.n),
+            action_dim=int(space.n),
+            state_dim=int(space.n),
+            min_action=float(space.start),
+            max_action=float(space.start + space.n - 1),
+        )
+    else:
+        raise NotImplementedError("Unsupported space type")
```
MischaPanch: Log messages should be as informative as possible, otherwise it's difficult to make use of them.

Pls change to

```python
raise NotImplementedError(f"Unsupported space type: {space.__class__}. Currently supported types are Discrete and Box")
```

MischaPanch (changes requested):

test/utils.py
```diff
@@ -0,0 +1,51 @@
...
+
+from tianshou.data.collector import CollectStats
+
+
+@dataclass
+class SpaceInfo:
+    action_shape: int | Sequence[int]
+    state_shape: int | Sequence[int]
+    action_dim: int
+    state_dim: int
```
MischaPanch: state -> obs

Isn't action_dim and state_dim tied to the action_shape and state_shape? I don't fully understand why we need both.

If we do need both, it probably should be a property instead of a field that's being passed.

test/offline/test_cql.py
```diff
@@ -78,16 +81,26 @@ def test_cql(args=get_args()):
...
     env = gym.make(args.task)
-    args.state_shape = env.observation_space.shape or env.observation_space.n
-    args.action_shape = env.action_space.shape or env.action_space.n
-    args.max_action = env.action_space.high[0]  # float
+    action_space = cast(gym.spaces.Box, env.action_space)
+
+    action_space_info = get_space_info(action_space)
+    observation_space_info = get_space_info(env.observation_space)
+
+    args.state_shape = observation_space_info.state_shape
```
MischaPanch: There's some misconception here, the object extracted with `get_space_info(action_space)` should not have a `state_shape` field. 

A better interface might be given from the proposed `from_env` constructor method of SpaceInfo, which can collect both the observation and the action space info. Then you also need only a single SpaceInfo object here to fill the args

MischaPanch:
> If you can take a look at it would be helpful, before applying it to the remaining files.
> 

I took a quick look at the new util methods and wrote some comments on them

> * As mentioned, the total count will decrease substantially with solving the space-info issue.

Yes, that should bring us quite far along the way, thank you!

> P.S.: This week I will spend even more time to get it done. I apologize for any delay caused.

No problem at all, I'm grateful for your help on this!

test/utils.py
dantp-ai: > where is this used?

Many test files print the two collect stats (e.g, https://github.com/thu-ml/tianshou/pull/1017/commits/47d86af8d0f73494d2eda411765686c5f3466cb0). 

The problem is that `returns_stat` and `lens_stat` can be None. So, instead of checking for None in each file individually, I thought I put this check in a function and re-use it.

I agree, printing is not appropriate here. Returning a string is more useful.

We could overwrite the `__str__()` in `CollectStats` and check for None for `returns_stat` and `lens_stat` there, instead of having a separate function called `print_final_stats`.

MischaPanch: Yes, the CollectStats should inherit from the ToStringMixin or a dedicated Mixin that adds a suitable `__repr__`. I'd do that in a followup PR. For now, printing info like this in tests should be discouraged. If you want to fix it, please use a dedicated `log = logging.getLogger(__name__)` in each module and then `log.info` or `log.debug` instead of print. Also, you can just use the CollectStats repr, assuming that it will be improved soon. No need for a separate printing util. Feel free to open an issue for a better repr for CollectStats (and other things) if you want :)

test/utils.py
MischaPanch: I also think it's more of a `ValueError`

test/utils.py
dantp-ai: Good point. I fixed it (and will push asap).

I saw similar things done in the code-base (e.g., in data.stats.py).

test/utils.py
dantp-ai: > state -> obs

I agree, and it is no hassle to change it here, but should we also change then the names of the args in the test files, where they are named `args.state_shape` (that might be too much of refactoring in this PR, I think).

test/utils.py
dantp-ai: Moved it to `tianshou/utils/space_info.py`

test/utils.py
dantp-ai: Fixed.

test/offline/test_cql.py
dantp-ai: Fixed. Now caller calls once to get a space_info with all the attrs from action and observation space it needs.

test/utils.py
MischaPanch: Yes, opening a followup issue. There might be other places in the code where state_shape is used

dantp-ai:
I am not sure what's wrong with the workflow: Ubuntu GPU. It's not related to the code changes. It fails at checkout

> Error: fatal: unable to access 'https://github.com/thu-ml/tianshou/': Failed to connect to github.com port 443: Connection timed out

test/utils.py
dantp-ai: > Also, you can just use the CollectStats repr, assuming that it will be improved soon. No need for a separate printing util.

Yep, I'll do this and open a separate PR afterwards with the improvement.

MischaPanch: Pls wait a moment wit the followup PR, @carlocagnetta and I already wrote a Mixin that will largely resolve it and will be merged to master soon (Carlo is almost finished with bringing mypy to notebooks #1004 )

MischaPanch: I kinda forgot that the solution was on his branch and not on master yet :sweat_smile:

MischaPanch:
> I am not sure what's wrong with the workflow: Ubuntu GPU. It's not related to the code changes. It fails at checkout
> 
> > Error: fatal: unable to access 'https://github.com/thu-ml/tianshou/': Failed to connect to github.com port 443: Connection timed out

Don't worry about it, it's a problem with the runner that is located @Trinkle23897's alma mater in China. I can't fix it, but merging is fine without this test being green (especially for PRs that don't touch anything GPU related)

tianshou/utils/space_info.py
```diff
@@ -0,0 +1,81 @@
...
+from dataclasses import dataclass, field
+
+import numpy as np
+from gymnasium import spaces
+
+
+@dataclass(kw_only=True)
+class ActionSpaceInfo:
+    action_shape: int | Sequence[int]
+    action_dim: int = field(init=False)
```
MischaPanch: Could you pls add dostrings? Especially for this field, I still don't fully understand why it's needed.

You could also instead just add a method `get_action_dim` or a property that reads it it. The post-init thing is not really safe - a user could overwrite action_shape after instantiation and the instance would be "broken" then (action_dim out of sync with action_shape). It's a bit contrived, but still, no need to allow for this possibility

tianshou/utils/space_info.py
```diff
@@ -0,0 +1,81 @@
...
+    obs_dim: int = field(init=False)
+
+    def __post_init__(self) -> None:
+        if isinstance(self.obs_shape, int):
+            self.obs_dim = self.obs_shape
+        else:
+            self.obs_dim = int(self.obs_shape[0])
+
+    @classmethod
+    def from_space(cls, space: spaces.Space) -> "ObservationSpaceInfo":
```
MischaPanch: The return type should be `Self` (here and above)

tianshou/utils/space_info.py
```diff
@@ -0,0 +1,81 @@
...
+            )
+
+
+@dataclass(kw_only=True)
+class SpaceInfo:
+    action_info: ActionSpaceInfo
+    observation_info: ObservationSpaceInfo
+
+    @classmethod
+    def from_env(cls, action_space: spaces.Space, observation_space: spaces.Space) -> "SpaceInfo":
```
MischaPanch: return `Self`

tianshou/utils/space_info.py
MischaPanch: You can inherit from `ToStringMixin` and use it to include properties in the `__str__` and `__repr__`

MischaPanch:
Thx for the changes @dantp-ai, this is almost ready to merge. To make sure that mypy is actually happy, pls include the examples and tests dirs into the poe task. I.e., in pyproject.toml change line 196 to

```toml
type-check = "mypy tianshou examples test"
```

and let's hope for the CI to run through.

When this and @carlocagnetta's work is finished, all code in tianshou will be type-checked - a significant improvement towards ensuring a good future for this project!

tianshou/utils/space_info.py
dantp-ai: >  still don't fully understand why it's needed.
You could also instead just add a method get_action_dim or a property that reads it it.

After looking more into how it is being used, I think we can remove `action_dim` and `obs_dim` and let the caller use the `action_shape`/`obs_shape` appropriately 

(Some algorithms like [ICM take a product of all dimensions](https://github.com/dantp-ai/tianshou/blob/7f81ac73370bdfe398ada927297b0eff84c72de4/test/modelbased/test_ppo_icm.py#L133), so the current definition in `SpaceInfo` is not quite correct, although I suspect that the number of dimensions there is 1 so it works out).

MischaPanch: Sounds good, feel free to remove it

dantp-ai: However, if we remove it, we still need to check in the caller's code if `action/obs_dim` is `int | Sequence[int]` to ensure correct type checking, otherwise indexing the first value will not always work.

My suggestion is that we go ahead and remove `action/obs_dim`, but make `action_shape/obs_shape` of type `Sequence[int]` only. Then the caller can define `action/obs_dim` however they need (take the first value, multiply the dimension, ...).

dantp-ai: However, if we remove it, we still need to check in the caller's code if `action/obs_dim` is `int | Sequence[int]` to ensure correct type checking, otherwise indexing the first value will not always work.

We could remove `action/obs_dim`, but make `action_shape/obs_shape` of type `Sequence[int]` only. Then the caller can define `action/obs_dim` however they need (take the first value, multiply the dimension, ...).

Other networks accept either `int | Sequence[int]` (`Net`) or `Sequence[int]` (`Actor`) so it won't break APIs.

MischaPanch: What is the semantics here? I guess it currently only supports discrete or a flat Box, in which case the dim is `action_shape[0]`. For a general Box, I suppose the correct semantics is the product of the dimensions.  We can keep it, but better as a method (`get_action_dim`) that returns either `.n` or the product of dims. What do you think?

dantp-ai: Fixed this, and added doc strings. I think we are very close to use this util to fix the issue on space info typing that is present in all the other test files.

tianshou/utils/space_info.py
```diff
@@ -0,0 +1,111 @@
...
+    """The largest allowable action or in the continuous case the upper bound for allowable action value."""
+
+    @property
+    def action_dim(self) -> int:
+        """Return the number of distinct actions an agent can take it its action space."""
+        if isinstance(self.action_shape, int):
+            return self.action_shape
+        elif isinstance(self.action_shape, Sequence) and self.action_shape:
+            return int(np.prod(self.action_shape))
+        else:
```
MischaPanch, line 28: We're almost there. Accessing a property should never raise an Exception, it's highly unexpected from a user perspective.

In fact, checking for None is not necessary - the type annotation only allows for int and Sequence[int], so nothing can go wrong. Pls remove the checks and the error. Note that we have to trust people passing the annotated types, doesn't make sense to check all of them at runtime. Here an proper IDE or mypy would tell the user if they passed an wrong type for action_shape

NB: if you wanted to raise an error here, it would be a RuntimeError, since it's not caused by user-provided values to the call.

tianshou/utils/space_info.py
```diff
@@ -0,0 +1,111 @@
...
+            raise ValueError("Invalid action_shape: {self.action_shape}.")
+
+    @classmethod
+    def from_space(cls, space: spaces.Space) -> Self:
+        """Return the attributes of the action space based on the instance type of the space."""
```
MischaPanch:
```suggestion
        """Instantiate the `ActionSpaceInfo` object from a `Space`, supported spaces are Box and Discrete."""
```

tianshou/utils/space_info.py
```diff
@@ -0,0 +1,111 @@
...
+                action_shape=int(space.n),
+                min_action=float(space.start),
+                max_action=float(space.start + space.n - 1),
+            )
+        else:
+            raise ValueError(
+                f"Unsupported space type: {space.__class__}. Currently supported types are Discrete and Box.",
+            )
+
+    def _tostring_additional_entries(self) -> dict[str, Any]:
```
MischaPanch, line 51: I think that's unnecessary, properties are included by default

tianshou/utils/space_info.py
```diff
@@ -0,0 +1,111 @@
...
+        elif isinstance(space, spaces.Discrete):
+            return cls(
+                obs_shape=int(space.n),
+            )
+        else:
+            raise ValueError(
+                f"Unsupported space type: {space.__class__}. Currently supported types are Discrete and Box.",
+            )
+
+    def _tostring_additional_entries(self) -> dict[str, Any]:
```
MischaPanch, line 86: same as above

tianshou/utils/space_info.py
```diff
@@ -0,0 +1,111 @@
...
+
+@dataclass(kw_only=True)
+class ObservationSpaceInfo(ToStringMixin):
+    """A data structure for storing the different attributes of the observation space."""
+
+    obs_shape: int | Sequence[int]
+    """The shape of the observation space."""
+
+    @property
+    def obs_dim(self) -> int:
```
MischaPanch, line 63: same as above

tianshou/utils/space_info.py
MischaPanch: Ahh, now I see - you are checking for Sequence being empty, not the type.

Here it would be enough to document that the sequence should not be empty in the docstring. If you think validation of that is important, you could either do that in post-init, or instead of the property make a method `get_action_dim()` that raises a RuntimeError. For me it is fine to expect that action_space is non-empty and just documenting it, I have a hard time imagining an env with no actions

tianshou/utils/space_info.py
MischaPanch: Structure looks good to me now, thanks! I added some minor comments, feel free to "roll out" this solution to all test-files and examples after addressing them

tianshou/utils/space_info.py
dantp-ai: I tried without and properties are curiously not included by default (maybe something to correct !?).

In `ToStringMixing`, `_tostring_object_info` is calling the default `_tostring_additional_entries` which returns an empty dict (cf. https://github.com/thu-ml/tianshou/blob/4756ee80ff11cd8692aef3752f35c0af60a452e8/tianshou/utils/string.py#L318 )

MischaPanch:
space_info is finished now, thanks for addressing the comments! Feel free to use it everywhere to fix the typing issues

tianshou/utils/space_info.py
```diff
@@ -0,0 +1,107 @@
...
+class SpaceInfo(ToStringMixin):
+    """A data structure for storing the attributes of both the action and observation space."""
+
+    action_info: ActionSpaceInfo
+    """Stores the attributes of the action space."""
+    observation_info: ObservationSpaceInfo
+    """Stores the attributes of the observation space."""
+
+    @classmethod
+    def from_env(cls, action_space: spaces.Space, observation_space: spaces.Space) -> Self:
```
MischaPanch: Ah, sorry, one last comment: `from_env` should either take a `gym.Env` as input, ot be called `from_spaces`. If you want, you could rename this to `from_spaces` and add a second constructor `from_env` that uses the first one. I think that would be the cleanest way

MischaPanch:
Let me know when you're done and the additional checks have been included in CI, then I'd have a final look and we can merge it
@dantp-ai

dantp-ai:
Yep, for sure! I'll let you know once it's ready for review or some additional thing needs a closer look from your side. 

I'm working on the updates using SpaceInfo at the moment.

dantp-ai:
I'll have a big update coming today (will ping again here), that we should consider for review toward closing this PR.

We can solve any remaining mypy issues (not too many left) in a new PR.

carlocagnetta:
I just opened the other PR running mypy on notebooks #1041. Checking your changes I don't think this should affect you, but it might rather resolve some more mypy errors.
Note possible conflict in the pyproject since I add a poe sub command to run mypy and ruff checks on notebooks.

dantp-ai:
I pushed now the changes for review.

I didn't add examples and test dir to `poe type-check` because there are still some mypy issues open.

The PR is quite big already, and I thought we could maybe close this one now after the review. Then I can solve the other issues in a second PR.

examples/common.py
```diff
@@ -0,0 +1,3 @@
+from tianshou.highlevel.logger import LoggerFactoryDefault
+
+logger_factory = LoggerFactoryDefault()
```
MischaPanch, line 3: I'm merging this PR, but for the next PR: this introduces a mutable default throughout the code. Why was this neccessary?

examples/discrete/discrete_dqn.py
```diff
@@ -25,12 +28,14 @@ def main():
...
-    action_shape = env.action_space.shape or env.action_space.n
+    env.action_space = cast(gym.spaces.Discrete, env.action_space)
+    space_info = SpaceInfo.from_env(env)
+    state_shape = space_info.observation_info.obs_shape
+    action_shape = space_info.action_info.action_shape
     net = Net(state_shape=state_shape, action_shape=action_shape, hidden_sizes=[128, 128, 128])
     optim = torch.optim.Adam(net.parameters(), lr=lr)
 
-    policy = ts.policy.DQNPolicy(
+    policy: ts.policy.DQNPolicy = ts.policy.DQNPolicy(
```
MischaPanch, line 38: again for future PR: why is this type of annotation necessary? Mypy shouldn't need it, typed python is still not java :)

MischaPanch (approved):
Thanks a lot for this work! Sorry for my delayed reaction, I was on vacation till today. Merging this now, and looking forward to the follow-up that will close the remaining mypy issues

Note: at many places there is something like `policy: GailPolicy = GailPolicy(...)`. I don't think this is required, am I wrong?

examples/discrete/discrete_dqn.py
dantp-ai: I also found it weird since there is no ambiguity of which policy to use, but mypy was throwing this error in these contexts:

```
examples/discrete/discrete_dqn.py:38: error: Need type annotation for "policy"  [var-annotated]
        policy = ts.policy.DQNPolicy(
                 ^
Found 1 error in 1 file (checked 1 source file)
```

I'll check in more depth why it complains.

examples/common.py
dantp-ai: > Why was this necessary?

The examples create a logger using the same code that `LoggerFactoryDefault.create_logger` already implements.

To be DRY, my idea was to re-use the LoggerFactory instead of duplicating the code.

Furthermore, I thought it is cleaner to have one single factory where each example is re-using it to create its own logger.

One alternative solution, using this approach, is to instantiate a `LoggerFactoryDefault` per example file.

MischaPanch: One instantiation per file sounds good to me

examples/discrete/discrete_dqn.py
dantp-ai: Resolved, cf. https://github.com/thu-ml/tianshou/pull/1077#discussion_r1546459488.

examples/common.py
dantp-ai: Resolved, cf. https://github.com/thu-ml/tianshou/pull/1077#discussion_r1547404024.

PR #1022, opened 2024-01-12
Improvements in README and high-level API

opcode81:
- [x] I have added the correct label(s) to this Pull Request or linked the relevant issue(s)
- [x] I have provided a description of the changes in this Pull Request
- [x] I have added documentation for my changes

This makes several largely unrelated improvements in the high-level API and in the README.

Main improvements in high-level API:
  * Improve naming in trainer-related abstractions, moved some classes from examples to the library
  * Improve environment factory abstraction
  * Some bug-fixes

Main changes in README:
  * Add high-level example and update procedural/low-level example
  * Improve language/wording

It makes the most sense to **review this at the commit level** and to **merge it without squashing**.


examples/atari/atari_wrapper.py
```diff
@@ -373,23 +388,49 @@ def make_atari_env(task, seed, training_num, test_num, **kwargs):
...
+        return wrap_deepmind(
+            env,
+            episode_life=is_train,
+            clip_rewards=is_train,
             frame_stack=self.frame_stack,
+            scale=self.scale,
         )
-        return DiscreteEnvironments(env=env, train_envs=train_envs, test_envs=test_envs)
+
+    class EnvPoolFactory(EnvPoolFactory):
```
MischaPanch, line 385: Rename to AtariEnvPoolFactory?

tianshou/highlevel/env.py
```diff
@@ -260,7 +269,156 @@ def get_type(self) -> EnvType:
...
+        match EnvType.from_env(env):
+            case EnvType.DISCRETE:
+                return DiscreteEnvironments(env, train_envs, test_envs)
+            case EnvType.CONTINUOUS:
+                return ContinuousEnvironments(env, train_envs, test_envs)
+            case _:
+                raise ValueError
+
+
+class EnvFactoryGymnasium(EnvFactory):
```
MischaPanch: The naming is a bit weird, I think. The main purpose of this class is to be able to create an envpool env if appropriate/possible, and have a fallback otherwise. However, the name suggests that it has something to do with gymnasium, like a specification of EnvFactory, which then would seem to not have anything to do with gymnasium. On top of that, there is an EnvPoolFactory, which is not an EnvFactory itself and doesn't have a common base with it, despite having a similar interface and purpose.

How about something like EnvFactoryGymnasium -> EnvFactoryEnvPoolReady ?

In principle, EnvPoolFactory can be an EnvFactory - it can implement create_env from the task.

The purpose of the classes could be documented more clearly - that EnvFactoryEnvPoolReady is a factory with multiple fallback cases, allowing you to create envpool accelerated venvs if desired and possible, but with fallback to a safe env creation if not. The main use case of EnvPoolFactory is to be used within this EnvFactoryEnvPoolReady. And so on,

The way default behavior is handled also deserves some discussion, I believe. If a user explicitly passes an envpool_factory, I wouldn't expect the code to just run through with a warning if envpool is not installed. Instead, I think it should break then.

MischaPanch (changes requested):
First part of review, I haven't viewed all changes and will continue tomorrow

examples/atari/atari_wrapper.py
opcode81: Why? The class is internal to `AtariEnvFactory`. Doubling the "Atari" would be redundant. It is `AtariEnvFactory`'s specific adaptation of `EnvPoolFactory`. The class can only ever be referenced via the qualified name `AtariEnvFactory.EnvPoolFactory`.

tianshou/highlevel/env.py
opcode81: I think you may have misread the purpose of the classes:

* `EnvFactoryGymnasium` is, as pointed out in the docstring, for the creation of any environments that have been directly registered with gymnasium and therefore can be constructed via `gymnasium.make`, implementing creation via this interface. It is the class users will want to use for all such cases. Since envpool also supports many such standard environments, it has envpool support built-in, but this is an _add-on_, not the main point of the class. Note that the envpool factory is completely optional, whereas `venv_type` is not! So it is really the other way round: the envpool factory is an optional alternative; the `venv_type` is the "regular" way of doing it (which however, becomes the fallback in the case where envpool is being used).
I wasn't entirely happy with the name either but I couldn't think of a better one. This will be a widely used class, so its name shouldn't be too "out there". (And it does have much more to do with gymnasium than its base class, because it actually calls gymnasium functions to implement the factory rather than just using a gymnasium-defined output type in one of its interfaces. In untyped Python, the base class would have absolutely nothing to do with gymnasium, so there's that.)
* `EnvPoolFactory` is only a helper class to support envpool-based construction. It has its own abstraction such that envpool-based construction can be overridden if needed (_strategy pattern_) - without affecting the non-envpool case. While it is, at present, only used in conjunction with `EnvFactoryGymnasium`, it has a well-defined purpose which could be applied outside of this context (so I prefer to not document limitations that aren't actually there). 
`EnvPoolFactory` probably shouldn't be an `EnvFactory` because envpool always creates vectorized environments not plain `gym.Env`s (though I think it could work via duck typing with n=1); `EnvPoolFactory` implements `EnvFactory.create_venv` not `EnvFactory.create_env`, as is evident from its usage in `EnvFactoryGymnasium`. Also, envpool does not support any type of rendering afaik, so it is not suitable for creating environments for "watching" and should really only be applied for training. 
(On a largely unlretated note, I was thinking of adding an additional option whether to even use envpool if it is installed to give the user more control; in the Atari subclass there is perhaps too little control.)

Regarding default behaviour: The behaviour is clearly documented, so a user cannot rightly complain about unexpected behaviour. There is, importantly, no undocumented "default behaviour"! Note that I am not triggering a warning but a debug message, because expected behaviour should _never_ trigger a warning. (Incidentally, one commit in this PR removes an inappropriate warning for precisely that reason.)
However, we could discuss whether the implemented and documented behaviour is indeed the behaviour we want. (I was simply going with what existing applications do and how they handle it.)

MischaPanch (dismissed):

MischaPanch (approved):

PR #1032, opened 2024-01-23
Naming and typing improvements in Actor/Critic/Policy forwards

### LINKED ISSUE #917: Rename logits to `model_output`
The TODO in the forward of `PGPolicy`  about renaming `logits` should be addressed. It will likely affect the naming in `ModelOutputBatchProtocol` and in other places, as well as docstrings. The exact usage of the model outputs should be clarified everywhere (they are passed to a torch `Distribution` which is then used for sampling or for computing `log_prob`).

_Originally posted by @Trinkle23897 in https://github.com/thu-ml/tianshou/pull/912#discussion_r1300865261_

arnaujc91:
Closes #917 

### Internal Improvements
- Better variable names related to model outputs (logits, dist input etc.). #1032
- Improved typing for actors and critics, using Tianshou classes like `Actor`, `ActorProb`, etc., 
instead of just `nn.Module`. #1032
- Added interfaces for most `Actor` and `Critic` classes to enforce the presence of `forward` methods. #1032
- Simplified `PGPolicy` forward by unifying the `dist_fn` interface (see associated breaking change). #1032
- Use `.mode` of distribution instead of relying on knowledge of the distribution type. #1032

### Breaking Changes

- Changed interface of `dist_fn` in `PGPolicy` and all subclasses to take a single argument in both
continuous and discrete cases. #1032


arnaujc91:
@MischaPanch as of now if you agree with the naming there will be other places, as you mentioned, where logits should be replaced. So will continue once we agree the naming is fine for you.

tianshou/policy/modelfree/pg.py
```diff
@@ -190,18 +190,19 @@ def forward(
...
         #  assumptions about the order of the output and on distribution type
-        logits, hidden = self.actor(batch.obs, state=state, info=batch.info)
-        if isinstance(logits, tuple):
-            dist = self.dist_fn(*logits)
+        # See https://pytorch.org/docs/stable/distributions.html
+        action_dist_input, hidden = self.actor(batch.obs, state=state, info=batch.info)
+        if isinstance(action_dist_input, tuple):
+            dist = self.dist_fn(*action_dist_input)
         else:
-            dist = self.dist_fn(logits)
```
arnaujc91, line 184: Also i am surprised to call it `logits`, in the Pytorch documentation https://pytorch.org/docs/stable/distributions.html#categorical the first positional argument is always `probs`.

arnaujc91:
`ModelOutputBatchProtocol` might need to be modified too if changes are accepted.

tianshou/policy/modelfree/pg.py
```diff
@@ -190,18 +190,19 @@ def forward(
...
+            dist = self.dist_fn(action_dist_input)
 
         # in this case, the dist is unused!
         if self.deterministic_eval and not self.training:
-            act = self._get_deterministic_action(logits)
+            act = self._get_deterministic_action(action_dist_input)
         else:
             act = dist.sample()
-        result = Batch(logits=logits, act=act, state=hidden, dist=dist)
+        result = Batch(logits=action_dist_input, act=act, state=hidden, dist=dist)
```
MischaPanch: Yes, the DistBatchProtocol should be adjusted everywhere. Note that you then have to construct the Batch with
`action_dist_input=action_dist_input`, and you will need to scan the whole codebase for places where the logits attribute of a Batch is used (which can be through `.logits` or through `["logits"]`)

tianshou/policy/modelfree/pg.py
opcode81: Note, however, that the PyTorch class can also take `logits` as an argument (for the case where the inputs are unnormalised log probabilities) - and some implementations do pass `logits` explicitly (using a keyword argument).
Importantly, `dist_fn` need not be a constructor such as `Categorical` but could also something like `lambda x: Categorical(logits=x)`.

In fact, the documentation in `PGPolicy` has led me to implement the default categorical distribution factory in the high-level API using `logits` instead of `probs`. If this is, in fact, not appropriate, then we should change the default. In particular, if there are activation functions at play that prevent the required values from being output, then the use of one instead of the other can have a fatal effect.

arnaujc91: Hi @opcode81 

But here

https://github.com/thu-ml/tianshou/blob/69a0e6bbd788509fe26bd93e142d0e78228114d6/test/discrete/test_pg.py#L72-L77

and here

https://github.com/thu-ml/tianshou/blob/69a0e6bbd788509fe26bd93e142d0e78228114d6/test/continuous/test_npg.py#L97-L105

are the places where i found test distributions ([Normal](https://pytorch.org/docs/stable/distributions.html#normal) and [Categorical](https://pytorch.org/docs/stable/distributions.html#categorical)), in both cases the first positional argument will be used here (no keyword arguments used):

https://github.com/thu-ml/tianshou/blob/69a0e6bbd788509fe26bd93e142d0e78228114d6/tianshou/policy/modelfree/pg.py#L193-L197

The high level implementation is this one right?

https://github.com/thu-ml/tianshou/blob/69a0e6bbd788509fe26bd93e142d0e78228114d6/tianshou/highlevel/agent.py#L244-L278

Where can I see you had to use `logits` instead of `probs` ? i assume the `dist_fn` is passed in the `**kwargs` of line 277.

Sorry if there is something confusing, i am trying to understand better your last comment. Thanks :)

tianshou/policy/modelfree/pg.py
arnaujc91: So I assume you are fine with the new name right?

MischaPanch: yes, action_dist_input sounds like the right name, thanks!

tianshou/policy/modelfree/pg.py
opcode81: In the high-level API, the starting point is the parameter object for PG, which contains a `dist_fn` parameter.
The kwargs you mentioned are generated from this object (see line 264 in your last quote).

https://github.com/thu-ml/tianshou/blob/4756ee80ff11cd8692aef3752f35c0af60a452e8/tianshou/highlevel/params/policy_params.py#L310-L331

By default, the parameter will be an instance of `DistributionFunctionFactoryDefault` (as described in the docstring),

https://github.com/thu-ml/tianshou/blob/4756ee80ff11cd8692aef3752f35c0af60a452e8/tianshou/highlevel/params/dist_fn.py#L36-L44

and for the discrete case, you get `DistributionFunctionFactoryCategorical`, which is implemented like this:

https://github.com/thu-ml/tianshou/blob/4756ee80ff11cd8692aef3752f35c0af60a452e8/tianshou/highlevel/params/dist_fn.py#L16-L23

Note that I didn't "have to" use logits. I used logits in this implementation only because that's what the docstring in `PGPolicy` suggested to me:

https://github.com/thu-ml/tianshou/blob/4756ee80ff11cd8692aef3752f35c0af60a452e8/tianshou/policy/modelfree/pg.py#L46-L51

arnaujc91:
Hi guys, these places i think logits 

```
logits --> action_dist_input
```

should also be reformated:

ImitationPolicy Class
https://github.com/thu-ml/tianshou/blob/4756ee80ff11cd8692aef3752f35c0af60a452e8/tianshou/policy/imitation/base.py#L74-L78

DiscreteSACPolicy Class
https://github.com/thu-ml/tianshou/blob/4756ee80ff11cd8692aef3752f35c0af60a452e8/tianshou/policy/modelfree/discrete_sac.py#L109-L115

SACPolicy Class
https://github.com/thu-ml/tianshou/blob/4756ee80ff11cd8692aef3752f35c0af60a452e8/tianshou/policy/modelfree/sac.py#L176-L198

REDQPolicy Class: Also here loc_scale maybe should be renamed? @MischaPanch 
https://github.com/thu-ml/tianshou/blob/4756ee80ff11cd8692aef3752f35c0af60a452e8/tianshou/policy/modelfree/redq.py#L147-L166

Net Class
https://github.com/thu-ml/tianshou/blob/4756ee80ff11cd8692aef3752f35c0af60a452e8/tianshou/utils/net/common.py#L258-L284

Actor Class
https://github.com/thu-ml/tianshou/blob/4756ee80ff11cd8692aef3752f35c0af60a452e8/tianshou/utils/net/continuous.py#L74-L85

Actor Class
https://github.com/thu-ml/tianshou/blob/4756ee80ff11cd8692aef3752f35c0af60a452e8/tianshou/utils/net/discrete.py#L69-L82

So basically for all actors and all policies that uses actors. Also seems the Net class which i have to figure out how it is properly related to Policies and Actors.

Any objections? @MischaPanch @opcode81

MischaPanch:
> REDQPolicy Class: Also here loc_scale maybe should be renamed? @MischaPanch
> 
> https://github.com/thu-ml/tianshou/blob/4756ee80ff11cd8692aef3752f35c0af60a452e8/tianshou/policy/modelfree/redq.py#L147-L166
> 

What's wrong with the loc_scale?

> So basically for all actors and all policies that uses actors. Also seems the Net class which i have to figure out how it is properly related to Policies and Actors.
> 
> Any objections? @MischaPanch @opcode81

Looks good, go ahead! But you should also rename things inside of some BatchPrototypes, as parrtially discussed above

arnaujc91:
> > REDQPolicy Class: Also here loc_scale maybe should be renamed? @MischaPanch
> > https://github.com/thu-ml/tianshou/blob/4756ee80ff11cd8692aef3752f35c0af60a452e8/tianshou/policy/modelfree/redq.py#L147-L166
> 
> What's wrong with the loc_scale?
> 
> > So basically for all actors and all policies that uses actors. Also seems the Net class which i have to figure out how it is properly related to Policies and Actors.
> > Any objections? @MischaPanch @opcode81
> 
> Looks good, go ahead! But you should also rename things inside of some BatchPrototypes, as parrtially discussed above

Because the logical flow i have seen is usually 
```
self.policy -> dist
```
so the output of the policy network (torch `forward` method) is usually a tuple, where the second element is some `state` and the first element is what you will always input to the Torch distribution as the code demonstrates:

```python
loc_scale, h = self.actor(batch.obs, state=state, info=batch.info) 
loc, scale = loc_scale 
dist = Independent(Normal(loc, scale), 1) 
```
I just wanted to keep consistency. If i call it `action_dist_input` somewhere because is the input of the Torch distribution i should also do it anywhere else where thats the case too. That was my thought.

MischaPanch:
> I just wanted to keep consistency. If i call it `action_dist_input` somewhere because is the input of the Torch distribution i should also do it anywhere else where thats the case too. That was my thought.

In this case it is not passed to a general action dist, but to a gaussian, where it is used as loc and scale. So here the name `loc_scale` is the most appropriate one, I think. Names should always reflect the semantics of the variable as precisely as possible

arnaujc91:
Hmm I bet that in the case of a continuous actor:

https://github.com/thu-ml/tianshou/blob/4756ee80ff11cd8692aef3752f35c0af60a452e8/tianshou/utils/net/continuous.py#L74-L85

The output of line 85 that is named logits is also a mean and standard deviation, no? But is still called `logits` and not `loc_scale `. Maybe i am misunderstanding something.

MischaPanch:
> The output of line 85 that is named logits is also a mean and standard deviation, no? But is still called `logits` and not `loc_scale `. Maybe i am misunderstanding something.

here you don't know how the output of forward will be used (into which kind of continuous action_dist it would go), so you shouldn't call it `loc_scale`. In the code above, on the other hand, the variable is fed into a Gaussian as loc and scale, so the semantics are clear.

If tianshou would only support Gaussians for continuous actors, then the output of the continuous `Actor` should indeed be `loc_scale`. *For sure it should never be logits* - please change that as part of your PR!

The fact that tianshou right now implicitly depends on the `action_dist_input` being of the `loc_scale` format in some places is a separate problem and deserves a separate issue. Fixing that is outside of the scope.

In this PR, the goal should be to make the naming at least a bit better and consistent with the current interfaces. If you know that the variable has to be of the form `(loc, scale)` in some place (like in the example above), then it should be named as such. If you only know that it will be added to a Batch's field called `action_dist_input`, then the var should be called `action_dist_input`, b/c in principle tianshou's current interfaces don't allow you to infer a more precise semantics.

That's what I mean when I say the the var names should reflect semantics as precisely as possible

arnaujc91:
Understood, so wherever there are `logits` appearing that can be interpreted as `action_dist_input` i will make the replacement 

```
logits -> action_dist_input
```
but if already a more accurate name (e.g. `loc_scale`) is already given, i will leave it untouched. I think we agree here.

tianshou/policy/modelfree/pg.py
```diff
@@ -190,18 +190,19 @@ def forward(
...
+        if isinstance(action_dist_input, tuple):
+            dist = self.dist_fn(*action_dist_input)
         else:
-            dist = self.dist_fn(logits)
+            dist = self.dist_fn(action_dist_input)
 
         # in this case, the dist is unused!
         if self.deterministic_eval and not self.training:
-            act = self._get_deterministic_action(logits)
+            act = self._get_deterministic_action(action_dist_input)
```
arnaujc91: @MischaPanch I am thinking about changing this line to:

```python
# Instead of using the action_dist_input use dist
act = self._get_deterministic_action(dist)
# and then use dist.mode or torch.argmax(dist.probs)
```
I think is much more compact if the `act` variable comes *always* from a distribution and not from raw output (i guess simply unnormalized probabilities or mean, std_dev).

The code is also more clear: if the action is deterministic i take either the mode or the most probable action from the distribution and if not i just sample from the distribution. Then it is clear that in both cases the action comes from the distribution.

tianshou/policy/imitation/discrete_bcq.py
```diff
@@ -34,8 +34,7 @@ class DiscreteBCQTrainingStats(DQNTrainingStats):
 class DiscreteBCQPolicy(DQNPolicy[TDiscreteBCQTrainingStats]):
     """Implementation of discrete BCQ algorithm. arXiv:1910.01708.
 
-    :param model: a model following the rules in
-        :class:`~tianshou.policy.BasePolicy`. (s -> q_value)
+    :param model: a model following the rules (s -> action_values_BA)
```
MischaPanch:
```suggestion
    :param model: a model following the rules (s_B -> action_values_BA)
```

tianshou/policy/imitation/discrete_cql.py
```diff
@@ -25,8 +25,7 @@ class DiscreteCQLTrainingStats(QRDQNTrainingStats):
 class DiscreteCQLPolicy(QRDQNPolicy[TDiscreteCQLTrainingStats]):
     """Implementation of discrete Conservative Q-Learning algorithm. arXiv:2006.04779.
 
-    :param model: a model following the rules in
-        :class:`~tianshou.policy.BasePolicy`. (s -> logits)
+    :param model: a model following the rules (s -> action_values_BA)
```
MischaPanch:
```suggestion
    :param model: a model following the rules (s_B -> action_values_BA)
```

tianshou/policy/imitation/discrete_crr.py
```diff
@@ -26,8 +26,9 @@ class DiscreteCRRTrainingStats(PGTrainingStats):
 class DiscreteCRRPolicy(PGPolicy[TDiscreteCRRTrainingStats]):
     r"""Implementation of discrete Critic Regularized Regression. arXiv:2006.15134.
 
-    :param actor: the actor network following the rules in
-        :class:`~tianshou.policy.BasePolicy`. (s -> logits)
+    :param actor: the actor network following the rules:
+        If `self.action_type == "discrete"`: (`s` ->`action_values_BA`).
```
MischaPanch:
```suggestion
        If `self.action_type == "discrete"`: (`s_B` ->`action_values_BA`).
```

tianshou/policy/imitation/discrete_crr.py
```diff
@@ -26,8 +26,9 @@ class DiscreteCRRTrainingStats(PGTrainingStats):
...
+        If `self.action_type == "continuous"`: (`s` -> `dist_input_BD`).
```
MischaPanch:
```suggestion
        If `self.action_type == "continuous"`: (`s_B` -> `dist_input_BD`).
```

tianshou/policy/imitation/gail.py
```diff
@@ -32,7 +32,9 @@ class GailTrainingStats(PPOTrainingStats):
 class GAILPolicy(PPOPolicy[TGailTrainingStats]):
     r"""Implementation of Generative Adversarial Imitation Learning. arXiv:1606.03476.
 
-    :param actor: the actor network following the rules in BasePolicy. (s -> logits)
+    :param actor: the actor network following the rules:
+        If `self.action_type == "discrete"`: (`s` ->`action_values_BA`).
```
MischaPanch:
```suggestion
        If `self.action_type == "discrete"`: (`s_B` ->`action_values_BA`).
```

tianshou/policy/imitation/gail.py
```diff
@@ -32,7 +32,9 @@ class GailTrainingStats(PPOTrainingStats):
...
+        If `self.action_type == "continuous"`: (`s` -> `dist_input_BD`).
```
MischaPanch:
```suggestion
        If `self.action_type == "continuous"`: (`s_B` -> `dist_input_BD`).
```

tianshou/policy/modelfree/a2c.py
```diff
@@ -30,7 +32,9 @@ class A2CTrainingStats(TrainingStats):
 class A2CPolicy(PGPolicy[TA2CTrainingStats], Generic[TA2CTrainingStats]):  # type: ignore[type-var]
     """Implementation of Synchronous Advantage Actor-Critic. arXiv:1602.01783.
 
-    :param actor: the actor network following the rules in BasePolicy. (s -> logits)
+    :param actor: the actor network following the rules:
+        If `self.action_type == "discrete"`: (`s` ->`action_values_BA`).
```
MischaPanch:
```suggestion
        If `self.action_type == "discrete"`: (`s_B` ->`action_values_BA`).
```

tianshou/policy/modelfree/a2c.py
```diff
@@ -30,7 +32,9 @@ class A2CTrainingStats(TrainingStats):
...
+        If `self.action_type == "continuous"`: (`s` -> `dist_input_BD`).
```
MischaPanch:
```suggestion
        If `self.action_type == "continuous"`: (`s_B` -> `dist_input_BD`).
```

tianshou/policy/modelfree/c51.py
```diff
@@ -23,8 +23,7 @@ class C51TrainingStats(DQNTrainingStats):
 class C51Policy(DQNPolicy[TC51TrainingStats], Generic[TC51TrainingStats]):
     """Implementation of Categorical Deep Q-Network. arXiv:1707.06887.
 
-    :param model: a model following the rules in
-        :class:`~tianshou.policy.BasePolicy`. (s -> logits)
+    :param model: a model following the rules (s -> action_values_BA)
```
MischaPanch:
```suggestion
    :param model: a model following the rules (s_B -> action_values_BA)
```

tianshou/policy/modelfree/discrete_sac.py
```diff
@@ -25,8 +26,7 @@ class DiscreteSACTrainingStats(SACTrainingStats):
 class DiscreteSACPolicy(SACPolicy[TDiscreteSACTrainingStats]):
     """Implementation of SAC for Discrete Action Settings. arXiv:1910.07207.
 
-    :param actor: the actor network following the rules in
-        :class:`~tianshou.policy.BasePolicy`. (s -> logits)
+    :param actor: the actor network following the rules (s -> dist_input_BD)
```
MischaPanch:
```suggestion
    :param actor: the actor network following the rules (s_B -> dist_input_BD)
```

tianshou/policy/modelfree/dqn.py
```diff
@@ -189,7 +189,7 @@ def forward(
         :return: A :class:`~tianshou.data.Batch` which has 3 keys:
 
             * ``act`` the action.
-            * ``logits`` the network's raw output.
+            * ``action_values`` the network's raw output.
```
MischaPanch:
```suggestion
            * ``logits`` the network's raw output.
```

tianshou/policy/modelfree/fqf.py
```diff
@@ -27,8 +27,7 @@ class FQFTrainingStats(QRDQNTrainingStats):
 class FQFPolicy(QRDQNPolicy[TFQFTrainingStats]):
     """Implementation of Fully-parameterized Quantile Function. arXiv:1911.02140.
 
-    :param model: a model following the rules in
-        :class:`~tianshou.policy.BasePolicy`. (s -> logits)
+    :param model: a model following the rules (s -> action_values_BA)
```
MischaPanch:
```suggestion
    :param model: a model following the rules (s_B -> action_values_BA)
```

tianshou/policy/modelfree/iqn.py
```diff
@@ -29,8 +29,7 @@ class IQNTrainingStats(QRDQNTrainingStats):
 class IQNPolicy(QRDQNPolicy[TIQNTrainingStats]):
     """Implementation of Implicit Quantile Network. arXiv:1806.06923.
 
-    :param model: a model following the rules in
-        :class:`~tianshou.policy.BasePolicy`. (s -> logits)
+    :param model: a model following the rules (s -> action_values_BA)
```
MischaPanch:
```suggestion
    :param model: a model following the rules (s_B -> action_values_BA)
```

tianshou/policy/modelfree/pg.py
```diff
@@ -40,8 +48,9 @@ class PGTrainingStats(TrainingStats):
 class PGPolicy(BasePolicy[TPGTrainingStats], Generic[TPGTrainingStats]):
     """Implementation of REINFORCE algorithm.
 
-    :param actor: mapping (s->model_output), should follow the rules in
-        :class:`~tianshou.policy.BasePolicy`.
+    :param actor: the actor network following the rules:
+        If `self.action_type == "discrete"`: (`s` ->`action_values_BA`).
```
MischaPanch:
```suggestion
        If `self.action_type == "discrete"`: (`s_B` ->`action_values_BA`).
```

tianshou/policy/modelfree/pg.py
```diff
@@ -40,8 +48,9 @@ class PGTrainingStats(TrainingStats):
...
+        If `self.action_type == "continuous"`: (`s` -> `dist_input_BD`).
```
MischaPanch:
```suggestion
        If `self.action_type == "continuous"`: (`s_B` -> `dist_input_BD`).
```

tianshou/policy/modelfree/pg.py
```diff
@@ -175,20 +184,21 @@ def forward(
...
-        # TODO: rename? It's not really logits and there are particular
-        #  assumptions about the order of the output and on distribution type
-        logits, hidden = self.actor(batch.obs, state=state, info=batch.info)
-        if isinstance(logits, tuple):
-            dist = self.dist_fn(*logits)
+        # TODO - ALGO: marked for algorithm refactoring
+        action_dist_input_BD, hidden_BH = self.actor(batch.obs, state=state, info=batch.info)
+        # in the case that self.action_type == "discrete", the dist should always be Categorical, and D=A
+        # therefore action_dist_input_BD is equivalent to logits_BA
+        if self.action_type == "discrete":
```
MischaPanch: This if-else seems weird and unnecessary, I'll look into it

tianshou/policy/modelfree/sac.py
```diff
@@ -173,24 +173,23 @@ def forward(  # type: ignore
         state: dict | Batch | np.ndarray | None = None,
         **kwargs: Any,
     ) -> DistLogProbBatchProtocol:
-        logits, hidden = self.actor(batch.obs, state=state, info=batch.info)
-        assert isinstance(logits, tuple)
-        dist = Independent(Normal(*logits), 1)
+        (loc_B, scale_B), hidden = self.actor(batch.obs, state=state, info=batch.info)
```
MischaPanch: hidden should follow naming convention (end in `_BH`)

tianshou/policy/modelfree/td3.py
```diff
@@ -29,7 +29,7 @@ class TD3Policy(DDPGPolicy[TTD3TrainingStats], Generic[TTD3TrainingStats]):  # t
     """Implementation of TD3, arXiv:1802.09477.
 
     :param actor: the actor network following the rules in
-        :class:`~tianshou.policy.BasePolicy`. (s -> logits)
+        :class:`~tianshou.policy.BasePolicy`. (s -> actions)
```
MischaPanch, line 32: Not sure it's true, the actor might be expected to output dist inputs. Gonna look into it

tianshou/utils/net/continuous.py
```diff
@@ -77,12 +77,16 @@ def forward(
         state: Any = None,
         info: dict[str, Any] | None = None,
     ) -> tuple[torch.Tensor, Any]:
-        """Mapping: obs -> logits -> action."""
-        if info is None:
-            info = {}
-        logits, hidden = self.preprocess(obs, state)
-        logits = self.max_action * torch.tanh(self.last(logits))
-        return logits, hidden
+        """Mapping: s -> action_values, hidden_state.
```
MischaPanch: Follow naming convention

MischaPanch (commented):
Thanks @arnaujc91 , it brings us an important step towards removing wrong names across the source code!

The issue is not fully complete at this point since all kinds of BatchProtocol still carry `logits`, but it can't be fully completed before introducing an `Algorithm` abstraction as part of release `2.0.0` anyway. So for the version `1.x,y` this is likely as far as we can push it.

I will address the comments myself and merge this

tianshou/policy/modelfree/trpo.py
```diff
@@ -25,7 +25,9 @@ class TRPOTrainingStats(NPGTrainingStats):
 class TRPOPolicy(NPGPolicy[TTRPOTrainingStats]):
     """Implementation of Trust Region Policy Optimization. arXiv:1502.05477.
 
-    :param actor: the actor network following the rules in BasePolicy. (s -> logits)
+    :param actor: the actor network following the rules:
+        If `self.action_type == "discrete"`: (`s` ->`action_values_BA`).
```
MischaPanch:
```suggestion
        If `self.action_type == "discrete"`: (`s_B` ->`action_values_BA`).
```

tianshou/policy/modelfree/trpo.py
```diff
@@ -25,7 +25,9 @@ class TRPOTrainingStats(NPGTrainingStats):
...
+        If `self.action_type == "continuous"`: (`s` -> `dist_input_BD`).
```
MischaPanch:
```suggestion
        If `self.action_type == "continuous"`: (`s_B` -> `dist_input_BD`).
```

tianshou/policy/modelfree/pg.py
MischaPanch: Resolving this, part of the naming can't be improved now because we don't have Algorithm yet

tianshou/policy/modelfree/sac.py
MischaPanch: Done

tianshou/policy/modelfree/pg.py
MischaPanch: Required a breaking change to fix, see d0d9d454b9. Also required a change in the (interiors of the) high-level interfaces

@opcode81 @maxhuettenrauch pls have a look at that commit. Max, I believe you had written a todo on how to improve the typing of the dist-fn - this commit addresses the todo

MischaPanch: I ran several low-level and high-level examples, things work. I think I haven't missed anything :)

MischaPanch (approved):

Issue #1015, opened 2023-12-28
How to successfully run a demo


zhiyu2020:
Hello, i have following problems.
First of all, i want to try demo to see the process of the whole infrastructure. But in the example part, it seems that the documentation is not clear. I don't know how to run demo successfully after installing tianshou with pip.
Please give detailed documentation about using example demo so that i can see how to run successfully.
![image](https://github.com/thu-ml/tianshou/assets/38719444/adca43dc-3170-44db-95a8-40f2711ed3dd)

Trinkle23897:
Please use source code installation, i.e. pip install -e .
We did a major refactor few months ago, and didn’t update the pypi wheel. cc @MischaPanch

MischaPanch:
FYI: Release of the wheel in next version is planned for late January or early February

MischaPanch:
If you clone the repo, the surest say to install is with poetry, e.g. using `poetry install`

dantp-ai:
My system is:
* macOS 13.6.3
* Python
  ```
  >>> import sys
  >>> print(sys.version, sys.platform)
  3.11.7 (main, Dec 27 2023, 17:48:28) [Clang 15.0.0 (clang-1500.1.0.2.5)] darwin
  ```

---

* I tried running the same demo from above, after having installed the dependencies with poetry like so: `poetry install --all-extras --with dev` (without the extras, `opencv-python` and `atari-py` are not being installed).

* For running this demo, I had to additionally install manually `pip install gymnasium[atari]` and `pip install gymnasium[accept-rom-license]`. I also had to manually install `pip install cmake`

* `envpool` fails to install:
  ```
  ERROR: Could not find a version that satisfies the requirement envpool (from versions: none)
  ERROR: No matching distribution found for envpool
  ```
  * This seems to be an issue with envpool reported also on its Issues board [here](https://github.com/sail-sg/envpool/issues/189).

* After all this, when running `python atari/atari_rainbow.py` the program seems to be stuck because I get intermittently the message that ALE is loaded
  ```
  A.L.E: Arcade Learning Environment (version 0.8.1+53f58b7)
  [Powered by Stella]
  ```

MischaPanch:
Thank you a lot for going through the installation and looking into this! I will check the atari example in particular and address the installation details, such that everything is installed at once with the extras.

Could you please also try mujoco or one of the other environments? Atari is a particularly complicated one, the problems might be due to gymnasium (on mac) and not due to tianshou

[...]

PR #1044, opened 2024-02-05
Update and fix dependencies related to mac install

maxhuettenrauch:
Addresses part of #1015 

### Dependencies

- move jsonargparse and docstring-parser to dependencies to run hl examples without dev
- create mujoco-py extra for legacy mujoco envs
- updated atari extra
    - removed atari-py and gym dependencies
    - added ALE-py, autorom, and shimmy
- created robotics extra for HER-DDPG

### Mac specific

- only install envpool when not on mac
- mujoco-py not working on macOS newer than Monterey (https://github.com/openai/mujoco-py/issues/777)
- D4RL also fails due to dependency on mujoco-py (https://github.com/Farama-Foundation/D4RL/issues/232)

### Other

- reduced training-num/test-num in example files to a number ≤ 20 (examples with 100 led to too many open files)
- rendering for Mujoco envs needs to be fixed on gymnasium side (https://github.com/Farama-Foundation/Gymnasium/issues/749)


README.md
```diff
@@ -96,6 +96,23 @@ You can also install the dev requirements by adding `--with dev` or the extras
...
+- `mujoco-py` (for legacy mujoco-py environments[^1])
+- `pybullet` (for pybullet environments)
+- `robotics` (for gymnasium-robotics environments)
+- `vizdoom` (for ViZDoom environments)
+- `envpool` (for envpool integration)
+- `dev` (for development requirements)
+
+[^1]: `mujoco-py` is a legacy package and is not recommended for new projects. 
+It is only included for compatibility with older projects. 
+Also note, that there may be compatibility issues with macOS newer than
```
MischaPanch:
```suggestion
Also note that there may be compatibility issues with macOS newer than 
```

README.md
```diff
@@ -96,6 +96,23 @@ You can also install the dev requirements by adding `--with dev` or the extras
...
+- `atari` (for Atari environments)
+- `box2d` (for Box2D environments)
+- `classic_control` (for classic control (discrete) environments)
+- `mujoco` (for MuJoCo environments)
+- `mujoco-py` (for legacy mujoco-py environments[^1])
+- `pybullet` (for pybullet environments)
+- `robotics` (for gymnasium-robotics environments)
+- `vizdoom` (for ViZDoom environments)
+- `envpool` (for envpool integration)
+- `dev` (for development requirements)
```
MischaPanch: Isn't that a group rather than an extra?

examples/box2d/bipedal_hardcore_sac.py
```diff
@@ -60,13 +60,22 @@ def __init__(self, env, action_repeat=3, reward_scale=5, rm_done=True):
     def step(self, action):
         rew_sum = 0.0
         for _ in range(self.action_repeat):
-            obs, rew, done, info = self.env.step(action)
+            step_result = self.env.step(action)
+            if len(step_result) == 4:
```
MischaPanch: This should never happen, we use gymnasium envs, not (older) gym envs. How did you run into this?

MischaPanch: Note that if it needs fixing in one example, it will need fixing in all of them :(

examples/mujoco/fetch_her_ddpg.py
```diff
@@ -30,7 +29,7 @@
 
 def get_args():
     parser = argparse.ArgumentParser()
-    parser.add_argument("--task", type=str, default="FetchReach-v3")
+    parser.add_argument("--task", type=str, default="FetchReach-v2")
```
MischaPanch, line 34: why this change?

examples/mujoco/fetch_her_ddpg.py
```diff
@@ -96,6 +95,8 @@ def test_ddpg(args=get_args()):
 
     # logger
     if args.logger == "wandb":
+        import wandb
```
MischaPanch: If you do it here, you need to do it everywhere. I thought wandb was not an optional dependency, am I wrong?

pyproject.toml
```diff
@@ -27,8 +27,10 @@ exclude = ["test/*", "examples/*", "docs/*"]
 
 [tool.poetry.dependencies]
 python = "^3.11"
+docstring-parser = "^0.15"
```
MischaPanch: This and jsonargparse should probably be within its own group. Maybe becoming optional later or even now, what do you think @opcode81 ?

Tianshou doesn't need these dependencies, but they are needed to run the examples. I guess it's a bit weird to have to install an extra in order to run examples. On the other side, we should avoid adding any unnecessary requirements to the core package, python package conflicts can be a real PITA

pyproject.toml
```diff
@@ -44,28 +46,43 @@ virtualenv = [
...
 
 [tool.poetry.extras]
-atari = ["atari_py", "opencv-python"]
-mujoco = ["mujoco_py"]
+atari = ["ale-py", "autorom", "opencv-python", "shimmy"]
+box2d = ["box2d-py", "pygame", "swig"]
+classic-control = ["pygame"]
+classic_control = ["pygame"]
+mujoco = ["mujoco", "imageio"]
+mujoco-py = ["mujoco-py", "cython"]
```
MischaPanch:
```suggestion
```

pyproject.toml
```diff
@@ -44,28 +46,43 @@ virtualenv = [
...
+shimmy = { version = ">=0.1.0,<1.0", optional = true }
+swig = { version = "4.*", optional = true }
+vizdoom = { version = "*", optional = true }
 
 [tool.poetry.extras]
-atari = ["atari_py", "opencv-python"]
-mujoco = ["mujoco_py"]
+atari = ["ale-py", "autorom", "opencv-python", "shimmy"]
+box2d = ["box2d-py", "pygame", "swig"]
+classic-control = ["pygame"]
```
MischaPanch: Is poetry not warning against duplicated entries?

```suggestion
```

pyproject.toml
```diff
@@ -44,28 +46,43 @@ virtualenv = [
...
 pybullet = { version = "*", optional = true }
+pygame = { version = ">=2.1.3", optional = true }
+shimmy = { version = ">=0.1.0,<1.0", optional = true }
+swig = { version = "4.*", optional = true }
+vizdoom = { version = "*", optional = true }
 
 [tool.poetry.extras]
-atari = ["atari_py", "opencv-python"]
-mujoco = ["mujoco_py"]
+atari = ["ale-py", "autorom", "opencv-python", "shimmy"]
```
MischaPanch, line 73: Please explain in a block comment why we have to do it like that (with a link to the relevant poetry issue).

This part requires attention and monitoring gymnasium for changes

README.md
```diff
@@ -96,6 +96,23 @@ You can also install the dev requirements by adding `--with dev` or the extras
...
+Available extras are:
+- `atari` (for Atari environments)
+- `box2d` (for Box2D environments)
+- `classic_control` (for classic control (discrete) environments)
+- `mujoco` (for MuJoCo environments)
+- `mujoco-py` (for legacy mujoco-py environments[^1])
+- `pybullet` (for pybullet environments)
+- `robotics` (for gymnasium-robotics environments)
+- `vizdoom` (for ViZDoom environments)
+- `envpool` (for envpool integration)
```
MischaPanch:
```suggestion
- `envpool` (for [envpool](https://github.com/sail-sg/envpool/)  integration)
```

MischaPanch (changes requested):

README.md
maxhuettenrauch: If you're referring to 'dev', then you're right. 'robotics' and 'vizdoom' should be extras (in the same fashion as 'atari' and 'mujoco' before). Should I just remove it from readme?

examples/box2d/bipedal_hardcore_sac.py
maxhuettenrauch: I guess it was some older example that noone touched in a while. The check for old vs new api is still present in atari_wrapper.py, that's where I just copied it from. You're right though, the Collector assumes terminated and truncated coming from env.step, I'll just delete this part

examples/mujoco/fetch_her_ddpg.py
maxhuettenrauch: There is no v3 in gymnasium robotics (https://robotics.farama.org/envs/fetch/)

examples/mujoco/fetch_her_ddpg.py
maxhuettenrauch: I checked, it's the only example where wandb is import. I also thought about removing it since wandb currently is only part of the 'dev' group dependencies

README.md
MischaPanch: Yes, just dev. Feel free to include it separately in the readme if not done yet, as groups are installed with a different command (not --extras)

examples/mujoco/fetch_her_ddpg.py
MischaPanch: Remove pls

pyproject.toml
maxhuettenrauch: If we want it to be pip-installable, it should be an extra (for example 'high-level-api'). A group would only be installable using poetry

MischaPanch: I guess an extra called `argparse` would make sense. You'd need to very visibly document that this extra is required to run the high-level-api examples

MischaPanch (approved):

MischaPanch:
Thank you, now only a final testing on windows and ubuntu is needed, and #1015 can be resolved. Merging this when the pipeline passed

PR #1046, opened 2024-02-07
Add option to collect same number of episodes in each collector env

bordeauxred:
- [x] If applicable, I have added tests to cover my changes.
- [ ] I have reformatted the code using `poe format` 
- [ ] I have checked style and types with `poe lint` and `poe type-check`

### DIFF
--- a/tianshou/data/buffer/base.py
+++ b/tianshou/data/buffer/base.py
@@ -50,6 +50,7 @@ class ReplayBuffer:
         "info",
         "policy",
     )
+    _required_keys = frozenset({"obs", "act", "rew", "terminated", "truncated", "done"})
 
     def __init__(
         self,
@@ -258,17 +259,15 @@ class ReplayBuffer:
             new_batch.__dict__[key] = batch[key]
         batch = new_batch
         batch.__dict__["done"] = np.logical_or(batch.terminated, batch.truncated)
-        if not {"obs", "act", "rew", "terminated", "truncated", "done"}.issubset(
+        if not self._required_keys.issubset(
             batch.keys(),
         ):  # important to do this after preprocessing the batch
-            missing_keys = {"obs", "act", "rew", "terminated", "truncated", "done"}.difference(
-                batch.keys(),
-            )
+            missing_keys = self._required_keys.difference(batch.keys())
             raise RuntimeError(
                 f"The input batch you try to add is missing the keys {missing_keys}.",
             )
         stacked_batch = buffer_ids is not None
-        if stacked_batch and not len(batch) == 1:
+        if stacked_batch and len(batch) != 1:
             raise RuntimeError(
                 f"len(batch) has to equal 1 when buffer_ids is not None (currently it is {buffer_ids}), but instead it is {len(batch)}.",
             )
--- a/tianshou/data/buffer/base.py
+++ b/tianshou/data/buffer/base.py
@@ -29,7 +29,7 @@ class ReplayBuffer:
         when using frame-stack sampling method.
     """
 
-    _reserved_keys = (
+    _RESERVED_KEYS = (
         "obs",
         "act",
         "rew",
@@ -40,7 +40,7 @@ class ReplayBuffer:
         "info",
         "policy",
     )
-    _input_keys = (
+    _INPUT_KEYS = (
         "obs",
         "act",
         "rew",
@@ -50,7 +50,7 @@ class ReplayBuffer:
         "info",
         "policy",
     )
-    _required_keys = frozenset({"obs", "act", "rew", "terminated", "truncated", "done"})
+    _REQUIRED_KEYS = frozenset({"obs", "act", "rew", "terminated", "truncated", "done"})
 
     def __init__(
         self,
@@ -104,7 +104,8 @@ class ReplayBuffer:
 
     def __setattr__(self, key: str, value: Any) -> None:
         """Set self.key = value."""
-        assert key not in self._reserved_keys, f"key '{key}' is reserved and cannot be assigned"
+        if key in self._RESERVED_KEYS:
+            raise ValueError(f"key '{key}' is reserved and cannot be assigned")
         super().__setattr__(key, value)
 
     def save_hdf5(self, path: str, compression: str | None = None) -> None:
@@ -162,7 +163,7 @@ class ReplayBuffer:
     def set_batch(self, batch: RolloutBatchProtocol) -> None:
         """Manually choose the batch you want the ReplayBuffer to manage."""
         assert len(batch) == self.maxsize and set(batch.keys()).issubset(
-            self._reserved_keys,
+            self._RESERVED_KEYS,
         ), "Input batch doesn't meet ReplayBuffer's data form requirement."
         self._meta = batch
 
@@ -259,10 +260,8 @@ class ReplayBuffer:
             new_batch.__dict__[key] = batch[key]
         batch = new_batch
         batch.__dict__["done"] = np.logical_or(batch.terminated, batch.truncated)
-        if not self._required_keys.issubset(
-            batch.keys(),
-        ):  # important to do this after preprocessing the batch
-            missing_keys = self._required_keys.difference(batch.keys())
+        # important to do this after preprocessing the batch
+        if missing_keys := self._REQUIRED_KEYS.difference(batch.keys()):
             raise RuntimeError(
                 f"The input batch you try to add is missing the keys {missing_keys}.",
             )
@@ -426,6 +425,6 @@ class ReplayBuffer:
             "policy": self.get(indices, "policy", Batch()),
         }
         for key in self._meta.__dict__:
-            if key not in self._input_keys:
+            if key not in self._INPUT_KEYS:
                 batch_dict[key] = self._meta[key][indices]
         return cast(RolloutBatchProtocol, Batch(batch_dict))
--- a/tianshou/data/buffer/base.py
+++ b/tianshou/data/buffer/base.py
@@ -21,13 +21,13 @@ class ReplayBuffer:
     :doc:`/01_tutorials/01_concepts`.
 
     :param size: the maximum size of replay buffer.
-    :param stack_num: the frame-stack sampling argument, should be greater than or
-        equal to 1. Default to 1 (no stacking).
-    :param ignore_obs_next: whether to not store obs_next. Default to False.
+    :param stack_num: the frame-stack sampling argument. It Should be greater than or
+        equal to 1.
+    :param ignore_obs_next: whether to not store obs_next.
     :param save_only_last_obs: only save the last obs/obs_next when it has a shape
-        of (timestep, ...) because of temporal stacking. Default to False.
+        of (timestep, ...) because of temporal stacking.
     :param sample_avail: the parameter indicating sampling only available index
-        when using frame-stack sampling method. Default to False.
+        when using frame-stack sampling method.
     """
 
     _reserved_keys = (
@@ -240,7 +240,7 @@ class ReplayBuffer:
         batch: RolloutBatchProtocol,
         buffer_ids: np.ndarray | list[int] | None = None,
     ) -> tuple[np.ndarray, np.ndarray, np.ndarray, np.ndarray]:
-        """Add a batch of data into replay buffer.
+        """Add a batch of data into the replay buffer.
 
         :param batch: the input data batch. "obs", "act", "rew",
             "terminated", "truncated" are required keys.
@@ -274,9 +274,11 @@ class ReplayBuffer:
             rew, done = batch.rew[0], batch.done[0]
         else:
             rew, done = batch.rew, batch.done
-        ptr, ep_rew, ep_len, ep_idx = (np.array([x]) for x in self._add_index(rew, done))
+        ep_last_idx, ep_rew, ep_len, ep_start_idx = (
+            np.array([x]) for x in self._add_index(rew, done)
+        )
         try:
-            self._meta[ptr] = batch
+            self._meta[ep_last_idx] = batch
         except ValueError:
             stack = not stacked_batch
             batch.rew = batch.rew.astype(float)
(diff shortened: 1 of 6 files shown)


bordeauxred:
@MischaPanch

tianshou/data/collector.py
```diff
@@ -237,6 +260,9 @@ def collect(
             True (no gradient retaining).
         :param gym_reset_kwargs: extra keyword arguments to pass into the environment's
             reset function. Defaults to None (extra keyword arguments)
+        :param sample_equal_from_each_env: whether to sample equal number of episodes
+            from each env. Otherwise it is only ensured that at least one episode
```
MischaPanch:
```suggestion
            from each env. Otherwise, it is only ensured that at least one episode
```

tianshou/data/collector.py
```diff
@@ -237,6 +260,9 @@ def collect(
...
+            is collected from every env when using n_episode. Default to False.
```
MischaPanch: Pls remove all "Default to ..." statements wherever you encounter them

MischaPanch:
```suggestion
            is collected from every env when using `n_episode != None`. Default to False.
```

tianshou/data/collector.py
```diff
@@ -257,9 +283,16 @@ def collect(
                     f"n_step={n_step} is not a multiple of #env ({self.env_num}), "
                     "which may cause extra transitions collected into the buffer.",
                 )
+            if sample_equal_from_each_env:
```
MischaPanch: Should we then ensure that roughly the same number of steps is collected from each env?

tianshou/data/collector.py
```diff
@@ -237,6 +260,9 @@ def collect(
             True (no gradient retaining).
         :param gym_reset_kwargs: extra keyword arguments to pass into the environment's
             reset function. Defaults to None (extra keyword arguments)
+        :param sample_equal_from_each_env: whether to sample equal number of episodes
```
MischaPanch: You should explain when one might want to consider this - e.g. when episodes are of varying lengths, or when different env instantiations are of different speeds

tianshou/data/collector.py
```diff
@@ -220,6 +242,7 @@ def collect(
         render: float | None = None,
         no_grad: bool = True,
         gym_reset_kwargs: dict[str, Any] | None = None,
+        sample_equal_from_each_env: bool = False,
```
MischaPanch: Rename to `collect_equally_from_each_env`? At least equal -> equally

tianshou/data/collector.py
```diff
@@ -257,9 +283,16 @@ def collect(
                     f"n_step={n_step} is not a multiple of #env ({self.env_num}), "
                     "which may cause extra transitions collected into the buffer.",
                 )
+            if sample_equal_from_each_env:
+                warnings.warn("sample_equal_from_each_env is ignored when using n_step.")
             ready_env_ids = np.arange(self.env_num)
         elif n_episode is not None:
             assert n_episode > 0
+            if sample_equal_from_each_env:
```
MischaPanch: I generally dislike asserts, let's use ValueErrors instead. There is an option to run python programs such that asserts are ignored, so they should rather be viewed as "optional error checking" (even though many projects don't adhere to this).

Also, pls include input information in log messages, it makes them much more useful. Should be something like
```python
f"{n_episode=} must be a "
f"multiple of the number of envs when sample_equal_from_each_env is True but got {self.env_num=}."
```
```

File: tianshou/data/collector.py [hunk 7]
```diff
@@ -356,21 +390,36 @@ def collect(
                 episode_lens.extend(ep_len[env_ind_local])
                 episode_returns.extend(ep_rew[env_ind_local])
                 episode_start_indices.extend(ep_idx[env_ind_local])
-                # now we copy obs_next to obs, but since there might be
-                # finished episodes, we have to reset finished envs first.
-                self._reset_env_with_ids(env_ind_local, env_ind_global, gym_reset_kwargs)
-                for i in env_ind_local:
-                    self._reset_state(i)
+                for idx, ret in zip(env_ind_global, ep_rew[env_ind_local], strict=True):
```
MischaPanch: I can't properly review this b/c the collect method is a huge conglomeration which is unintelligible to me :(. Let's try to refactor it and break it down into smaller chunks in a pair programming session

tianshou/data/collector.py
```diff
@@ -476,6 +526,9 @@ def collect(
 
         :return: A dataclass object
         """
+        assert (
```
MischaPanch, line 692: Why should we allow a param if it can't actually be changed?

MischaPanch (changes requested):

tianshou/data/buffer/base.py
```diff
@@ -257,28 +258,36 @@ def add(
             new_batch.__dict__[key] = batch[key]
         batch = new_batch
         batch.__dict__["done"] = np.logical_or(batch.terminated, batch.truncated)
-        assert {"obs", "act", "rew", "terminated", "truncated", "done"}.issubset(
+        if not {"obs", "act", "rew", "terminated", "truncated", "done"}.issubset(
```
MischaPanch: Better:

Put a classattribute

```python
REQUIRED_KEYS = {"obs", "act", "rew", "terminated", "truncated", "done"}
```

and then here 
```python
missing_keys = self.REQUIRED_KEYS.difference(batch.keys())
if missing_keys:
    raise ...
```

tianshou/data/buffer/base.py
```diff
@@ -257,28 +258,36 @@ def add(
...
+            missing_keys = {"obs", "act", "rew", "terminated", "truncated", "done"}.difference(
+                batch.keys(),
+            )
+            raise RuntimeError(
+                f"The input batch you try to add is missing the keys {missing_keys}.",
+            )
         stacked_batch = buffer_ids is not None
-        if stacked_batch:
-            assert len(batch) == 1
+        if stacked_batch and not len(batch) == 1:
```
MischaPanch:
```suggestion
        if stacked_batch and len(batch) != 1:
```

tianshou/data/buffer/manager.py
```diff
@@ -141,22 +142,24 @@ def add(
...
-            ep_last_idxs.append(ep_last_idx + self._offset[buffer_id])
+            ep_add_at_idxs.append(ep_add_at_idx + self._offset[buffer_id])
             ep_lens.append(ep_len)
             ep_rews.append(ep_rew)
             ep_start_idxs.append(ep_start_idx + self._offset[buffer_id])
-            self.last_index[buffer_id] = ep_last_idx + self._offset[buffer_id]
+            self.last_index[buffer_id] = ep_add_at_idx + self._offset[buffer_id]
             self._lengths[buffer_id] = len(self.buffers[buffer_id])
-        ep_last_idxs = np.array(ep_last_idxs)
+        ep_add_at_idxs = np.array(ep_add_at_idxs)
```
MischaPanch, line 164: what is the semantics of this variable? The docstring calls it the current index. If it's not the ep_last_idxs, what does it mean?

bordeauxred: The index at which the transition is added to the buffer. 
As there is ep_start_idx that indicates the first transition of the current episode, ep_last_idx should be the  index of the last transition in the episode. Whenever the current transition does not contain done, this is not the last index of the episode (as it continues) but the index at which to add the current transition.

MischaPanch: Marking for discussion in pair programming

tianshou/data/buffer/manager.py
```diff
@@ -118,20 +118,26 @@ def add(
...
         """
+        # todo heavy code duplication with ReplayBuffer in buffer/base.py
         # preprocess batch
         new_batch = Batch()
         for key in set(self._reserved_keys).intersection(batch.keys()):
             new_batch.__dict__[key] = batch[key]
         batch = new_batch
         batch.__dict__["done"] = np.logical_or(batch.terminated, batch.truncated)
-        assert {"obs", "act", "rew", "terminated", "truncated", "done"}.issubset(batch.keys())
+        if not self._required_keys.issubset(batch.keys()):
```
MischaPanch: computing the difference and checking if it's empty is both faster and arguably clearer to read. You can even use the newer python syntax 

```python
if not (missing_keys := self._required_keys.difference(batch.keys())):
   ...
```

Disclaimer: I never used it so far, so maybe my syntax is a bit off

MischaPanch: https://peps.python.org/pep-0572/

tianshou/data/collector.py
```diff
@@ -42,30 +58,56 @@ class CollectStats(CollectStatsBase):
...
+        episode_lens: np.ndarray,
+    ) -> Self:
+        """Instantiate from variables inside the scope of collector.collect()."""
+        return cls(
+            n_collected_episodes=episode_count,
+            n_collected_steps=step_count,
+            collect_time=collect_call_duration,
+            collect_speed=step_count / collect_call_duration,
+            returns=np.array(episode_returns),
+            returns_stat=SequenceSummaryStats.from_sequence(episode_returns)
```
MischaPanch, line 86: Let's move the declaration of this and lens_stats above the instantiation of the object, a bit easier to read

I'm also wondering whether this could be done in the CollectStatsBase itself in post_init, let's have a look together later

tianshou/data/collector.py
```diff
@@ -140,42 +196,36 @@ def reset(
...
     def reset_buffer(self, keep_statistics: bool = False) -> None:
         """Reset the data buffer."""
         self.buffer.reset(keep_statistics=keep_statistics)
 
-    def reset_env(self, gym_reset_kwargs: dict[str, Any] | None = None) -> None:
-        """Reset all of the environments."""
+    def reset_env(
+        self,
+        gym_reset_kwargs: dict[str, Any] | None = None,
+        set_obs_next_to_obs: bool = False,
```
MischaPanch: This is a non-trivial functionality, why is it needed and when would one want to use it? Pls add a docstring

tianshou/data/collector.py
```diff
@@ -212,6 +264,32 @@ def _reset_env_with_ids(
...
+        :param n_episode: number of episodes to take
+        """
+        if n_step is None and n_episode is None:
+            raise RuntimeError("This should not have happened!")
+        if n_step:
+            return current_n_step >= n_step
+        elif n_episode:
+            return current_n_episode >= n_episode
+        else:
+            raise RuntimeError
```
MischaPanch: Error message missing

tianshou/data/collector.py
```diff
@@ -220,198 +298,299 @@ def collect(
...
+                        env_ids_done,
+                        non_idle_env_ids_done,
+                        gym_reset_kwargs,
+                    )
+                    for i in env_ids_done:
+                        self._reset_state(i)
 
-            if (n_step and step_count >= n_step) or (n_episode and episode_count >= n_episode):
-                break
+                if n_episode:
```
MischaPanch, line 404: Marking this block for discussion:

- why gym_reset_kwargs used only in one place
- factor out to separate method?
- Avoid usage of np.where, instead use getitem on boolean array

tianshou/data/collector.py
```diff
@@ -220,198 +298,299 @@ def collect(
...
-            collect_speed=step_count / collect_time,
-            returns=np.array(episode_returns),
-            returns_stat=SequenceSummaryStats.from_sequence(episode_returns)
-            if len(episode_returns) > 0
-            else None,
-            lens=np.array(episode_lens, int),
-            lens_stat=SequenceSummaryStats.from_sequence(episode_lens)
-            if len(episode_lens) > 0
-            else None,
+    def sample_at_least_one_episode_per_worker_postprocessing_on_done_env(
```
MischaPanch, line 461: this doesn't seem to sample anything, rather only filter.

Marking for discussion

tianshou/data/collector.py
```diff
@@ -220,198 +298,299 @@ def collect(
...
+        """Remove surplus env id from ready_env_ids to record at least one episode per worker."""
+        surplus_env_num = len(ready_env_ids) - (n_episode - episode_count)
+        if surplus_env_num > 0:
+            mask = np.ones_like(ready_env_ids, dtype=bool)
+            mask[env_ind_local[:surplus_env_num]] = False
+            ready_env_ids = ready_env_ids[mask]
+            self.data = self.data[mask]
+        return ready_env_ids
+
+    def sample_equal_episodes_per_worker_postprocessing_on_done_env(
```
MischaPanch, line 477: same as above

tianshou/data/collector.py
```diff
@@ -220,198 +298,299 @@ def collect(
...
+        done = np.logical_or(terminated, truncated)
+        self.data.update(
+            obs_next=obs_next,
+            rew=rew,
+            terminated=terminated,
+            truncated=truncated,
+            done=done,
+            info=info,
+        )
+        if self.preprocess_fn:
```
MischaPanch, line 513: marking for discussion - when is this option used? Is it dangerous to update data twice?

tianshou/data/collector.py
```diff
@@ -220,198 +298,299 @@ def collect(
...
+        ready_env_ids: np.ndarray,
+    ) -> None:
+        """Compute the next action in the active workers and update the data accordingly."""
+        if random:
+            try:
+                act_sample = [self._action_space[i].sample() for i in ready_env_ids]
+            except TypeError:  # envpool's action space is not for per-env
+                act_sample = [self._action_space.sample() for _ in ready_env_ids]
+            act_sample = self.policy.map_action_inverse(act_sample)  # type: ignore
+            self.data.update(act=act_sample)
```
MischaPanch: From the method name I wouldn't expect it to update data. Generally, data is modified inplace all over the class, we should avoid it where possible

MischaPanch (commented):
Overall, a much cleaner and more readable structure than before, but there are still some issues to address. Let's talk later today

tianshou/data/buffer/manager.py
bordeauxred: nice stuff, thanks.

MischaPanch:
Closing this, @bordeauxred will make a new one after #1063 is merged

PR #1055, opened 2024-02-14
Improvements in High-Level API and Poe Tasks

### LINKED ISSUE #1054: Allow number of episodes per test step to be configured in high-level API
At present, this is hard-coded to the number of test environments, i.e. one episode per test environment, because this is how it was implemented in all examples, but obviously, it should be configurable.

I propose to add a configuration item with the total number of episodes to collect per episode. If it is not a multiple of the number of test environments, then we round up to the nearest multiple.

opcode81:
- [X] I have added documentation for my changes

* Add an option to SamplingConfig which allows to configure number of test episodes
* Make OptimizerFactory more flexible, adding method `create_optimizer_for_params`
* Fix AutoAlphaFactoryDefault using hard-coded Adam optimizer
* Fix mypy issues that were platform/installation-dependent
* Limit scope of nbqa, resolving issues with files generated by old versions of the build

Fixes #1054


tianshou/highlevel/optim.py
```diff
@@ -1,20 +1,30 @@
...
+    def create_optimizer(
+        self,
+        module: torch.nn.Module,
+        lr: float,
+    ) -> torch.optim.Optimizer:
+        return self.create_optimizer_for_params(module.parameters(), lr)
+
     @abstractmethod
-    def create_optimizer(self, module: torch.nn.Module, lr: float) -> torch.optim.Optimizer:
+    def create_optimizer_for_params(self, params: TParams, lr: float) -> torch.optim.Optimizer:
```
MischaPanch, line 27: Would it have a downside to have a non-abstract `create_optimizer` here with the previous signature that uses `create_optimizer_for_params`? Then you wouldn't need the downstream changes while still keeping flexibility.

Using the `module.params` is a good default that works for most cases

tianshou/highlevel/params/alpha.py
```diff
@@ -32,5 +32,5 @@ def create_auto_alpha(
     ) -> tuple[float, torch.Tensor, torch.optim.Optimizer]:
         target_entropy = float(-np.prod(envs.get_action_shape()))
         log_alpha = torch.zeros(1, requires_grad=True, device=device)
-        alpha_optim = torch.optim.Adam([log_alpha], lr=self.lr)
+        alpha_optim = optim_factory.create_optimizer_for_params([log_alpha], self.lr)
```
MischaPanch, line 35: I prefer using kwargs (i.e. `params=[log_alpha], lr=self.lr`), also safer against refactorings. Maybe force them in the new method?

You could also force all kwargs after `params`, that one kinda makes sense as positional arg

opcode81:
Recommend to merge without squashing.

tianshou/highlevel/params/alpha.py
opcode81: It is not reasonable to require kwargs for a two-arg method with two entirely different types.

tianshou/highlevel/optim.py
opcode81: I'm confused: `create_optimizer` still exists, is non-abstract, is implemented via `create_optimizer_for_params` and there are no downstream changes (in user code; only this module).

MischaPanch: Sorry, I glanced over the diff too fast and misread it. All good, merging

MischaPanch:
The windows CI job fails b/c you changed pyproject and something about caching fails. I'll delete the current github caches before merging to master, that should fix it

MischaPanch (approved):

PR #1061, opened 2024-02-23
Fix/add watch env with obs rms

maxhuettenrauch:

Currently, the high-level experiment does not set the obs_rms property for watching a policy learned on environments using observation normalization. This PR adds a "watch environment" to the high-level experiment and sets the obs_rms property based on the train_env.

### DIFF
--- a/examples/mujoco/mujoco_env.py
+++ b/examples/mujoco/mujoco_env.py
@@ -57,8 +57,8 @@ class MujocoEnvObsRmsPersistence(Persistence):
             obs_rms = pickle.load(f)
         world.envs.train_envs.set_obs_rms(obs_rms)
         world.envs.test_envs.set_obs_rms(obs_rms)
-        if world.watch_env is not None:
-            world.watch_env.set_obs_rms(obs_rms)
+        if world.envs.watch_env is not None:
+            world.envs.watch_env.set_obs_rms(obs_rms)
 
 
 class MujocoEnvFactory(EnvFactoryRegistered):
@@ -79,17 +79,23 @@ class MujocoEnvFactory(EnvFactoryRegistered):
         :return: the vectorized environments
         """
         env = super().create_venv(num_envs, mode)
+        # obs norm wrapper
         if self.obs_norm:
             env = VectorEnvNormObs(env, update_obs_rms=mode == EnvMode.TRAIN)
         return env
 
-    def create_envs(self, num_training_envs: int, num_test_envs: int) -> ContinuousEnvironments:
-        envs = super().create_envs(num_training_envs, num_test_envs)
+    def create_envs(
+        self,
+        num_training_envs: int,
+        num_test_envs: int,
+        create_watch_env: bool = False,
+    ) -> ContinuousEnvironments:
+        envs = super().create_envs(num_training_envs, num_test_envs, create_watch_env)
         assert isinstance(envs, ContinuousEnvironments)
 
-        # obs norm wrapper
         if self.obs_norm:
             envs.test_envs.set_obs_rms(envs.train_envs.get_obs_rms())
+            if envs.watch_env is not None:
+                envs.watch_env.set_obs_rms(envs.train_envs.get_obs_rms())
             envs.set_persistence(MujocoEnvObsRmsPersistence())
-
         return envs
--- a/examples/mujoco/mujoco_env.py
+++ b/examples/mujoco/mujoco_env.py
@@ -1,10 +1,11 @@
 import logging
 import pickle
 
-from tianshou.env import VectorEnvNormObs
+from tianshou.env import BaseVectorEnv, VectorEnvNormObs
 from tianshou.highlevel.env import (
     ContinuousEnvironments,
     EnvFactoryRegistered,
+    EnvMode,
     EnvPoolFactory,
     VectorEnvType,
 )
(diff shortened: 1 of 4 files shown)


examples/mujoco/mujoco_env.py
```diff
@@ -68,14 +71,24 @@ def __init__(self, task: str, seed: int, obs_norm=True) -> None:
...
+
     def create_envs(self, num_training_envs: int, num_test_envs: int) -> ContinuousEnvironments:
         envs = super().create_envs(num_training_envs, num_test_envs)
         assert isinstance(envs, ContinuousEnvironments)
 
         # obs norm wrapper
         if self.obs_norm:
-            envs.train_envs = VectorEnvNormObs(envs.train_envs)
-            envs.test_envs = VectorEnvNormObs(envs.test_envs, update_obs_rms=False)
             envs.test_envs.set_obs_rms(envs.train_envs.get_obs_rms())
```
opcode81, line 97: Oh, we actually need to do this for the watch env as well, because otherwise, we're only fixing it for the case you had initially, where the agent is being restored, but we also need to fix it for the regular training case, where the user wants to watch agent performance after training - and we need to make sure that the watch env then has the rms.

So what we need to do is to add the watch_env to `Environments` rather than to `World`, which also means that we will need to change the interface for `create_envs` (unless we just always create the watch env, no matter if the user will actually use it.)

examples/mujoco/mujoco_env.py
```diff
@@ -68,14 +71,24 @@ def __init__(self, task: str, seed: int, obs_norm=True) -> None:
...
+        env = super().create_venv(num_envs, mode)
+        if self.obs_norm:
+            env = VectorEnvNormObs(env, update_obs_rms=mode == EnvMode.TRAIN)
+        return env
+
     def create_envs(self, num_training_envs: int, num_test_envs: int) -> ContinuousEnvironments:
         envs = super().create_envs(num_training_envs, num_test_envs)
         assert isinstance(envs, ContinuousEnvironments)
 
         # obs norm wrapper
```
opcode81: This comment should be removed because we are no longer wrapping here.

examples/mujoco/mujoco_env.py
opcode81: @MischaPanch the idea behind this line is related to the small discussion we had recently about whether to handle this sort of thing in the env or the agent. The trouble is, if you associate learnt components with the env, you literally have to transfer all of them from the training envs to the test envs before every single test step - or, as is done here, make sure that the parameters are _inside a container_ that can be shared by all the envs. 

If they were associated with the agent, there would be far less that can go wrong...

opcode81: To make the change backwards compatible, I'd suggest that we change the interface by adding `create_watch_env: bool = False`.

maxhuettenrauch: Right, I missed the part where we would like to watch during train. I guess then there is no other way than adding the watch_env during `create_envs` (with everything what follows). Always adding it is not a good idea, I think. At least for me, this opens a mujoco viewer upon env creation.

MischaPanch: Yes, watching should be optional. I'm generally not happy about the current watch Dynamics (users shouldn't be forced or enticed to use tianshou abstractions to watch a trained policy), but that's a story for another PR. For now, let's just make it optional and easy to turn off from any interface

tianshou/highlevel/experiment.py
```diff
@@ -293,7 +297,7 @@ def run(
                 self._watch_agent(
                     self.config.watch_num_episodes,
                     policy,
-                    self.env_factory,
+                    watch_env,
```
MischaPanch: Let's use kwargs here pls

examples/mujoco/mujoco_env.py
opcode81: @MischaPanch, watching has always been optional, and users aren't forced to use Tianshou abstractions for it either; they just have an option, which, however, as a post-training stage, doesn't really make sense to use for learning processes that take more than a few minutes. But even for longer learning processes, users can still conveniently use the watch function by reloading the experiment as follows:

```
    e = Experiment.from_directory("wherever")
    e.config.watch = True
    e.config.train = False
    e.run()
```

This change is purely about when/where the watch env is created and making it receive the obs_rms from the train env; and it looks like we'll have to create it early if we are to gracefully handle the obs_rms sharing - a problem which could be avoided entirely if the normalisation was handled in the agent rather than the environment; that's why I mentioned you.

opcode81: In case it wasn't clear, I never suggested that we should create the watch_env without an option to control it/by default; I just mentioned it as a possibility that would not require changing the interface. I suggested the opposite (in the subsequent comment).

tianshou/highlevel/experiment.py
opcode81: Why? It's an internal function where all args have different types. And its only application uses clearly named variables for all args. Keyword arguments would only repeat the same names for no reason.

MischaPanch: Closing this thread, to some extent it's a personal preference. We can later discuss if we generally want something like a code-style guide (beyond what's already imposed by ruff)

examples/mujoco/mujoco_env.py
MischaPanch: I misunderstood the original reason you pinged me.

As mentioned in a private conversation, binding this kind of processing to the agent has mostly upsides, but also some downsides. Let's have a chat about it next week. It's not within the scope of this PR, right?

opcode81:
Please fix the one mypy issue. Otherwise approved.

MischaPanch (approved):

PR #1063, opened 2024-02-23
Feat/refactor collector

### LINKED ISSUE #1058: Remove data from state in Collector, and remove preprocess_fn there
The Collector tracks the `self.data` field that is a `RolloutBatch`. However, rather confusingly, this value of this field is recreated and updated at every call to `collect`. This leads to a lot of methods mutating the state, and a less comprehensible and debuggable code.

There is no need to bind this to the lifetime of the Collector, instead it should be created and used as a var in `collect`.

Moreover, there is the possibility to pass `preprocess_fn`, which
1. doesn't have a clear interfaces
2. doesn't have a clear purpose
3. Is *only used in a test* and nowhere else
It seems like the original purpose of that was to enable specific logging needs, but these needs can be met by different means.
4. Significally complicates the Collector

The `preprocess_fn` in `Collector` should be removed

This issue blocks #1046 since we need to reduce the complexity of the Collector before doing anything further with it

@bordeauxred is working on this issue

bordeauxred:
Closes: #1058 

Description (copied from newly established Changelog):

### Api Extensions
- Batch received two new methods: `to_dict` and `to_list_of_dicts`. #1063
- `Collector`s can now be closed, and their reset is more granular. #1063
- Trainers can control whether collectors should be reset prior to training. #1063
- Convenience constructor for `CollectStats` called `with_autogenerated_stats`. #1063

### Internal Improvements
- `Collector`s rely less on state, the few stateful things are stored explicitly instead of through a `.data` attribute. #1063
- Introduced a first iteration of a naming convention for vars in `Collector`s. #1063
- Generally improved readability of Collector code and associated tests (still quite some way to go). #1063
- Improved typing for `exploration_noise` and within Collector. #1063

### Breaking Changes

- Removed `.data` attribute from `Collector` and its child classes. #1063
- Collectors no longer reset the environment on initialization. Instead, the user might have to call `reset` 
expicitly or pass `reset_before_collect=True` . #1063
- VectorEnvs now return an array of info-dicts on reset instead of a list. #1063
- Fixed `iter(Batch(...)` which now behaves the same way as `Batch(...).__iter__()`. Can be considered a bugfix. #1063


bordeauxred:
@MischaPanch

tianshou/data/collector.py
```diff
@@ -177,43 +172,44 @@ def reset_buffer(self, keep_statistics: bool = False) -> None:
...
-        self.data.info = info  # type: ignore
-        self.data.obs = obs
-
-    def _reset_state(self, id: int | list[int]) -> None:
-        """Reset the hidden state: self.data.state[id]."""
-        if hasattr(self.data.policy, "hidden_state"):
-            state = self.data.policy.hidden_state  # it is a reference
+        return obs, info
+
+    def _reset_state(
```
MischaPanch: this function should be called _reset_hidden_state and be defined outside of the collector (note that the IDE should have warned you that the method can be static). The input should just be called `rollout_batch`, there is no need for the first `if hasattr` (that should happen in downstream code), and the return should be `None`.

Note that this modifies the batch inplace (side effect), and such operations should usually not return anything.

The parameter `id` deserves an explanation

tianshou/data/collector.py
```diff
@@ -177,43 +172,44 @@ def reset_buffer(self, keep_statistics: bool = False) -> None:
...
+        return cur_rollout_batch
 
     def _reset_env_with_ids(
         self,
+        cur_rollout_batch: RolloutBatchProtocol,
         local_ids: list[int] | np.ndarray,
         global_ids: list[int] | np.ndarray,
         gym_reset_kwargs: dict[str, Any] | None = None,
-    ) -> None:
+    ) -> RolloutBatchProtocol:
```
MischaPanch: Rename `cur_rollout_batch` to `rollout_batch`

The method modifies the batch, it shouldn't return it. The name should also be adjusted, I'm generally not sure the method should exist (is it used in many places?)

tianshou/data/collector.py
```diff
@@ -53,6 +52,28 @@ class CollectStats(CollectStatsBase):
     """Stats of the collected episode lengths."""
 
 
+def get_fresh_collect_batch_with_current_obs(
+    obs: np.ndarray,
+    info: dict | list[dict],
+) -> RolloutBatchProtocol:
+    """Empty batch with obs and info of current state, useful for adding new data."""
+    result = Batch(
+        obs={},
```
MischaPanch: `obs=obs` instead of mutating below

rename to `get_empty_rollout_batch`, turn `obs` and `info` params to optional.

Pls adjust the docstring to explain why one might want an empty batch with only obs and info (b/c that's what you get out of reset before making any steps, I guess)

tianshou/data/collector.py
```diff
@@ -145,26 +154,12 @@ def reset(
...
-            terminated={},
-            truncated={},
-            done={},
-            obs_next={},
-            info={},
-            policy={},
-        )
-        self.data = cast(RolloutBatchProtocol, data)
-        self.reset_env(gym_reset_kwargs)
+        self._last_obs, self._last_info = self.reset_env(gym_reset_kwargs)
```
MischaPanch: rename to `_reset_obs` and `_reset_info`?

Do we generally need to reset in the constructor at all? Could we reset in `.collect`?

MischaPanch (changes requested):
A first review, I haven't looked at the implementation of collect in detail here yet.

I have the feeling more reset methods can disappear from the collector if we just reset the env during `.collect` (which can be done optionally, I guess)

MischaPanch:
Failing pipeline needs attention before the next review

tianshou/data/collector.py
bordeauxred: _reset_obs could be easily taken as the obs to which to reset the env to. 
How about `_obs_from_reset` and `_info_from_reset`?

For the n_episode case, resetting in the beginning of `collect` is just fine. 
In the n_episode case, resetting is only happening the first time `collect` is called (which could be checked i.e. by checking for empty buffer). Do you prefer this?

MischaPanch: Empty buffer is not a reliable check, the user could have provided a filled buffer to the collector.

Instead, we could extend the signature of collect to contain `reset_env: bool = True`. The user passes the env and should be able to control whether it should be reset at the beginning of collect.

Note that the collector currently resets the env when it is instantiated, which is a different functionality. We should check for uses of the collector to see that we don't break anything when the resetting logic is moved to `.collect`.

After this has been done, the collector becomes essentially stateless, which will be a significant improvement.

tianshou/policy/modelfree/dqn.py
```diff
@@ -232,11 +233,27 @@ def learn(self, batch: RolloutBatchProtocol, *args: Any, **kwargs: Any) -> TDQNT
...
         self,
-        act: np.ndarray | BatchProtocol,
-        batch: RolloutBatchProtocol,
-    ) -> np.ndarray | BatchProtocol:
+        act: np.ndarray,
+        batch: ObsBatchProtocol,
+    ) -> np.ndarray:
+        pass
+
+    @overload
```
MischaPanch: @bordeauxred This can be done with generics instead of expressing it with overload, could you adjust it in all relevant places pls?

MischaPanch (approved):
Just one minor comment left

MischaPanch:
@Trinkle23897 this is ready for review now, ~~just the minor comment about replacing of overload by generics is left~~

@bordeauxred and I worked on this together, so it's reviewed from my side. There are only very minor breaking changes, see PR description

Overall, I think it's a good step towards a more readable and transparent Collector.

We have encountered many issues while working on that, so there are many new TODOs. We'll address them later on, and I'll add a new notebook documenting "gotchas" in Batch and Buffer in a separate PR.

I think a quick review from your side should be enough, understanding all details of the changes would be time consuming. Long story short: collector implementations rely less on state, are better typed and are more readable now. Hacky batch things have been marked as such explicitly

tianshou/policy/modelfree/dqn.py
MischaPanch: Thanks, resolving! You made the types as class args, not sure if it's a problem. If it is, I might push a minor change. Maybe it does make sense to have a central type instead of repeating the generic, I'll think about it and might adjust

MischaPanch:
I'll merge this now since it should be uncontroversial and it is blocking further development like #1077 . @Trinkle23897 if you have any comments, we can address them post-merge

Trinkle23897:
Thanks for doing that!

PR #1067, opened 2024-03-02
Add publish workflow, first preparation for next release

MischaPanch:
@Trinkle23897 This PR requires your attention! 

I think time has come to make the next release. Since there are rather massive changes/extensions, I believe we should make a bump in the major version, i.e., go to 1.0.0

It doesn't mean everything is stable, of course, but we can go to 2.0.0 with more stable low-level interfaces later.

I have added a publish workflow which should be triggered by a github release. We can discuss the release notes together, if you want. For it to work, you'll need to set the PYPI_TOKEN var in the repositories secrets.

I also made adjustments in the readme and in the docs about contributors, pls have a look whether you agree (especially with the latter).

In the next week we'll add more docs, more adjustments in the readme, and shift around some tutorials to make everything more cohesive for the next release.

You mentioned that you had some action-based release mechanism in mind already. I don't mind replacing the current publishing workflow with something else, if you prefer

@opcode81 @maxhuettenrauch @carlocagnetta @bordeauxred FYI


Trinkle23897:
Could you provide your PYPI account, so that I can add you to maintainer?

Added:
Screenshot 2024-03-03 at 9 01 56 AM

Trinkle23897 (dismissed):

PR #1072, opened 2024-03-06
Allow explicit setting of multiprocessing context for SubprocEnvWorker

maxhuettenrauch:

Running multiple training runs in parallel (with, for example, joblib) fails on macOS due to a change in the standard context for multiprocessing (see [here](https://stackoverflow.com/questions/65098398/why-using-fork-works-but-using-spawn-fails-in-python3-8-multiprocessing) or [here](https://www.reddit.com/r/learnpython/comments/g5372v/multiprocessing_with_fork_on_macos/)). This PR adds the ability to explicitly set a multiprocessing context for the SubProcEnvWorker (similar to gymnasium's [AsyncVecEnv](https://github.com/Farama-Foundation/Gymnasium/blob/main/gymnasium/vector/async_vector_env.py)). ~~It's not very nice, though, since there are common kwargs for BaseVectorEnv but different workers may have different kwargs (SubprocEnvWorker especially). I guess that's also why there exists an explicit ShmemVectorEnv (with the the only difference being share_memory=True compared to SubprocVectorEnv ).~~


examples/mujoco/mujoco_env.py
```diff
@@ -62,11 +62,19 @@ def restore(self, event: RestoreEvent, world: World):
 
 
 class MujocoEnvFactory(EnvFactoryRegistered):
-    def __init__(self, task: str, seed: int, obs_norm=True) -> None:
+    def __init__(
+        self,
+        task: str,
+        seed: int,
+        obs_norm=True,
```
MischaPanch:
```suggestion
        obs_norm: bool = True,
```

tianshou/env/venvs.py
```diff
@@ -383,9 +383,15 @@ class SubprocVectorEnv(BaseVectorEnv):
...
     """
 
-    def __init__(self, env_fns: Sequence[Callable[[], ENV_TYPE]], **kwargs: Any) -> None:
+    def __init__(
+        self,
+        env_fns: Sequence[Callable[[], ENV_TYPE]],
+        share_memory: bool = False,
+        context: Literal["fork", "spawn"] | None = None,
+        **kwargs: Any,
+    ) -> None:
```
MischaPanch, line 409: Pls add a docstring

tianshou/env/worker/subproc.py
```diff
@@ -33,8 +34,8 @@
 class ShArray:
     """Wrapper of multiprocessing Array."""
 
-    def __init__(self, dtype: np.generic, shape: tuple[int]) -> None:
-        self.arr = Array(_NP_TO_CT[dtype.type], int(np.prod(shape)))  # type: ignore
+    def __init__(self, dtype: np.generic, shape: tuple[int], ctx: BaseContext) -> None:
```
MischaPanch: that's a breaking change, is it ok to do that? Can it be done in a backwards compatible way?

tianshou/highlevel/env.py
```diff
@@ -72,14 +71,18 @@ class VectorEnvType(Enum):
     RAY = "ray"
     """Parallelization based on the `ray` library"""
 
-    def create_venv(self, factories: Sequence[Callable[[], gym.Env]]) -> BaseVectorEnv:
+    def create_venv(
+        self,
+        factories: Sequence[Callable[[], gym.Env]],
+        venv_kwargs: dict,
```
MischaPanch: this is also breaking, default to None?

tianshou/highlevel/env.py
```diff
@@ -72,14 +71,18 @@ class VectorEnvType(Enum):
...
+    ) -> BaseVectorEnv:
         match self:
             case VectorEnvType.DUMMY:
                 return DummyVectorEnv(factories)
```
MischaPanch, line 83: this should also take the kwargs

tianshou/highlevel/env.py
MischaPanch: Generally, I'm not a huge fan of passing kwargs - one could also solve this with VenvFactories. It would be somehow more in line with the rest of the code, although for this particular case it may also be an overkill.

@opcode81 what do you think?

tianshou/highlevel/env.py
```diff
@@ -265,6 +279,7 @@ def __init__(
     def from_factory(
         factory_fn: Callable[[EnvMode], gym.Env],
         venv_type: VectorEnvType,
+        venv_kwargs: dict,
```
MischaPanch: breaking

MischaPanch (changes requested):
Preliminary review, mainly I'm unsure whether we should be passing around venv_kwargs or use a different mechanism. @opcode81 pls also have a look, I tagged you in a comment.

There are also many breaking changes in interfaces which are probably all avoidable, but we should agree on the general kwargs issue first

tianshou/env/worker/subproc.py
maxhuettenrauch: It's only used within SubprocEnvWorker and by the time an object is created, the mp context has been created, so it might be ok?

tianshou/highlevel/env.py
maxhuettenrauch: I made all the kwargs in venvs explicit with default values. At least it shouldn't be breaking anymore?

opcode81: Not a fan of passing on kwargs either.
Which kwargs might a user reasonably pass here? Can you provide some examples how a user might user this? 
As far as I can tell, the only argument a user could pass when using the subproc env type would be the newly introduced `context`. Under which circumstances would a user set it and why?

Maybe we can directly map all the reasonable options to a separate VectorEnvType enum item.

tianshou/env/venvs.py
```diff
@@ -383,11 +388,27 @@ class SubprocVectorEnv(BaseVectorEnv):
...
+    def __init__(
+        self,
+        env_fns: Sequence[Callable[[], ENV_TYPE]],
+        wait_num: int | None = None,
+        timeout: float | None = None,
+        share_memory: bool = False,
+        context: Literal["fork", "spawn"] | None = None,
+    ) -> None:
+        """:param share_memory: whether to share memory between the main process and the worker process.
+        :param context: the context to use for multiprocessing. It can be "fork" or "spawn".
```
opcode81: Mentioning the possible options is redundant given the type information. What we would need instead is guidance when to use which option.

tianshou/highlevel/env.py
maxhuettenrauch: Correct, it's mainly about `context`. It's only special cases such as the aforementioned parallelization under MacOS where you would explicitly like to set a context as the default `None` simply uses the system default. A `SUBPROC_SHARED_MEM_FORK_CONTEXT` could do the trick, not exposing theses options to the user

tianshou/env/venvs.py
maxhuettenrauch: Tried to improve it

tianshou/highlevel/env.py
maxhuettenrauch: removed venv_kwargs

examples/mujoco/mujoco_ppo_hl.py
```diff
@@ -61,7 +62,12 @@ def main(
         repeat_per_collect=repeat_per_collect,
     )
 
-    env_factory = MujocoEnvFactory(task, experiment_config.seed, obs_norm=True)
+    env_factory = MujocoEnvFactory(
+        task,
+        experiment_config.seed,
+        obs_norm=True,
+        venv_type=VectorEnvType.SUBPROC_SHARED_MEM_FORK_CONTEXT,
```
MischaPanch: Why not using the default value here?

tianshou/env/venvs.py
```diff
@@ -371,8 +371,13 @@ class DummyVectorEnv(BaseVectorEnv):
...
 
-    def __init__(self, env_fns: Sequence[Callable[[], ENV_TYPE]], **kwargs: Any) -> None:
-        super().__init__(env_fns, DummyEnvWorker, **kwargs)
+    def __init__(
+        self,
+        env_fns: Sequence[Callable[[], ENV_TYPE]],
+        wait_num: int | None = None,
+        timeout: float | None = None,
+    ) -> None:
+        super().__init__(env_fns, DummyEnvWorker, wait_num, timeout)
```
MischaPanch, line 380: this is breaking. Have you made sure that **kwargs was never used? Pls edit the description of the PR to contain all breaking changes 

Pls also mention that users should refer to the docstring of the base class where appropriate

tianshou/env/venvs.py
```diff
@@ -383,11 +388,29 @@ class SubprocVectorEnv(BaseVectorEnv):
...
+        env_fns: Sequence[Callable[[], ENV_TYPE]],
+        wait_num: int | None = None,
+        timeout: float | None = None,
+        share_memory: bool = False,
+        context: Literal["fork", "spawn"] | None = None,
+    ) -> None:
+        """:param share_memory: whether to share memory between the main process and the worker process. Allows for
+        shared buffers to exchange observations
+        :param context: the context to use for multiprocessing. Consider using "fork" when using macOS and additional
+         parallelization through, for example, joblib. Defaults to None, which will use the default system context.
```
MischaPanch: not indented correctly (here and above), probably that's why the CI pipeline fails

tianshou/env/venvs.py
```diff
@@ -400,11 +423,16 @@ class ShmemVectorEnv(BaseVectorEnv):
...
+        self,
+        env_fns: Sequence[Callable[[], ENV_TYPE]],
+        wait_num: int | None = None,
+        timeout: float | None = None,
+    ) -> None:
         def worker_fn(fn: Callable[[], gym.Env]) -> SubprocEnvWorker:
             return SubprocEnvWorker(fn, share_memory=True)
 
-        super().__init__(env_fns, worker_fn, **kwargs)
+        super().__init__(env_fns, worker_fn, wait_num, timeout)
```
MischaPanch, line 440: I think kwargs should always be used as such and not as positional args, otherwise refactorings may fail, I believe

tianshou/env/worker/subproc.py
MischaPanch: Ok, I guess it's fine then. Pls add a docstring. In particular, it should be clear how the class is supposed to be used, and how the context is being created. Mentioning an example in the code is enough for that.

The class needs to remain public b/c instances of it can be returned, it seems. But the user should be informed when they might be interested in it (if ever at all)

tianshou/env/worker/subproc.py
```diff
@@ -125,23 +126,29 @@ def _encode_obs(
 class SubprocEnvWorker(EnvWorker):
     """Subprocess worker used in SubprocVectorEnv and ShmemVectorEnv."""
 
-    def __init__(self, env_fn: Callable[[], gym.Env], share_memory: bool = False) -> None:
+    def __init__(
+        self,
+        env_fn: Callable[[], gym.Env],
+        share_memory: bool = False,
+        context: Literal["fork", "spawn"] | None = None,
+    ) -> None:
```
MischaPanch, line 150: Could you add a docstring explaining the semantics of fork, spawn and None? When would I want to use which? Would it make sense to also allow passing a context instance?

tianshou/highlevel/env.py
```diff
@@ -69,17 +68,24 @@ class VectorEnvType(Enum):
     """Parallelization based on `subprocess`"""
     SUBPROC_SHARED_MEM = "shmem"
     """Parallelization based on `subprocess` with shared memory"""
+    SUBPROC_SHARED_MEM_FORK_CONTEXT = "shmem_fork"
+    """Parallelization based on `subprocess` with shared memory and fork context (relevant for macOS)"""
```
MischaPanch: If you have a link to some docs, might be good to add it :). Or is the info trivial to google?

tianshou/highlevel/env.py
```diff
@@ -343,7 +353,10 @@ def create_venv(
 class EnvFactory(ToStringMixin, ABC):
     """Main interface for the creation of environments (in various forms)."""
 
-    def __init__(self, venv_type: VectorEnvType):
+    def __init__(
+        self,
+        venv_type: VectorEnvType,
```
MischaPanch: I guess a comma got in accidentally and poe-format did the rest ^^

tianshou/highlevel/env.py
```diff
@@ -358,7 +371,12 @@ def create_venv(self, num_envs: int, mode: EnvMode) -> BaseVectorEnv:
         :param mode: the mode for which to create
```
MischaPanch: pls add that the user provided venv_type will be ignored for watch mode

MischaPanch (changes requested):
Close to completion, thanks for removing the venv_kwargs!

After comments are addressed and pipeline is green we can merge this

examples/mujoco/mujoco_ppo_hl.py
maxhuettenrauch: Oops, that was only for testing all types. Will remove

tianshou/env/venvs.py
maxhuettenrauch: Not sure I understand how this is breaking. The kwargs `wait_num` and `timeout` were already there and take on the same default values as in the base class. How could a `DummyVectorEnv` object be created that is not working anymore? By passing a dict with `wait_num` and `timeout` keys?

MischaPanch: It's breaking because it's a public interface, and previously the user could have passed any kwargs without an error being raised. Or am I wrong about this? I haven't checked how the kwargs were previous used

maxhuettenrauch: Any other kwarg would throw an error because `BaseVectorEnv` only takes 4 arguments
https://github.com/thu-ml/tianshou/blob/1714c7f2c7295e58c9b94f6dcdc270da23f3d74c/tianshou/env/venvs.py#L76-L82

MischaPanch: Two bads cancelling each other ^^

Since we are basing off python 3.11, we could use TypedDict together with **kwargs in the future to prevent such code. For now let's keep it as is, resolving

tianshou/env/worker/subproc.py
maxhuettenrauch: I think it's fine to even pass `None` and create the default context in that case. Then it's a non breaking change

tianshou/env/worker/subproc.py
maxhuettenrauch: In theory, one could also pass a context instance. Although I can't imagine a scenario where I would. Ultimately, it's a parameter that hardly ever needs to be touched, except something is not working :D

maxhuettenrauch: Just to give an example
https://github.com/google-deepmind/mujoco/issues/742

tianshou/highlevel/env.py
MischaPanch: kwargs are gone and backwards compat restored, resolving

tianshou/env/worker/subproc.py
MischaPanch: sounds good!

tianshou/highlevel/env.py
maxhuettenrauch: The basics can be read about in the python docs. The implications, however, can surface in the weirdest ways... (like in the example above from mujoco)

maxhuettenrauch: Also https://github.com/Farama-Foundation/Gymnasium/issues/222

tianshou/env/worker/subproc.py
MischaPanch: Then I would mention exactly that :)

tianshou/highlevel/env.py
MischaPanch: Then worth mentioning. It's nice if we can save a user (or future us) from going down a longish web-search to understand when and why these things would need to be adjusted

tianshou/env/worker/subproc.py
```diff
@@ -125,23 +142,32 @@ def _encode_obs(
...
+        env_fn: Callable[[], gym.Env],
+        share_memory: bool = False,
+        context: BaseContext | Literal["fork", "spawn"] | None = None,
+    ) -> None:
         self.parent_remote, self.child_remote = Pipe()
         self.share_memory = share_memory
         self.buffer: dict | tuple | ShArray | None = None
+        if not isinstance(context, BaseContext):
+            context = multiprocessing.get_context(context)
+        else:
```
MischaPanch: I don't think this is necessary or helpful. Just because Process attr exists, doesn't mean that the object is the right one.

The type annotation also excludes this case. I'm gonna remove it

tianshou/highlevel/env.py
MischaPanch:
```suggestion
        :param mode: the mode for which to create. In `WATCH` mode the resulting venv will always be of type `DUMMY` with a single env.
    
```

tianshou/env/venvs.py
```diff
@@ -381,13 +386,36 @@ class SubprocVectorEnv(BaseVectorEnv):
...
+
+        Additional arguments are:
+
+        :param share_memory: whether to share memory between the main process and the worker process. Allows for
+            shared buffers to exchange observations
+        :param context: the context to use for multiprocessing. Usually it's fine to use the default context, but
+            `spawn` as well as `fork` can have non-obvious side effects, see for example
+            https://github.com/google-deepmind/mujoco/issues/742, or
+            https://github.com/Farama-Foundation/Gymnasium/issues/222.
+            Consider using 'fork' when using macOS and additional parallelization through, for example, joblib.
```
MischaPanch:
```suggestion
            Consider using 'fork' when using macOS and additional parallelization though, for example, joblib.
```

MischaPanch (approved):

tianshou/env/venvs.py
MischaPanch: nvm, not worth the commit

tianshou/env/worker/subproc.py
maxhuettenrauch: This one was odd. The `BaseContext` is the base class for all other contexts, such as `SpawnContext` or `ForkContext`, but it is missing the `Process` attribute, therefore mypy complained...

maxhuettenrauch: Added an assert that checks for type and attribute presence

MischaPanch: Cool, thanks! We can leave it as is, but for future I would have preferred an assert without the `else`, and a  block comment that this is for mypy to be happy

MischaPanch: Looking at it again - it's a bit confusing. This `assert isinstance(context, BaseContext)` is always true in the `else` part.

Could you adjust to add

```python
assert hasattr(context, "Process")  # for mypy
```

without an `else` pls? Then this is ready to merge

@maxhuettenrauch

maxhuettenrauch: Done

PR #1074, opened 2024-03-12
Feature/algo eval

maxhuettenrauch:
# Changes

## Dependencies

- New extra "eval"

## Api Extension
- `Experiment` and  `ExperimentConfig` now have a `name`, that can however be overridden when `Experiment.run()` is called
- When building an `Experiment` from an `ExperimentConfig`, the user has the option to add info about seeds to the name.
- New method in `ExperimentConfig` called `build_default_seeded_experiments`
- `SamplingConfig` has an explicit training seed, `test_seed` is inferred.
- New `evaluation` package for repeating the same experiment with multiple seeds and aggregating the results (important extension!). Currently in alpha state.
- Loggers can now restore the logged data into python by using the new `restore_logged_data`

## Breaking Changes
- `AtariEnvFactory` (in examples) now receives explicit train and test seeds
- `EnvFactoryRegistered` now requires an explicit `test_seed`
- `BaseLogger.prepare_dict_for_logging` is now abstract


maxhuettenrauch:
@MischaPanch @bordeauxred please have a look

examples/mujoco/mujoco_ppo_hl_multi.py
```diff
@@ -0,0 +1,117 @@
...
+                else None,
+                dist_fn=DistributionFunctionFactoryIndependentGaussians(),
+            ),
+        )
+        .with_actor_factory_default(hidden_sizes, torch.nn.Tanh, continuous_unbounded=True)
+        .with_critic_factory_default(hidden_sizes, torch.nn.Tanh)
+        .build_default_seeded_experiments(num_experiments)
+    )
+
+    for experiment_name, experiment in experiments.items():
```
MischaPanch: Let's add a TODO here to parallelize this asap. In fact, already here something like a hydra launcher could be useful, but maybe we do that in a separate PR later

examples/mujoco/mujoco_ppo_hl_multi.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+def eval_experiments(log_dir: str):
+    results = RLiableExperimentResult.load_from_disk(log_dir, "PPO", None)
+    eval_results(results)
+
+
+if __name__ == "__main__":
+    # logging.run_cli(main)
+    # experiment_config = ExperimentConfig(watch=False)
```
MischaPanch: main should be executed, right?

examples/mujoco/tools.py
```diff
@@ -5,11 +5,111 @@
...
+@dataclass
+class RLiableExperimentResult:
+    exp_dir: str
+    algorithms: list[str]
+    score_dict: dict[str, np.ndarray]  # (n_runs x n_epochs + 1)
+    env_steps: np.ndarray  # (n_epochs + 1)
+    score_thresholds: np.ndarray
+
+    @staticmethod
+    def load_from_disk(exp_dir: str, algo_name: str, score_thresholds: np.ndarray | None):
```
MischaPanch: I feel like loading should only require one path parameter. Is it possible to have default behavior without passing algo_name and trhesholds? Should the latter be None by default?

examples/mujoco/tools.py
```diff
@@ -5,11 +5,111 @@
...
 
+from tianshou.highlevel.experiment import Experiment
+
+
+@dataclass
+class RLiableExperimentResult:
+    exp_dir: str
+    algorithms: list[str]
+    score_dict: dict[str, np.ndarray]  # (n_runs x n_epochs + 1)
+    env_steps: np.ndarray  # (n_epochs + 1)
```
MischaPanch: in tianshou 1 epoch has many steps, is it not true in this context?

examples/mujoco/tools.py
```diff
@@ -5,11 +5,111 @@
...
+            score_thresholds=score_thresholds,
+            exp_dir=exp_dir,
+        )
+
+
+def eval_results(results: RLiableExperimentResult):
+    import matplotlib.pyplot as plt
+    import scipy.stats as sst
+    import seaborn as sns
+    from rliable import library as rly
```
MischaPanch: rliable should become an extra dependency in the pyproject. On import error we should tell the user to install with `-E rliable`

As to the other dependencies - are they not part of tianshou yet? Do we need the local import? If they aren't, the extra should be something like `eval` and contain all these dependencies

MischaPanch: Alternatively, if theyre not too big, we could just add them to the direct dependencies. Up to rliable, they're very standard. If rliable itself is not heavyweight, we can just add it, I think

examples/mujoco/tools.py
```diff
@@ -5,11 +5,111 @@
...
+    from rliable import plot_utils
+
+    iqm = lambda scores: sst.trim_mean(scores, proportiontocut=0.25, axis=0)
```
MischaPanch: Why the hardcoded 0.25 and reps=50000. Would the user never be interested in adjusting them?

examples/mujoco/tools.py
```diff
@@ -5,11 +5,111 @@
...
+    plot_utils.plot_sample_efficiency_curve(
+        results.env_steps,
+        iqm_scores,
+        iqm_cis,
+        algorithms=results.algorithms,
+        xlabel=r"Number of env steps",
+        ylabel="IQM episode return",
+        ax=ax,
+    )
+    plt.savefig(os.path.join(results.exp_dir, "iqm_sample_efficiency_curve.png"))
```
MischaPanch: I feel like the path should be configurable with a default value.

The function needs a docstring explaining that images will be saved to disk, I wouldn't have expected it. Is it possible to easily separate plotting from saving here? I feel like saving itself should be optional

MischaPanch: Note that the user might want to decide how to save files - as svg, with different qualities, to different places, etc. You could just return figure or ax objects and let the downstream code save them

tianshou/highlevel/experiment.py
```diff
@@ -86,6 +87,31 @@
 log = logging.getLogger(__name__)
 
 
+def shortener(input_string: str | None = None, length: int = 1) -> str:
+    """Shorten the input string by keeping only the first `length` characters of each word.
+
+    If the input string is None or empty, return "default".
```
MischaPanch: No need for such a behavior in this function

tianshou/highlevel/experiment.py
```diff
@@ -86,6 +87,31 @@
 log = logging.getLogger(__name__)
 
 
+def shortener(input_string: str | None = None, length: int = 1) -> str:
+    """Shorten the input string by keeping only the first `length` characters of each word.
```
MischaPanch: How are words separated? The separator should be configurable

tianshou/highlevel/experiment.py
```diff
@@ -86,6 +87,31 @@
...
+
+    If the input string is None or empty, return "default".
+    """
+    if input_string is None or len(input_string) == 0:
+        return "default"
+    output_parts = []
+
+    for part in input_string.split(","):
+        key, value = part.split("=")
```
MischaPanch: This doesn't do what the docstring says, nowhere is `=` mentioned.

What is the reason behind creating this function at all? Let's discuss in a short call instead of here

tianshou/highlevel/experiment.py
```diff
@@ -442,6 +484,46 @@ def build(self) -> Experiment:
...
+            exp = self.build()
+
+            full_name = ",".join(
+                [
+                    f"experiment_seed={exp.config.seed}",
+                    f"train_seed={exp.sampling_config.train_seed}",
+                    f"test_seed={exp.sampling_config.test_seed}",
+                ],
+            )
+            experiment_name = shortener(full_name, 4)
```
MischaPanch: I think instead the experiment needs a name. Why does one need to return a dict instead of a list of experiments?

examples/mujoco/mujoco_ppo_hl_multi.py
```diff
@@ -0,0 +1,117 @@
+#!/usr/bin/env python3
```
MischaPanch, line 1: There should be a docstring explaining what the script does and how it's different from the other scripts. To work well with CLI tools, should probably be a docstring of the main function

MischaPanch (commented):
Preliminary review, we can have a closer look together later,

One thing to do already is to separate saving plots from plotting, and/or give the user the possibility to configure how plots should be saved

examples/mujoco/mujoco_ppo_hl_multi.py
maxhuettenrauch: of course, forgot to comment in after testing

examples/mujoco/tools.py
maxhuettenrauch: score_threshold can default to None, then it is just inferred from the rewards achieved by the training runs. Regarding the name, maybe we can extract a default from the loaded experiment or from the exp_dir

examples/mujoco/tools.py
maxhuettenrauch: It's basically the number of times the trainer calls test_step(), which is usually once in reset and once after each epoch. We should find a better a way to tell the user what dimensions rliable expects the ndarrays to have (i.e., the arrays in score_dict should be 2D with each row representing a training run and env_steps should be 1D)

examples/mujoco/tools.py
maxhuettenrauch: rliable is actually not too heavyweight, I just kept it in there until we decide which way we go. I'll start with adding an extra `eval`

examples/mujoco/tools.py
maxhuettenrauch: The 0.25 makes makes the Q in IQM :) The reps parameter I need to check

examples/mujoco/tools.py
maxhuettenrauch: True, I just let the function return the ax objects and added a save_figure flag that defaults to None.

tianshou/highlevel/experiment.py
maxhuettenrauch: Just wanted to avoid excessivley long dir names :D Yeah, let's discuss

tianshou/highlevel/experiment.py
maxhuettenrauch: The experiment's path is composed of two parts. The persistence_dir in the experiment_config and a name given to run() of the Experiment. The name created here is just a convenience suggestion to have a (somewhat) meaningful default. It can obviosuly just be a list, too, and the user has to come up with their own names

examples/mujoco/tools.py
MischaPanch: I'll push the solution

tianshou/utils/logger/wandb.py
```diff
@@ -156,3 +160,8 @@ def restore_data(self) -> tuple[int, int, int]:
         except KeyError:
             env_step = 0
         return epoch, env_step, gradient_step
+
+    def restore_logged_data(self, log_path: str) -> dict:
+        # TODO: add support for restoring logged data from W&B
+        assert self.tensorboard_logger is not None
```
MischaPanch: replace by value error or remove optionality of the field

examples/atari/atari_wrapper.py
```diff
@@ -340,7 +340,7 @@ def make_atari_env(
 
     :return: a tuple of (single env, training envs, test envs).
     """
-    env_factory = AtariEnvFactory(task, seed, frame_stack, scale=bool(scale))
+    env_factory = AtariEnvFactory(task, seed, seed + training_num, frame_stack, scale=bool(scale))
```
bordeauxred, line 386:
```suggestion
    env_factory = AtariEnvFactory(task, seed, seed + training_num + 1, frame_stack, scale=bool(scale))
```

bordeauxred: Add the  + 1 for consistency with test_seed property in config.py

examples/mujoco/mujoco_env.py
```diff
@@ -29,7 +29,7 @@ def make_mujoco_env(task: str, seed: int, num_train_envs: int, num_test_envs: in
 
     :return: a tuple of (single env, training envs, test envs).
     """
-    envs = MujocoEnvFactory(task, seed, obs_norm=obs_norm).create_envs(
+    envs = MujocoEnvFactory(task, seed, seed + num_train_envs, obs_norm=obs_norm).create_envs(
```
bordeauxred, line 40:
```suggestion
    envs = MujocoEnvFactory(task, seed, seed + num_train_envs + 1, obs_norm=obs_norm).create_envs(
```

tianshou/utils/logger/pandas_logger.py
```diff
@@ -0,0 +1,123 @@
...
+            test_interval,
+            update_interval,
+            info_interval,
+            exclude_arrays,
+        )
+        self.log_path = log_dir
+        self.csv_name = os.path.join(self.log_path, "log.csv")
+        self.pkl_name = os.path.join(self.log_path, "log.pkl")
+        self.data: dict[str, list] = defaultdict(list)
+        self.last_save_step = -1
```
bordeauxred: Can we make some attributes private?

maxhuettenrauch: removed some of them. from the leftovers, which one should be private?

examples/mujoco/tools.py
maxhuettenrauch: Did a larger refactor to separate loading from preparing for rliable (new location is still not quite right though, as long as rliable is not in standard dependencies right now)

tianshou/highlevel/experiment.py
maxhuettenrauch: removed

tianshou/utils/logger/wandb.py
maxhuettenrauch: done

tianshou/utils/logger/tensorboard.py
```diff
@@ -132,14 +132,17 @@ def restore_data(self) -> tuple[int, int, int]:
...
+    ) -> dict[str, dict[str, VALID_LOG_VALS_TYPE | dict[str, VALID_LOG_VALS_TYPE]]]:
         ea = event_accumulator.EventAccumulator(log_path)
         ea.Reload()
 
         def add_to_dict(data_dict: dict[str, Any], keys: list[str], value: Any) -> None:
             current_dict = data_dict
-            for key in keys[:-1]:
-                current_dict = current_dict.setdefault(key, {})
+            for k in keys[:-1]:
+                current_dict = current_dict.setdefault(k, {})
```
bordeauxred: How about using a defaultdict?

tianshou/utils/logger/tensorboard.py
```diff
@@ -132,14 +132,17 @@ def restore_data(self) -> tuple[int, int, int]:
...
             current_dict[keys[-1]] = value
```
bordeauxred: Since only the last key is assigned a value, I guess this is a fairly specialised add_to_dict. 
You might wish to reflect this in the naming.

examples/mujoco/mujoco_ppo_hl_multi.py
maxhuettenrauch: I added a very basic joblib launcher that allows for parallel experiments on a single machine. We can think about other launchers as well, but that's probably really out of scope.

examples/atari/atari_wrapper.py
maxhuettenrauch: Actually the + 1 is not necessary. I also removed it from config.py

tianshou/utils/logger/pandas_logger.py
maxhuettenrauch: removed the pandas logger from this PR

maxhuettenrauch:
I had to use contextlib.suppress in order to make the docs build, but at least tests are passing now. I also linked our fork of rliable in the project dependencies as long as the original does not update the dependency on arch.

MischaPanch:
> I also linked our fork of rliable in the project dependencies as long as the original does not update the dependency on arch.

Thanks, that works! Seems like rliable is no longer maintained, my PR bumping the arch version got no attention. Let's see for how long we don't need to touch it again. If it creates further problems, might be worth to move the functionality over to tianshou or somewhere else

MischaPanch:
@maxhuettenrauch  as to the contextlib suppression - I don't think that's the best solution. Instead you could adjust the command in the CI, which is currently `poetry install --with dev` to `poetry install --with dev -E evaluation` (or something like that).

Note that you can run github actions locally with [this](https://github.com/nektos/act), maybe it makes sense to have a poe command for it

maxhuettenrauch:
Thanks for the recommendation, unfortunately I haven't managed to get act to work (getting `::error::Not Found
` but no idea what is actually not found)

MischaPanch:
Seems like this is ready, right? Gonna run a few tests tomorrow and then merge it

tianshou/utils/logger/tensorboard.py
MischaPanch: The logic is a bit complicated here, I renamed the function name and added more comments to explain it. Defaultdict wouldn't help here, unless there's something like an infinitely nested default dict ^^

MischaPanch:
I couldn't make parallel execution work with joblib. It also might not make too much sense on a single machine, so we can consider using ray for parallelization instead. For now the limitations are documented, so I'd merge this.

In my last commits I improved the plots a bit (axis labels were cut off), added more docs, fixed some installation problems and did minor enhancements to the interfaces

MischaPanch (approved):

PR #1077, opened 2024-03-21
Fix mypy issues in tests and examples 

### LINKED ISSUE #952: Make tests and examples pass  mypy

dantp-ai:
Closes #952 

- `SamplingConfig` supports `batch_size=None`. #1077
- tests and examples are covered by `mypy`. #1077
- `NetBase` is more used, stricter typing by making it generic. #1077
- `utils.net.common.Recurrent` now receives and returns a `RecurrentStateBatch` instead of a dict. #1077


dantp-ai:
@MischaPanch I have finished solving the issues. Below is a list of commits that I believe require more attention than others:

- https://github.com/thu-ml/tianshou/pull/1077/commits/7c91575a69721270b4fe82c67439d70a8fd28ec3
  - I used the [recommendations from the cheatsheet](https://tianshou.org/en/stable/01_tutorials/07_cheatsheet.html#user-defined-environment-and-different-state-representation).
- https://github.com/thu-ml/tianshou/pull/1077/commits/6f31ac12e935820e0d878d4673ef30aaef41daff
  - ALE and Gym API pre (only done flag) and post version 0.26 (terminated, truncated flags).
- https://github.com/thu-ml/tianshou/pull/1077/commits/f282d814afb3b1a76d4cb97535051de752e9527f
  - I think here one could solve it differently by storing temporary variables and asserting their type, and then continue using dot-notation.
- https://github.com/thu-ml/tianshou/pull/1077/commits/56608dad369dd92cc597396273082687b1a76905
- https://github.com/thu-ml/tianshou/pull/1077/commits/2d6f5f171df0420ee4914e53a0c47d0ff8176a8c
  - mypy does not seem to understand dynamically added attrs to the dataclass.
- https://github.com/thu-ml/tianshou/pull/1077/commits/10352d4a97c19c771696eeefe69e0fb6d03972b0
  - Respect >0.26 Gym API for <0.26 API calls.
- https://github.com/thu-ml/tianshou/pull/1077/commits/2255f6e7c72a6a446ef8d0c2d55c3c79e5ccc292
  - Should `BranchingDQNPolicy` support also `MultiDiscrete` action spaces?
  - I think `tianshou.env.ContinuousToDiscrete` should be renamed to `ContinuousToMultiDiscrete` as it converts action space from `Box` to `MultiDiscrete`. `Discrete` is a special case of `MultiDiscrete`. Also, as it converts only the action space, a better name would be `ContinuousActionSpaceToMultiDiscrete`.

dantp-ai:
@MischaPanch Should I first merge the Collector-refactoring PR (https://github.com/thu-ml/tianshou/pull/1063) into here, before doing the review of this PR?

examples/atari/atari_dqn.py
```diff
@@ -104,7 +104,8 @@ def test_dqn(args: argparse.Namespace = get_args()) -> None:
     net = DQN(*args.state_shape, args.action_shape, args.device).to(args.device)
     optim = torch.optim.Adam(net.parameters(), lr=args.lr)
     # define policy
-    policy: DQNPolicy = DQNPolicy(
+    policy: DQNPolicy | ICMPolicy
```
MischaPanch, line 107: This shouldn't be necessary, did you look into it?

examples/atari/atari_dqn.py
```diff
@@ -157,6 +158,7 @@ def test_dqn(args: argparse.Namespace = get_args()) -> None:
     log_path = os.path.join(args.logdir, log_name)
 
     # logger
+    logger_factory = LoggerFactoryDefault()
```
MischaPanch, line 161: Why was this change necessary?

examples/atari/atari_network.py
```diff
@@ -251,12 +252,25 @@ def __init__(
         self.features_only = features_only
 
     def create_module(self, envs: Environments, device: TDevice) -> Actor:
+        obs_shape = envs.get_observation_shape()
+        if isinstance(obs_shape, int):
+            obs_shape = [obs_shape]
+        assert len(obs_shape) == 3
```
MischaPanch: pls replace by a ValueError with an informative message

examples/atari/atari_network.py
```diff
@@ -251,12 +252,25 @@ def __init__(
         self.features_only = features_only
 
     def create_module(self, envs: Environments, device: TDevice) -> Actor:
+        obs_shape = envs.get_observation_shape()
+        if isinstance(obs_shape, int):
+            obs_shape = [obs_shape]
```
MischaPanch: This is in direct contradiction with the assert below, I guess obs_shape cannot be int?

examples/atari/atari_network.py
```diff
@@ -251,12 +252,25 @@ def __init__(
...
+        assert len(obs_shape) == 3
+        c, h, w = obs_shape
+        action_shape = envs.get_action_shape()
+        if isinstance(action_shape, np.int64):
+            action_shape = int(action_shape)
+        net: nn.Module
         net = DQN(
-            *envs.get_observation_shape(),
-            envs.get_action_shape(),
+            c,
```
MischaPanch: could you use kwargs here pls? (for better readability)

examples/atari/atari_network.py
```diff
@@ -251,12 +252,25 @@ def __init__(
...
     def create_module(self, envs: Environments, device: TDevice) -> Actor:
+        obs_shape = envs.get_observation_shape()
+        if isinstance(obs_shape, int):
+            obs_shape = [obs_shape]
+        assert len(obs_shape) == 3
+        c, h, w = obs_shape
+        action_shape = envs.get_action_shape()
+        if isinstance(action_shape, np.int64):
+            action_shape = int(action_shape)
+        net: nn.Module
```
MischaPanch: why is this necessary?

examples/atari/atari_network.py
```diff
@@ -251,12 +252,25 @@ def __init__(
         self.features_only = features_only
 
     def create_module(self, envs: Environments, device: TDevice) -> Actor:
+        obs_shape = envs.get_observation_shape()
+        if isinstance(obs_shape, int):
+            obs_shape = [obs_shape]
+        assert len(obs_shape) == 3
+        c, h, w = obs_shape
+        action_shape = envs.get_action_shape()
+        if isinstance(action_shape, np.int64):
```
MischaPanch, line 261: Is it possible to move the casting to int to `get_action_shape` instead? It might create problems with envpool, which we don't control.

If envpool is the reason for this line, pls add a block comment about it

examples/atari/atari_network.py
```diff
@@ -270,9 +284,19 @@ def __init__(self, features_only: bool = False, net_only: bool = False) -> None:
         self.net_only = net_only
 
     def create_intermediate_module(self, envs: Environments, device: TDevice) -> IntermediateModule:
+        obs_shape = envs.get_observation_shape()
+        if isinstance(obs_shape, int):
```
MischaPanch, line 286: same as above. Also note the code duplication

examples/atari/atari_network.py
```diff
@@ -251,12 +252,25 @@ def __init__(
...
-            *envs.get_observation_shape(),
-            envs.get_action_shape(),
+            c,
+            h,
+            w,
+            action_shape,
             device=device,
             features_only=self.features_only,
-            output_dim=self.hidden_size,
+            output_dim=self.hidden_size
```
MischaPanch: this is a logical change, why did it work before this? Have you tested that atari examples run through?

examples/atari/atari_ppo_hl.py
```diff
@@ -98,10 +98,11 @@ def main(
         .with_epoch_stop_callback(AtariEpochStopCallback(task))
     )
     if icm_lr_scale > 0:
+        hidden_sizes = [hidden_sizes] if isinstance(hidden_sizes, int) else hidden_sizes
```
MischaPanch: I feel like hidden sizes should always be a sequence. @opcode81 was there a reason you allowed an int to be passed as well? For backwards compatibility?

I backwards compat was the reason, I think we should break it here in favor of simpler types

examples/box2d/bipedal_bdq.py
```diff
@@ -98,11 +101,11 @@ def test_bdq(args: argparse.Namespace = get_args()) -> None:
         device=args.device,
     ).to(args.device)
     optim = torch.optim.Adam(net.parameters(), lr=args.lr)
-    policy = BranchingDQNPolicy(
+    policy: BranchingDQNPolicy = BranchingDQNPolicy(
```
MischaPanch, line 104: This kind of annotation should really not be necessary. Does mypy throw an error?

examples/box2d/bipedal_bdq.py
```diff
@@ -125,7 +128,9 @@ def save_best_fn(policy: BasePolicy) -> None:
         torch.save(policy.state_dict(), os.path.join(log_path, "policy.pth"))
 
     def stop_fn(mean_rewards: float) -> bool:
-        return mean_rewards >= getattr(env.spec.reward_threshold)
+        if env.spec and env.spec.reward_threshold:
```
MischaPanch, line 131: How did this ever work before? Isn't getattr supposed to get two params? Here it got one

examples/mujoco/fetch_her_ddpg.py
```diff
@@ -96,33 +101,37 @@ def test_ddpg(args: argparse.Namespace = get_args()) -> None:
...
+    logger = logger_factory.create_logger(
+        log_dir=log_path,
+        experiment_name=log_name,
+        run_id=args.resume_id,
+        config_dict=vars(args),
+    )
 
     env, train_envs, test_envs = make_fetch_env(args.task, args.training_num, args.test_num)
+    # The method HER works with goal-based environments
+    assert isinstance(env.observation_space, gym.spaces.Dict)
```
MischaPanch: pls add error messages behind the asserts. You can also replace them by ValueErrors

examples/mujoco/fetch_her_ddpg.py
```diff
@@ -96,33 +101,37 @@ def test_ddpg(args: argparse.Namespace = get_args()) -> None:
...
+    assert hasattr(env, "compute_reward")
```
MischaPanch: Especially this one requires info, what is the user supposed to do if the attr is missing?

Generally, one shouldn't rely on hasattr - why is it necessary?

examples/mujoco/mujoco_ddpg.py
```diff
@@ -115,6 +115,7 @@ def test_ddpg(args: argparse.Namespace = get_args()) -> None:
         print("Loaded agent from: ", args.resume_path)
 
     # collector
+    buffer: VectorReplayBuffer | ReplayBuffer
```
MischaPanch, line 118: Shouldn't be necessary

examples/mujoco/mujoco_reinforce_hl.py
```diff
@@ -29,7 +29,7 @@ def main(
     step_per_epoch: int = 30000,
     step_per_collect: int = 2048,
     repeat_per_collect: int = 1,
-    batch_size: int | None = None,
+    batch_size: int = 16,
```
MischaPanch: this changes the logic! Revert pls

examples/mujoco/plotter.py
```diff
@@ -3,14 +3,20 @@
...
 import matplotlib.ticker as mticker
 import numpy as np
 from tools import csv2numpy, find_all_files, group_files
 
 
-def smooth(y, radius, mode="two_sided", valid_only=False):
+def smooth(
+    y: np.ndarray,
+    radius: int,
+    mode: str = "two_sided",
```
MischaPanch: pls use Literal as type, then you can remove the assert

examples/offline/atari_bcq.py
```diff
@@ -82,17 +83,23 @@ def test_discrete_bcq(args: argparse.Namespace = get_args()) -> None:
         scale=args.scale_obs,
         frame_stack=args.frames_stack,
     )
-    args.state_shape = env.observation_space.shape or env.observation_space.n
-    args.action_shape = env.action_space.shape or env.action_space.n
+    assert isinstance(env.action_space, Discrete)
+    args.state_shape = env.observation_space.shape
```
MischaPanch, line 87: one shouldn't modify args. I understand it didn't come in with your PR, but since you're touching it, could you remove it and just use `state_shape = env.observation_space.shape`, or even just directly use `env.observation_space.shape` without creating a new var

examples/offline/atari_cql.py
```diff
@@ -80,16 +82,21 @@ def test_discrete_cql(args: argparse.Namespace = get_args()) -> None:
...
+    c, h, w = args.state_shape
     # should be N_FRAMES x H x W
     print("Observations shape:", args.state_shape)
     print("Actions shape:", args.action_shape)
     # seed
     np.random.seed(args.seed)
     torch.manual_seed(args.seed)
     # model
-    net = QRDQN(*args.state_shape, args.action_shape, args.num_quantiles, args.device)
+    net = QRDQN(c, h, w, args.action_shape, args.num_quantiles, args.device)
```
MischaPanch: As above. Generally, I feel like the init of such classes should be kwargs only. If you feel like it's too much, can be done in a separate PR

examples/vizdoom/env.py
```diff
@@ -81,9 +94,9 @@ def reset(self):
         self.health = self.game.get_game_variable(vzd.GameVariable.HEALTH)
         self.killcount = self.game.get_game_variable(vzd.GameVariable.KILLCOUNT)
         self.ammo2 = self.game.get_game_variable(vzd.GameVariable.AMMO2)
-        return self.obs_buffer
+        return self.obs_buffer, {}
```
MischaPanch: This is breaking - I guess it was a bug that was previously undiscovered?

examples/vizdoom/env.py
```diff
@@ -105,17 +118,27 @@ def step(self, action):
...
-    test_num = min(os.cpu_count() - 1, test_num)
+def make_vizdoom_env(
+    task: str,
+    frame_skip: int,
+    res: tuple[int],
+    save_lmp: bool = False,
+    seed: int | None = None,
+    training_num: int = 10,
+    test_num: int = 10,
+) -> tuple[Any | Env, Any | ShmemVectorEnv, Any | ShmemVectorEnv]:
```
MischaPanch: I think `Any | ...` doesn't make much sense. Why do we need `Any`? If there's no better type, maybe just leave `Any` for now? 

Does this maybe merit a separate issue?

examples/vizdoom/replay.py
```diff
@@ -6,7 +6,7 @@
 import vizdoom as vzd
 
 
-def main(cfg_path="maps/D3_battle.cfg", lmp_path="test.lmp") -> None:
+def main(cfg_path: str = "maps/D3_battle.cfg", lmp_path: str = "test.lmp") -> None:
```
MischaPanch: It's a linux path - replace by `os.path.join`?

examples/vizdoom/vizdoom_ppo.py
```diff
@@ -140,7 +140,8 @@ def test_ppo(args: argparse.Namespace = get_args()) -> None:
     def dist(logits: torch.Tensor) -> Distribution:
         return Categorical(logits=logits)
 
-    policy: PPOPolicy = PPOPolicy(
+    policy: PPOPolicy | ICMPolicy
```
MischaPanch: as above

test/base/test_buffer.py
```diff
@@ -164,32 +166,34 @@ def test_ignore_obs_next(size=10) -> None:
...
-                [4, 4, 5, 6],
-                [4, 4, 5, 6],
-                [7, 7, 7, 8],
-                [7, 7, 8, 9],
-                [7, 7, 8, 9],
-            ],
-        ),
-    )
-    assert np.allclose(data.info["if"], data2.info["if"])
+    if hasattr(data.obs_next, "mask") and hasattr(data2.obs_next, "mask"):
```
MischaPanch: instead of `if hasattr` the test should do `assert hasattr` before the equality tests

test/base/test_collector.py
```diff
@@ -210,16 +236,19 @@ def test_collector(gym_reset_kwargs) -> None:
     with pytest.raises(TypeError):
         c2.collect()
 
+    def create_env(i: int, t: str) -> Callable[[], NXEnv]:
+        return lambda: NXEnv(i, t)
+
     # test NXEnv
     for obs_type in ["array", "object"]:
-        envs = SubprocVectorEnv([lambda i=x, t=obs_type: NXEnv(i, t) for x in [5, 10, 15, 20]])
+        envs = SubprocVectorEnv([create_env(x, obs_type) for x in [5, 10, 15, 20]])
```
MischaPanch: this is less safe than before because of [late binding](https://docs.python-guide.org/writing/gotchas/#late-binding-closures) in functions, a common python gotcha. Pls pass as kwargs instead to bind immediately.

test/base/test_env_finite.py
```diff
@@ -62,85 +72,104 @@ def step(self, action):
...
 
-    def step(self, action, id=None):
-        id = self._wrap_id(id)
-        id2idx = {i: k for k, i in enumerate(id)}
-        request_id = list(filter(lambda i: i in self._alive_env_ids, id))
-        result = [[None, 0.0, False, False, None] for _ in range(len(id))]
+        obs: np.ndarray
+        try:
+            obs = np.stack(obs_list)
+        except ValueError:
```
MischaPanch: Why is the try-except now necessary?

test/base/test_returns.py
```diff
@@ -70,7 +71,7 @@ def test_episodic_returns(size=2560) -> None:
         truncated=np.array([0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]),
         rew=np.array([101, 102, 103.0, 200, 104, 105, 106, 201, 107, 108, 109, 202]),
     )
-    for b in batch:
+    for b in iter(batch):
```
MischaPanch: With the newest PR merged, `__iter__` was added to Batch. It should be ok to remove the explicit iter now, right?

test/base/test_stats.py
```diff
@@ -37,4 +38,8 @@ def test_training_stats_wrapper() -> None:
         # existing fields, wrapped and not-wrapped, can be mutated
         wrapped_train_stats.loss_field = 13
         wrapped_train_stats.dummy_field = 43
-        assert wrapped_train_stats.wrapped_stats.loss_field == wrapped_train_stats.loss_field == 13
+        assert (
+            getattr(wrapped_train_stats.wrapped_stats, "loss_field")  # noqa: B009
```
MischaPanch: no need to use getattr, you can use `type: ignore` or `assert hasattr` to make mypy happy

tianshou/data/batch.py
```diff
@@ -17,7 +18,14 @@
...
 import torch
 
-IndexType = slice | int | np.ndarray | list[int]
+_SingleIndexType = slice | int | EllipsisType
+IndexType = (
+    np.ndarray
+    | _SingleIndexType
+    | list[_SingleIndexType]
+    | tuple[_SingleIndexType, ...]
+    | list[int]
```
MischaPanch: this shouldn't be necessary, already covered by `list[_SingleIndexType]`

MischaPanch (changes requested):
Great work @dantp-ai , thanks a  lot for it! It's going to be super useful.

I left a bunch of comments, but apart from them the tests and examples should now be included in the poe task `type-check`, you should modify `pyproject.toml` and `.pre-commit.yaml` for that.

Feel free to give me a call if you want to talk about the comments

MischaPanch:
> @MischaPanch Should I first merge the Collector-refactoring PR (#1063) into here, before doing the review of this PR?

@dantp-ai the PR was now merged into master. Could you pls merge master into your branch and resolve the conflicts?

MischaPanch:
In fact, I'm going to merge #1032 first and would then resolve the conflicts myself and push to your branch

examples/atari/atari_dqn.py
MischaPanch: Ahh, it's necessary b/c of an open issue in mypy: https://github.com/python/mypy/issues/3737

test/base/test_collector.py
MischaPanch: Done

MischaPanch:
@dantp-ai Done now, conflicts resolved, pipeline should run through. I extended the pyproject and pre-commit config. When the remaining issues are addressed, this can be merged.

Pls extend the changelog and copy the relevant parts of the changelog to the PR description (see the last merged PRs for examples on PR descriptions)

examples/atari/atari_dqn.py
dantp-ai: I didn't know about the issue in mypy, thx. 

I looked into our issue, whether it's b/c `policy` is defined multiple times with different policies, but mypy wants to type the `policy` even if it is defined once.

Btw, [in the overview tutorial](https://github.com/thu-ml/tianshou/blob/master/docs/02_notebooks/L0_overview.ipynb) `policy` is typed with `BasePolicy`. I think it should be typed with `PPOPolicy` there so that we get all the `PPOPolicy`-specific attrs and methods in the IDE.

examples/atari/atari_dqn.py
dantp-ai: A lot of the examples where duplicating code related to logging that already exists in `LoggerFactoryDefault.create_logger()`. [Initially](https://github.com/thu-ml/tianshou/pull/1017#discussion_r1481301565), I created a common `logger_factory`, but as you correctly pointed out that solution is leading to mutable default throughout the code.

The solution here and in other examples is to have a per-example `LoggerFactoryDefault`.

examples/atari/atari_ppo_hl.py
dantp-ai: Yes, a sequence for `hidden_sizes` would be way easier to work with and it includes the case with one layer. So no need to distinguish it with a special type (if that is the reason of allowing an `int` now).

examples/atari/atari_network.py
dantp-ai: > Have you tested that atari examples run through?

Yes.

> why did it work before this?

All the examples which use `ActorFactoryAtariDQN` are passing an int for `hidden_size`.

> this is a logical change,

Yes, so I assume that if it's a sequence of hidden sizes, then the last value should be the last layer which coincides with the output dimension.

examples/box2d/bipedal_bdq.py
dantp-ai: Yes true, it would have led to an error, but `stop_fn` is not being used in this example. For unknown reasons it's commented out when passed to the `OffpolicyTrainer` [here](https://github.com/thu-ml/tianshou/pull/1077/files/d01420c714d0a2bac82cd72027b8c7b3cca265ba#diff-3792a1a111e15c573f79326f5922fa97aad9f2022aa90d13a088ffada9b78434R153).

examples/mujoco/fetch_her_ddpg.py
dantp-ai: > Generally, one shouldn't rely on hasattr - why is it necessary?

Not all Gym environments have `compute_reward` attr. Mypy doesn't know that in this case `FetchReach-v2` or other [goal-based envs](https://robotics.farama.org/envs/fetch/index.html) are being used for HER-based algos.

examples/mujoco/mujoco_reinforce_hl.py
dantp-ai: I introduced here an `int` by default because `None` would not be compatible with `batch_size` in `SampleConfig` which is always an `int`. 

If we allow `None`, there is an extra (assert) check to be done to make mypy happy.

Should we rather change the default value from 16 to 64 as it is by default in `SampleConfig` (used a couple of lines below) ?

tianshou/data/batch.py
dantp-ai: I can remove it now. There are no issues when not explicitly defining `list[int]`. 

Previously, mypy was complaining because in `Collector._reset_state`, it couldn't infer that `list[_SingleIndexType]` includes `list[int]` too.

MischaPanch: Sounds good

examples/atari/atari_dqn.py
MischaPanch: Ok, that's reasonable

examples/atari/atari_dqn.py
MischaPanch: Right, would be better - could you fix it in the tutorial pls?

examples/atari/atari_ppo_hl.py
MischaPanch: I don't think there's a good reason for the int at all. Should it be done in a separate PR, or is the change small enough that you want to do it here?

examples/box2d/bipedal_bdq.py
MischaPanch: Wanna fix this? :)

I guess someone accidentally broke it and then just commented it out

examples/mujoco/fetch_her_ddpg.py
MischaPanch: Pls add this explanation as content of a ValueError

examples/mujoco/mujoco_reinforce_hl.py
MischaPanch: There is no good default value for batch size. None has a certain semantics now which was introduced with a PR not long ago. I guess sampling config should accept None and the docstring should be extended. If you want, I can do that

test/base/test_returns.py
dantp-ai: Correct.

examples/mujoco/mujoco_reinforce_hl.py
dantp-ai: > If you want, I can do that

OK.

examples/mujoco/plotter.py
dantp-ai: Yes. it's cleaner.

examples/atari/atari_dqn.py
dantp-ai: Yes.

examples/box2d/bipedal_bdq.py
dantp-ai: Yes. I'm running the script now to test it.

examples/atari/atari_ppo_hl.py
dantp-ai: It can be done here. There are only two examples using `ActorFactoryAtariDQN`.

dantp-ai: `ActorFactoryAtariDQN` currently names it `hidden_size`. Better naming for sequence is `hidden_sizes`.

examples/atari/atari_network.py
dantp-ai: If we go with the change in `ActorFactoryAtariDQN` of `hidden_size: int | Sequence[int]` to `hidden_sizes: Sequence[int]` then the logic above to `output_dim` needs to be considered, i.e., the `output_dim` is in the last layer of the `hidden_sizes` sequence.

examples/atari/atari_dqn.py
dantp-ai: Done.

examples/mujoco/fetch_her_ddpg.py
dantp-ai: Done.

examples/box2d/bipedal_bdq.py
dantp-ai: The `epoch` was set by default to 1000. I guess (and hope) that the agent can reach `reward_threshold` way sooner (like 10-20 epochs as in other examples). I am running now the script to determine what the lowest epoch is s.t. `stop_fn` is true.

examples/vizdoom/env.py
dantp-ai: > I guess it was a bug that was previously undiscovered?

Yes, kinda of. There is no bug in the code since the `info` dict is not used when the `env` is reset. However, it does not respect the Gym API.

examples/vizdoom/env.py
dantp-ai: Thx for spotting it. It should just be `tuple[Env, ShmemVectorEnv, ShmemVectorEnv]`.
I must have used VSCode's double-click to insert the type annotation there.

examples/mujoco/mujoco_reinforce_hl.py
MischaPanch: Done

MischaPanch (approved):

examples/atari/atari_wrapper.py
```diff
@@ -191,17 +195,24 @@ class WarpFrame(gym.ObservationWrapper):
...
+    def __init__(self, env: gym.Env) -> None:
         super().__init__(env)
         self.size = 84
+        obs_space = env.observation_space
+        obs_space_dtype: type[np.floating[Any]] | type[np.integer[Any]]
+        if np.issubdtype(type(obs_space.dtype), np.integer):
+            obs_space_dtype = np.integer
+        elif np.issubdtype(type(obs_space.dtype), np.floating):
+            obs_space_dtype = np.floating
+        assert isinstance(obs_space, gym.spaces.Box)
```
maxhuettenrauch, line 207: Hi @dantp-ai ! Could you let me know what was the reason for this change? It leads to a broken DQN example on Atari for me due to `obs_space_dtype` being undefined. The type check only works if I check `np.issubdtype(obs_space.dtype, np.integer)`, otherwise it will be False...

dantp-ai: Let me take a look how we can write this better.

The reason for the modification was a type mismatch previously between what `Box` allows for `dtype` compared to the more general type of `Space` for `dtype`:

```
examples/atari/atari_wrapper.py: note: In member "__init__" of class "WarpFrame":
examples/atari/atari_wrapper.py:212: error: Argument "dtype" to "Box" has incompatible type
"dtype[Any] | dtype[void] | None"; expected "type[floating[Any]] | type[integer[Any]]"  [arg-type]
                dtype=env.observation_space.dtype,
                      ^~~~~~~~~~~~~~~~~~~~~~~~~~~
```

maxhuettenrauch: ... and also a numpy DeprecationWarning for setting np.integer as dtype.

dantp-ai: I found a solution:

```
assert isinstance(obs_space, gym.spaces.Box)

dtype: type[np.integer] | type[np.floating]
if np.issubdtype(obs_space.dtype, np.floating):
    dtype = np.float32
elif np.issubdtype(obs_space.dtype, np.integer):
    dtype = np.int32

assert dtype is not None, "dtype for `Box` space must be type[floating[Any]] | type[integer[Any]]"

self.observation_space = gym.spaces.Box(
            low=np.min(obs_space.low),
            high=np.max(obs_space.high),
            shape=(self.size, self.size),
            dtype=dtype,
        )
```

One concern I have is that there could be other suitable floating-point types (e.g., `np.float64`) based on the system.

Since this mypy-related code gets so complex, we might just add a type ignore.

maxhuettenrauch: Yeah, I also feel that we shouldn't overwrite the dtype of the original observation space. What do you think @MischaPanch ?

dantp-ai: Thx for spotting it. Although I tried to run the examples whenever I changed something, I missed this one.

This makes me ponder whether we shouldn't add some small smoke tests (for now) for the examples as well?

maxhuettenrauch: @MischaPanch , this is probably the root for issue #1111

dantp-ai: Yes, it's definitely due to the issue here. Should I go ahead and ignore the type as it becomes so complex to implement it (cf. above one possible solution)?

MischaPanch: Sorry, it had escaped my attention. How about just replacing the assert by cast?

dantp-ai: Unfortunately casting the `obs_space` is weaker than the assert and mypy will complain that some attrs don't exist because it won't assume the space to be `Box`.

MischaPanch: I'll check and will push the fix directly to master. We need to find some way of testing the examples, I'm going to contact some people from AMD and ask them to donate us CI workers

dantp-ai: Ok, mind that this dtype issue is in `WarpFrame` and `FrameStack`.

MischaPanch: That was just a bug - the `type` was too much in the isinstance check. I factored it out into a function and fixed it :)

PR #1098, opened 2024-04-04
Allow two (same/different) Batch objs to be tested for equality

### LINKED ISSUE #1086: Batch: don't create new objects on getitem
Currently `Batch.__getitem__` will always create a new object. This is counterintuitive and destroys equality checks. E.g.,

```python
b = Batch(...)
id1 = id(b[0]) 
id2 = id(b[0])
````

will result in `id1 != id2`, which leads to `b[0] == b[0]` being `False`

Related to #922

### PR DESCRIPTION
**Pull request by dantp-ai**
Closes: https://github.com/thu-ml/tianshou/issues/1086

### Api Extensions

- Batch received new method: `to_numpy_`. #1098
- `to_dict` in Batch supports also non-recursive conversion. #1098
- Batch `__eq__` now implemented, semantic equality check of batches is now possible. #1098

### Breaking Changes

- The method `to_numpy` in `data.utils.batch.Batch` is not in-place anymore. Instead, a new method `to_numpy_` does the conversion in-place. #1098

### CODE REVIEW

File: tianshou/data/batch.py [hunk 1]
```diff
@@ -268,6 +268,9 @@ def __repr__(self) -> str:
     def __iter__(self) -> Iterator[Self]:
         ...
 
+    def __eq__(self, other: Self) -> bool:  # type: ignore
```
MischaPanch: why type ignore?

The second argument should be of type Any, and in the implementation you should do:

```
if not isinstance(other, self.__class__):
    return False
```

before any other logic. You can't restrict what a user will put on the other side of  `==`

tianshou/data/batch.py
```diff
@@ -500,6 +503,35 @@ def __getitem__(self, index: str | IndexType) -> Any:
             return new_batch
         raise IndexError("Cannot access item from empty Batch object.")
 
+    def __eq__(self, other: Self) -> bool:  # type: ignore
+        this_dict = self.__dict__
```
MischaPanch: See my comment above - without the isinstance this can give equality between different object that just both happen to have the same `__dict__`. Moreover, `==` with a non-batch object can cause an attribute error.

tianshou/data/batch.py
```diff
@@ -500,6 +503,35 @@ def __getitem__(self, index: str | IndexType) -> Any:
...
+        other_dict = other.__dict__
+
+        if len(this_dict) != len(other_dict):
```
MischaPanch: This is a bit complicated. Could you pls extend the implementation of `to_dict` with `recursive=False`, and thereby allow converting a `Batch` into a pure `python-numpy-torch` object. Then you can probably use [DeepDiff](https://github.com/seperman/deepdiff) to implement equality check

test/base/test_batch.py
```diff
@@ -565,6 +565,47 @@ def test_batch_standard_compatibility() -> None:
         Batch()[0]
 
 
+def test_batch_eq() -> None:
+    # Different keys
```
MischaPanch: pls add testing equality between non-batch objects

also pls add a test where there's first equality, then a value in an array inside one of the instances is modified which should lead to non-equality.

There's also no coverage of equality being true for simple cases - two empty batches, two batches with the same keys in different orders, two batches where one sequence is passed as tuple and one as list, two batches where a subbatch is passed as dict and the other as `Batch`.

Btw, copilot/ChatGPT can likely generate these tests directly from my comment :)

MischaPanch (changes requested):

tianshou/data/batch.py
dantp-ai: I ignored it because the current typing is in conflict with the super-class type that allows `other: object`.

You are right, we cannot control what the user will use for equality, so we should just make the check above right at the beginning.

test/base/test_batch.py
dantp-ai: Yep. I also realize that my assert messages are not expressing the opposite of their truthness. For ex.:

```
# Different keys
batch1 = Batch(a=[1, 2], b=[100, 50])
batch2 = Batch(b=[1, 2], c=[100, 50])
assert batch1 != batch2, "Keys are not matching."
```

Should be "Keys should not match." or "Keys are matching."

MischaPanch: For the future (or now, if you want to): it's better  to have an informative test name and many small tests instead of having many asserts and messages within a single test. This way the cases are all executed instead of being aborted after something in the sequence fails. For example, here I would have done:

```python
class TestBatchEquality:

   @staticmethod
   def test_keys_equal():
       assert Batch(a=.., b=...) == Batch(a=.., b=...)
       assert Batch(a=.., b=...) == Batch(b=.., a=...)
    

   @staticmethod
   def test_keys_unequal():
      ...
```

Again, if you asked ChatGPT to refactor this to such a structure, it would likely do a good job

MischaPanch: Note that now neither comments nor assert messages are necessary

tianshou/data/batch.py
dantp-ai: `DeepDiff` is great, but it seems that it doesn't work for `torch.Tensor`s. 

One solution would be to recursively parse the two Batch objects and convert any torch tensors to numpy arrays before applying `DeepDiff`.

I am investigating now whether there is some way to use DeepDiff's API and overcome this problem.

dantp-ai: I tried to pass a custom `hasher` to `DeepDiff` that would convert a `torch.Tensor` to `np.ndarray` but w/o success.

Then, I skimmed through the code and saw that `Batch` already has a convenient method `to_numpy()` which changes all `torch.Tensor` to `np.ndarray` albeit in-place.

So, what we can do here is call `to_numpy()` before converting to `dict` and comparing differences with `DeepDiff`.

I think we need to implement a non in-place version of `to_numpy()` here because:

- when testing for inequality the tensors of the two objects may be converted to numpy and the user is unaware
- it is weird that the types of the internal objects change when testing for equality.

I also saw that there is a non in-place `to_numpy()` function in `tianshou.data.utils.converter`. Using this one though requires some refactoring because currently it imports `Batch` and this leads to a circular import when using it in `Batch`.

dantp-ai: > Could you pls extend the implementation of to_dict with recursive=False

So, the previous implementation of `to_dict` was implicitly recursive. The extended version now has `recurse=False` by default. But I think by default it should remain recursive as this is what you would expect if you'd like to convert some object to dict.

Regarding nomenclature, I agree that `recursive` is better than `recurse`. I went with `recurse` to have some consistency like in `Batch.is_empty()`.

dantp-ai: P.S. IMO, the name of the method `Batch.to_numpy()` is not reflecting what it does. A better name would be `convert_tensor_values_to_numpy`?

On the other hand, if we reasonably assume that the user knows what a `Batch` is [1], then `to_numpy` makes sense because it is understood that the conversion is done on the values of a nested dict-like structure.



[1] "You can consider Batch as a numpy version of python dictionary." ([from the tutorial](https://tianshou.org/en/stable/02_notebooks/L1_Batch.html)).

tianshou/data/batch.py
```diff
@@ -500,6 +504,17 @@ def __getitem__(self, index: str | IndexType) -> Any:
             return new_batch
         raise IndexError("Cannot access item from empty Batch object.")
 
+    def __eq__(self, other: Any) -> bool:
+        if not isinstance(other, self.__class__):
+            return False
+
+        self.to_numpy()
```
MischaPanch: This modifies self and other inplace on a supposedly harmless equality check - this should never happen.

Starting a new thread about the questions you raised regarding `to_numpy`.

1. For other methods we have the convention that things ending in `_` are inplace. So it should really be that `to_numpy_` operates inplace and returns nothing, whereas `to_numpy()` doesn't change the original and returns a new batch. That's a breaking change, so you'd need to check all the places where `to_numpy` is currently being used.
2. I think the name `to_numpy` is ok for the moment, since all values of batch are either batches, tensors or arrays, right? Then it's clear that `to_numpy` would turn all tensors into arrays.

tianshou/data/batch.py
```diff
@@ -433,11 +437,11 @@ def __init__(
             # Feels like kwargs could be just merged into batch_dict in the beginning
             self.__init__(kwargs, copy=copy)  # type: ignore
 
-    def to_dict(self) -> dict[str, Any]:
+    def to_dict(self, recurse: bool = False) -> dict[str, Any]:
```
MischaPanch: I think recursive is better, feel free to rename `recurse`  to `recursive` wherever the former is used

MischaPanch: Btw, I had a look at the empty implementation, which seems very weird to me. I don't feel like the `recurse=False` option should exist at all. A user can always just check for lens of the keys instead. Moreover, the recursive=True corresponds to checking for `len` of the batch.

I feel like `is_empty` can be removed completely. If you agree, no need to rename the `recurse` - I'd set up an issue for remove `is_empty`

MischaPanch: In fact, adding this issue right now, since it's good for newcomers

dantp-ai: `Batch.to_torch()` has same problem because it is also in-place. We can change this in another PR?

tianshou/data/batch.py
dantp-ai: `Batch.to_torch()` has same problem because it is also in-place. We can change this in another PR?

MischaPanch: Yes, thanks for bringing it up! We should generally reduce the number of inplace operations as much as possible

tianshou/data/batch.py
MischaPanch: Could you change to `recursive` pls? Then this is good to go, pls just add the changes to changelog and the PR description.

Again, thanks for the work on this, highly appreciated!

tianshou/data/batch.py
```diff
@@ -433,11 +442,11 @@ def __init__(
...
         result = {}
         for k, v in self.__dict__.items():
-            if isinstance(v, Batch):
-                v = v.to_dict()
+            if recurse and isinstance(v, Batch):
```
MischaPanch: This is also a breaking change since you changed the default to False. Not a problem, but pls add it to the changelog. I'm actually not sure False is the better default, what was your reason?

dantp-ai: The default of recursively converting to dict was correct before. I wanted to revert back my initial change of logic there, and I just forgot. Will add it now.

MischaPanch (approved):

PR #1100, opened 2024-04-05
Fix/reset before collect in procedural examples, tests and hl experiment

maxhuettenrauch:

Calling collector.collect currently fails if the collector is not reset before collecting. Not sure if there's also a bug in collector where it says _...It has only an effect if n_episode is not None..._ for the `reset_before_collect` kwarg as it also happens when collecting `n_step`.


maxhuettenrauch:
@bordeauxred, @MischaPanch fyi

examples/box2d/lunarlander_dqn.py
```diff
@@ -94,6 +94,7 @@ def test_dqn(args: argparse.Namespace = get_args()) -> None:
     )
     test_collector = Collector(policy, test_envs, exploration_noise=True)
     # policy.set_eps(1)
+    train_collector.reset()
     train_collector.collect(n_step=args.batch_size * args.training_num)
```
bordeauxred, line 98: Did you try by chance what happens if we set here random = True (as it is done in a lot of ther algos
)

bordeauxred:
@MischaPanch the examples are not covered by the tests. Does it make to include them in cicd or run them once a day (if at least one commit on this day) 
or sth?

examples/box2d/lunarlander_dqn.py
maxhuettenrauch: Setting `random=True` still requires resetting before collecting, if that is what you meant?

MischaPanch:
> @MischaPanch the examples are not covered by the tests. Does it make to include them in cicd or run them once a day (if at least one commit on this day) or sth?

Once this is merged, we will follow up and write a script which generates the benchmarks that are currently stored in the examples dir and displayed in the docs. Then we can have a manually triggered GH Action that we trigger on each release. Running examples on each PR/daily is too expensive and unnecessary.

test/discrete/test_bdq.py
```diff
@@ -115,7 +115,8 @@ def test_bdq(args: argparse.Namespace = get_args()) -> None:
     )
     test_collector = Collector(policy, test_envs, exploration_noise=False)
     # policy.set_eps(1)
-    train_collector.collect(n_step=args.batch_size * args.training_num, reset_before_collect=True)
+    train_collector.reset()
+    train_collector.collect(n_step=args.batch_size * args.training_num)
```
MischaPanch, line 119: This shouldn't have been necessary as `reset_before_collect=True`. No need to rollback, but just FYI. The tests did run through in CI, so the problem was only in some examples (where stuff was collected before training started)

examples/atari/atari_dqn.py
```diff
@@ -232,6 +232,7 @@ def watch() -> None:
         sys.exit(0)
 
     # test train_collector and start filling replay buffer
+    train_collector.reset()
     train_collector.collect(n_step=args.batch_size * args.training_num)
```
MischaPanch, line 236: Just wondering: why is this necessary at all? It's not like we are filling the buffer to any significant extent. Does training fail without this? Is it something that should be part of the Trainer implementation?

MischaPanch (commented):
Thanks for spotting!

examples/box2d/lunarlander_dqn.py
MischaPanch: Without `random=True`, the steps are collected using the initial, untrained policy. Which I would guess is worse than with a random policy. Ths env is so easy that it likely doesn't matter much.

I suppose this is what @bordeauxred meant.

tianshou/data/collector.py
```diff
@@ -332,7 +332,7 @@ def collect(
         :param no_grad: whether to retain gradient in policy.forward().
         :param reset_before_collect: whether to reset the environment before
             collecting data.
-            It has only an effect if n_episode is not None, i.e.
+            It has only an effect if n_episode is not None, i.e.  # Fixme: this is not true
```
MischaPanch: Feel free to adjust
```suggestion
            Needed if  `collect` is being executed for the first time and the `Collector` instance was never reset before.
            If False, will continue collecting from the current envs' states.
```

examples/box2d/lunarlander_dqn.py
bordeauxred: Correct @MischaPanch.

test/discrete/test_bdq.py
maxhuettenrauch: I was also unsure about this one. When is the right time to pass the `reset_before_collect` and when should one be calling reset directly? I assumed that after instantiation I want to properly reset once.

MischaPanch:
@maxhuettenrauch I thought the sac_with_il seeding problem was fixed now. Is your branch based on the updated master? Or is there some other source of randomness that we're not controlling yet?

examples/atari/atari_dqn.py
maxhuettenrauch: That's a whole different question. I don't think the original DQN paper collected warm up steps before learning. But the line has been in there and unchanged since 2021.

maxhuettenrauch:
> @maxhuettenrauch I thought the sac_with_il seeding problem was fixed now. Is your branch based on the updated master? Or is there some other source of randomness that we're not controlling yet?

I doesn't fail on my machine... and it hasn't failed on other PRs, or has it?

test/discrete/test_bdq.py
MischaPanch: There's really no difference, it was just a convenience flag. We could also remove it if it causes confusion

examples/atari/atari_dqn.py
MischaPanch: That's not even a warm-up. Seems like an arbitrary decision, could you create an issue for that?

MischaPanch:
> > @maxhuettenrauch I thought the sac_with_il seeding problem was fixed now. Is your branch based on the updated master? Or is there some other source of randomness that we're not controlling yet?
> 
> I doesn't fail on my machine... and it hasn't failed on other PRs, or has it?

It fails occasionally :D, the worst kind of failing. We'll see with you next pushed commit

maxhuettenrauch:
Ah wait, this test uses envpool. I can't even test it on my machine. But then maybe something is wrong with envpool env seeding? Or they need to be seeded after numpy and torch seeding?

maxhuettenrauch:
> > > @maxhuettenrauch I thought the sac_with_il seeding problem was fixed now. Is your branch based on the updated master? Or is there some other source of randomness that we're not controlling yet?
> > 
> > 
> > I doesn't fail on my machine... and it hasn't failed on other PRs, or has it?
> 
> It fails occasionally :D, the worst kind of failing. We'll see with you next pushed commit

Weird. It's giving me deterministic results on my machine... but maybe another machine produces different numbers on that seed.

tianshou/data/collector.py
maxhuettenrauch: Removed the comment

examples/box2d/lunarlander_dqn.py
MischaPanch: Should be covered by the issue that generally investigates why/whether this pseudo-warm-up is necessary, resolving

maxhuettenrauch:
Well, I think this still needs some time. The atari_dqn example is broken in another place due to a recent update on master (at least for me).

MischaPanch:
@maxhuettenrauch the L0 notebook fails for some reason, could you check locally pls?

maxhuettenrauch:
> @maxhuettenrauch the L0 notebook fails for some reason, could you check locally pls?

`poe doc-build` and `poe doc-spellcheck` seem to run through on my end. Not sure why it's failing.

MischaPanch:
@maxhuettenrauch this is finished, right? I'd merge it

Do you know whether it's related (or even solves) #1111 ?

MischaPanch (approved):

maxhuettenrauch:
This should (hopefully) cover all instances of the collect before training starts, yes.

Regarding the other issue, I think #1077 broke it.

PR #1105, opened 2024-04-10
Changed .keys() to get_keys() in batch class

### LINKED ISSUE #922: Inconsistent `__iter__` in batch
Batch is essentially a wrapper around a dict. However, iterating over a batch leads to potentially unexpected behavior.

While `list({...}) == list({..}.keys())`, for batch `list(Batch(...))` gives a sequence of `Batch` objects and `list(Batch(...).keys())` gives a list of keys.

This collides with e.g. ruff's recommendation of [not using in-dict-keys](https://beta.ruff.rs/docs/rules/in-dict-keys/). I had to disable this rule in #921 because both `for b in batch` and `for key in batch.keys()` are currently used with their respective semantics throughout the code.

Fixing this would mean defining `__iter__` in `Batch` to return keys and something like creating a new method in Batch called `iter_entries` that would replace the current iteration semantics. This would be a **breaking change**.

@Trinkle23897 what do you think? One can also live with the current status of the "inconsistency", keep ignoring the ruff recommendation, and just document the current behavior somewhere. I would also add dedicated unit tests in that case

arnaujc91:
Solves the inconsistency that `iter(Batch)` is not the same as `Batch.keys()` by "deprecating" the implicit `.keys()` method

Closes: #922

### DIFF
--- a/test/base/test_batch.py
+++ b/test/base/test_batch.py
@@ -556,7 +556,7 @@ def test_batch_standard_compatibility() -> None:
     batch = Batch(a=np.array([[1.0, 2.0], [3.0, 4.0]]), b=Batch(), c=np.array([5.0, 6.0]))
     batch_mean = np.mean(batch)
     assert isinstance(batch_mean, Batch)  # type: ignore  # mypy doesn't know but it works, cf. `batch.rst`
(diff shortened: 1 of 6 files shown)


MischaPanch:
Thanks, merging this!

Issue #1088, opened 2024-04-03
Batch: don't just set 0 when elements have None entries


MischaPanch:
This is essentially a bug due to the current implementation of `__setitem__`:

![image](https://github.com/thu-ml/tianshou/assets/35432522/2867891d-f8a7-49d4-8cf1-44cbe1e68a1c)

Setting `0` when something is off is wrong. This can manifest itself in things like info of reset, where certain values might be unknown (like actions or `env_num`). Then `None` is erroneously replaced by `0`. There are multiple tests covering this erroneous behavior, e.g.,
line 112 in 
![image](https://github.com/thu-ml/tianshou/assets/35432522/724fef11-613a-4397-b999-a1da9b87faa0)

One should instead at least turn this into `NaN` entries

maxhuettenrauch:
As far as I can tell, we reach this point in two cases:

1) Adding an item to the buffer where the new item has an inconsistent structure compared to the already existing structure and
2) updating an existing item with an item that has an inconsistent structure compared to the already existing structure.

I think the main question here is if these operations should be allowed in the first place. In the RL context this probably only happens in case the info dict changes its structure between steps (as happens in MoveToRightEnv in the tests). One could argue, that the env is ill-defined in this case and instead of setting arbitrary default values, the env should be fixed?

maxhuettenrauch:
Regarding your suggestion to set values to NaN instead, assigning `np.nan` will fail on arrays whose dtype is not float (as is the case in the above mentioned test).

MischaPanch:
@maxhuettenrauch unfortunately, in the RL context this is bound to happen because some things might not be known at `reset` that are known at `step`, and thus will contain `None` entries. The prime example is the action, which is not available at `reset`, but some entrances from the `info` might also be missing

MischaPanch:
This issue is strongly related to #1087

maxhuettenrauch:
But when would you add obs directly after reset to the buffer (where a Batch object is created) and only after that retrieve an action, call `step`, and append the rest to this entry?

MischaPanch:
I thought it might happen in the collectors, but maybe I'm mistaken. 

For sure I've seen this happen with info objects somewhere in collector tests - though there it is a bad implementation of the env.

Unfortunately Gymnasium doesn't force any interface on the info dicts which are the main drivers of this problem.We can stop supporting such cases, or ask the use to specify what should happen then explicitly. 

[...]

PR #1106, opened 2024-04-11
Warn on batch.add when missing keys

maxhuettenrauch:
- [ ] I have reformatted the code using `poe format` 
- [ ] I have checked style and types with `poe lint` and `poe type-check`

Not really a solution but at least a warning is issued (addresses #1088 )


MischaPanch:
Is there no way to actually enforce this?

maxhuettenrauch:
I tried with a non-breaking change to Batch by adding an attribute like `set_default_value_on_missing_key`, but Batch basically is unable to have attributes because everything will land in `self.__dict__` and thus treated as part of the data in a Batch instance. Of course one could raise an Exception instead of issuing a warning in the future.

MischaPanch:
Closing for obsolete - I added new functionality that allows recursive checks on keys (to be merged later)

Moreover, this check should also happen on cat and maybe other functions. I believe it should raise an error, so let's fix this once and for all with strict batches that help users spot errors in their envs

Issue #1122, opened 2024-04-24
Adjust locations of setting the policy in train/eval mode


maxhuettenrauch:
Currently, tianshou sets the policy's mode in the trainer and `test_episode` function. The corresponding `training` attribute is then used to determine if a stochastic policy should be evaluated deterministically given that `policy.deterministic_eval` is `True`. This, however, is a misuse as the `training` attribute primarily has influence on modules like dropout and batchnorm. It should always be `False` during data collection and only be `True` inside `policy.learn`.

opcode81:
Max and I have implemented the following solution in #1123:

* We Introduced a new flag `is_within_training_step` which is enabled by the training algorithm when within a training step, where a training step encompasses training data collection and policy updates. This flag is now used by algorithms to decide whether their `deterministic_eval` setting should indeed apply instead of the torch training flag (which was abused!).
* The policy's training/eval mode (which should control torch-level learning only) no longer needs to be set in user code in order to control collector behaviour (this didn't make sense!). The respective calls have been removed.
* The policy should, in fact, always be in evaluation mode when applying data collection, as there is no reason to ever have gradient accumulation enabled for any type of rollout. We thus specifically set the policy to evaluation mode in Collector.collect.

MischaPanch:
@maxhuettenrauch @opcode81 This can be closed, right?

PR #1123, opened 2024-04-24
Adjust locations of setting the policy in train/eval mode

maxhuettenrauch:
Addresses #1122:
* We Introduced a new flag `is_within_training_step` which is enabled by the training algorithm when within a training step, where a training step encompasses training data collection and policy updates. This flag is now used by algorithms to decide whether their `deterministic_eval` setting should indeed apply instead of the torch training flag (which was abused!).
* The policy's training/eval mode (which should control torch-level learning only) no longer needs to be set in user code in order to control collector behaviour (this didn't make sense!). The respective calls have been removed.
* The policy should, in fact, always be in evaluation mode when applying data collection, as there is no reason to ever have gradient accumulation enabled for any type of rollout. We thus specifically set the policy to evaluation mode in Collector.collect. Further, it never makes sense to compute gradients during collection, so the possibility to pass `no_grad=False` was removed.

Further changes:
- Base class for collectors: `BaseCollector`
- New util context managers `in_eval_mode` and `in_train_mode` for torch modules. 
- `reset` of `Collectors` now returns `obs` and `info`. 
- `no-grad` no longer accepted as kwarg of `collect`
- Removed deprecations of `0.5.1` (will likely not affect anyone) and the unused `warnings` module.


MischaPanch:
I'll add the context manager for policies (maybe torch already has one), as discussed, and then this is good to go. Thanks for the extension @maxhuettenrauch !

MischaPanch (dismissed):

MischaPanch:
@ChenDRAG Once finalized (there's still a bit of review and discussion to be done), I'd merge this PR without squashing if you don't mind. It contains somehow unrelated changes and the history would be easier to bisect on commit level if we don't squash

tianshou/policy/modelfree/dqn.py
```diff
@@ -210,6 +210,8 @@ def forward(
         return cast(ModelOutputBatchProtocol, result)
 
     def learn(self, batch: RolloutBatchProtocol, *args: Any, **kwargs: Any) -> TDQNTrainingStats:
+        # set policy in train mode
```
opcode81: I would remove all of these comments, because this is clear from the docstring of `train`.

tianshou/policy/modelfree/a2c.py
```diff
@@ -165,6 +165,9 @@ def learn(  # type: ignore
         *args: Any,
         **kwargs: Any,
     ) -> TA2CTrainingStats:
+        # set policy in train mode
```
opcode81, line 168: Comment should be replaced with a docstring below the definition.

tianshou/policy/base.py
```diff
@@ -225,6 +225,8 @@ def __init__(
         self.action_scaling = action_scaling
         self.action_bound_method = action_bound_method
         self.lr_scheduler = lr_scheduler
+        # whether the policy is in evaluation mode
```
opcode81: Comment should be replaced with a docstring below the definition.

tianshou/policy/base.py
```diff
@@ -225,6 +225,8 @@ def __init__(
...
+        self.is_eval = False  # TODO: remove in favor of kwarg in compute_action/forward?
```
opcode81: This breaks compatibility with persisted policies. Must implement `__setstate__`

tianshou/data/collector.py
```diff
@@ -242,18 +236,168 @@ def reset_buffer(self, keep_statistics: bool = False) -> None:
     def reset_env(
         self,
         gym_reset_kwargs: dict[str, Any] | None = None,
-    ) -> None:
+    ) -> tuple[np.ndarray, np.ndarray]:
```
opcode81, line 239: Return value should be documented in docstring.

tianshou/data/collector.py
```diff
@@ -242,18 +236,168 @@ def reset_buffer(self, keep_statistics: bool = False) -> None:
...
 
+    def close(self) -> None:
+        super().close()
+        self._pre_collect_obs_RO = None
+        self._pre_collect_info_R = None
+
+    def reset_env(
+        self,
+        gym_reset_kwargs: dict[str, Any] | None = None,
+    ) -> tuple[np.ndarray, np.ndarray]:
```
opcode81, line 385: Docstring should be removed for overridden method. Several more occurrences below.

tianshou/data/collector.py
```diff
@@ -242,18 +236,168 @@ def reset_buffer(self, keep_statistics: bool = False) -> None:
...
+    #     For n_step, we always use all envs to collect the data, while for n_episode,
+    #     R will be at most n_episode at the beginning, but can decrease during the collection.
+    # O - dimension(s) of observations
+    # A - dimension(s) of actions
+    # H - dimension(s) of hidden state
+    # D - number of envs that reached done in the current collect iteration. Only relevant in n_episode case.
+    # S - number of surplus envs, i.e. envs that are ready but won't be used in the next iteration.
+    #     Only used in n_episode case. Then, R becomes R-S.
+
+    # set the policy's modules to eval mode (this affects modules like dropout and batchnorm) and the policy
```
MischaPanch: memo for self - remove this comment (no longer true)

opcode81:
Notebooks and documentation have not been updated and still use the old way of setting train/eval mode.

tianshou/data/collector.py
```diff
@@ -242,18 +236,168 @@ def reset_buffer(self, keep_statistics: bool = False) -> None:
...
+
+        :return: The collected stats
+        """
+        # check that exactly one of n_step or n_episode is set and that the other is larger than 0
+        self._validate_n_step_n_episode(n_episode, n_step)
+
+        if reset_before_collect:
+            self.reset(reset_buffer=False, gym_reset_kwargs=gym_reset_kwargs)
+
+        policy_mode_context = in_eval_mode if eval_mode else in_train_mode
```
opcode81: Suggest to add a context manager which accepts a flag, such that this usage pattern is simplified.

opcode81: Actually, Max and I have now concluded that there is really no case where the collector should ever use the policy in train mode. The _only_ reason why it currently needs to be in train mode is that some policies have a `deterministic_eval` flag, which is True by default (e.g. SAC) and which should basically never be active when collecting data for training - BUT which is deactivated based on the policy's "training" status only.

opcode81: The flag `no_grad` is also unnecessary and should be removed from `Collector.collect`.

opcode81: > Actually, Max and I have now concluded that there is really no case where the collector should ever use the policy in train mode. The only reason why it currently needs to be in train mode is that some policies have a deterministic_eval flag, which is True by default (e.g. SAC) and which should basically never be active when collecting data for training.

Max and I have resolved this. See recent commits.

tianshou/trainer/utils.py
```diff
@@ -47,8 +44,8 @@ def gather_info(
     gradient_step: int,
     best_reward: float,
     best_reward_std: float,
-    train_collector: Collector | None = None,
-    test_collector: Collector | None = None,
+    train_collector: BaseCollector | None = None,
+    test_collector: BaseCollector | None = None,
 ) -> InfoStats:
     """A simple wrapper of gathering information from collectors.
```
MischaPanch, line 51: note to self: remove docstring that documents info-stats members

tianshou/policy/base.py
```diff
@@ -511,7 +530,8 @@ def update(
         batch, indices = buffer.sample(sample_size)
         self.updating = True
         batch = self.process_fn(batch, buffer, indices)
-        training_stat = self.learn(batch, **kwargs)
+        with in_train_mode(self):
```
MischaPanch: shouldn't update or rather learn temporary set the `is_within_training_step` to True?

@opcode81 @maxhuettenrauch

MischaPanch: Ah, now I see, it's the correct behavior. We should still adjust it to ensure that it's set to False -with a private contextmanager?

In the current implementation in PGPolicy and so, it has the logic of `within_training_rollout`, where we want to sample from the dist. Here we want to compute gradients, so we are no longer within a rollout.

We are still within a training iteration, so I feel like the name is confusing. How about `within_training_rollout` or `within_training_data_collection`?

tianshou/trainer/base.py
```diff
@@ -426,64 +408,98 @@ def test_step(self) -> tuple[CollectStats, bool]:
 
         return test_stat, stop_fn_flag
 
-    def train_step(self) -> tuple[CollectStats, bool]:
-        """Perform one training step.
+    @contextmanager
+    def _is_training_step_enabled(self, is_training_step: bool):
```
MischaPanch: see my comment above - rename to `_in_training_rollout`?

tianshou/policy/base.py
opcode81: No, `is_within_training_step` has precisely the semantics we want:

image

The term **training step** should become part of a _glossary_ of well-defined terms in Tianshou.
It's already used in, for example, SamplingConfig - and now it is also reflected in the Trainer (method `training_step`).

> We should still adjust it to ensure that it's set to False -with a private contextmanager?

This is very unclear. Adjust what, where, and why?

tianshou/trainer/base.py
opcode81: See comment above.

tianshou/policy/base.py
MischaPanch: After discussion: the resolution is to raise an error if the flag is false when update is called

tianshou/data/collector.py
MischaPanch: I will remove no_grad as part of this PR

tianshou/trainer/utils.py
MischaPanch: Done

tianshou/policy/base.py
MischaPanch: Done

tianshou/data/collector.py
MischaPanch: Done

tianshou/data/collector.py
MischaPanch: Done

tianshou/data/collector.py
MischaPanch: Done

MischaPanch (approved):

MischaPanch:
This is done, will merge without squashing when tests have finished

PR #1126, opened 2024-04-26
Unified build method for HL experiment

maxhuettenrauch:
This is a suggestion for a single public build method that takes `num_experiments` as an argument and returns a  list of `num_experiments` experiments.

### DIFF
--- a/examples/mujoco/mujoco_ppo_hl_multi.py
+++ b/examples/mujoco/mujoco_ppo_hl_multi.py
@@ -13,7 +13,6 @@ These plots are saved in the log directory and displayed in the console.
 """
 
 import os
-import sys
 from collections.abc import Sequence
 from typing import Literal
 
@@ -48,7 +47,7 @@ def main(
     hidden_sizes: Sequence[int] = (64, 64),
     lr: float = 3e-4,
     gamma: float = 0.99,
-    epoch: int = 3,
+    epoch: int = 100,
     step_per_epoch: int = 30000,
     step_per_collect: int = 2048,
     repeat_per_collect: int = 10,
@@ -67,7 +66,7 @@ def main(
     value_clip: bool = False,
     norm_adv: bool = False,
     recompute_adv: bool = True,
-    run_experiments_sequentially: bool = True,
+    run_experiments_sequentially: bool = False,
 ) -> str:
     """Use the high-level API of TianShou to evaluate the PPO algorithm on a MuJoCo environment with multiple seeds for
     a given configuration. The results for each run are stored in separate sub-folders. After the agents are trained,
@@ -128,9 +127,7 @@ def main(
         train_seed=sampling_config.train_seed,
         test_seed=sampling_config.test_seed,
         obs_norm=True,
-        venv_type=VectorEnvType.SUBPROC_SHARED_MEM_FORK_CONTEXT
-        if sys.platform == "darwin"
-        else VectorEnvType.SUBPROC_SHARED_MEM,
+        venv_type=VectorEnvType.SUBPROC_SHARED_MEM_FORK_CONTEXT,
     )
 
     experiments = (
@@ -159,7 +156,7 @@ def main(
         .with_actor_factory_default(hidden_sizes, torch.nn.Tanh, continuous_unbounded=True)
         .with_critic_factory_default(hidden_sizes, torch.nn.Tanh)
         .with_logger_factory(LoggerFactoryDefault("tensorboard"))
-        .build_default_seeded_experiments(num_experiments)
+        .build(num_experiments=num_experiments)
     )
 
     if run_experiments_sequentially:
--- a/tianshou/highlevel/experiment.py
+++ b/tianshou/highlevel/experiment.py
@@ -491,14 +491,26 @@ class ExperimentBuilder:
         else:
             return self._optim_factory
 
-    def build(self, add_seeding_info_to_name: bool = False) -> Experiment:
-        """Creates the experiment based on the options specified via this builder.
+    def build(
+        self,
+        num_experiments: int = 1,
+        add_seeding_info_to_name: bool = False,
+    ) -> list[Experiment]:
+        """Creates num_experiments experiments based on the options specified via this builder.
 
+        :param num_experiments: the number of experiments to create
         :param add_seeding_info_to_name: whether to add a postfix to the experiment name that contains
             info about the training seeds. Useful for creating multiple experiments that only differ
             by seeds.
-        :return: the experiment
+        :return: a list of experiments
         """
+        if num_experiments == 1:
+            return [self._build_experiment(add_seeding_info_to_name)]
+        else:
+            return self._build_default_seeded_experiments(num_experiments)
+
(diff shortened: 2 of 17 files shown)


tianshou/highlevel/experiment.py
```diff
@@ -491,14 +491,26 @@ def _get_optim_factory(self) -> OptimizerFactory:
         else:
             return self._optim_factory
 
-    def build(self, add_seeding_info_to_name: bool = False) -> Experiment:
-        """Creates the experiment based on the options specified via this builder.
+    def build(
+        self,
+        num_experiments: int = 1,
+        add_seeding_info_to_name: bool = False,
```
MischaPanch, line 497: Default should be True, I think. What happens if it's false and n>1, don't we get name collisions of files?

tianshou/highlevel/experiment.py
```diff
@@ -491,14 +491,26 @@ def _get_optim_factory(self) -> OptimizerFactory:
...
+    ) -> list[Experiment]:
```
MischaPanch, line 498: The change in the return is breaking. We could make it backwards compatible by returning `Experiment | list[Experiment]`. While it's somehow less elegant, at the same time `num_experiments=1` is an important special case, and it might be a bit annoying to always have to use `build()[0]` for it.

Alternatively, or even additionally, we can adjust the launcher interface to just call `.run()` in the main process in case a list with a single entry is passed. @opcode81 wdyt?

examples/mujoco/mujoco_ppo_hl_multi.py
```diff
@@ -13,7 +13,6 @@
 """
```
MischaPanch, line 13: This script should be deleted as part of this PR, right?

tianshou/highlevel/experiment.py
maxhuettenrauch: I stumbled upon this as well. Both solutions seemed equally ugly...

tianshou/highlevel/experiment.py
maxhuettenrauch: Probably it's better to default to True. If `num_experiments` is >1, it will be set to True internally anyway.

tianshou/highlevel/experiment.py
MischaPanch: I think making any launcher just run in the main process if a single experiment is passed is a good approach. It would be easy to implement: make a refactoring renaming `launch` to `_launch` in the base class, and then add an actual implementation of the public `launch` in the base class that either calls the abstract `_launch` or calls `experiments[0].run()`

MischaPanch: Doesn't fully resolve the breaking situation (let's wait for @opcode81's view on this), but at least improves a bit. Then one can just always use the launcher

tianshou/highlevel/experiment.py
MischaPanch: Agree :)

examples/mujoco/mujoco_ppo_hl_multi.py
maxhuettenrauch: I thought I'd leave it in there until we decide on a final version. Then we'll have to change all examples to be consistent and make the new functionality visible

MischaPanch: Ok, makes sense. We can remove it once the new functionality is out of "alpha"

tianshou/highlevel/experiment.py
opcode81: This violates the principles of the builder pattern and leads to ugly usage patterns. The `build` method must take no arguments. You are conflating the idea of a factory for multiple experiments with the builder. This is a bad idea, which [defies expectations](https://en.wikipedia.org/wiki/Principle_of_least_astonishment).

Instead, what you want is 
 * an abstraction (specifically, a factory) which creates an experiment for a given seed (e.g. `SeededExperimentFactory` with `create_experiment(seed)` or sth along those lines). The experiment builder usage would then appear in the respective "create" method.
 * an abstraction which implements a particular way of creating multiple experiments from a base random seed, taking a `SeededExperimentFactory` as an argument. This can handle your current way of creating multiple experiments, which is, ultimately, just one (largely arbitrary) way of creating multiple experiments.

This would be the clean way to do it without adding any contrived mechanisms.
In general: Prefer adding new abstractions that build upon the existing abstractions to the addition of contrived mechanisms in existing, well-defined abstractions.

And it's not only `num_experiments`; the argument `add_seeding_info_to_name` is equally problematic. The builder has a method `with_experiment_name` which should be the sole place to handle the naming. It is rather strange to add a second way of influencing the naming - and the only reason it exists is because the builder pattern was, ultimately, abused.

MischaPanch: Let's have a call on how to handle this best. The seeding, the new dir structrure, and the multiple experiments should be the new default. I feel like we might get a bloated interface if we have a builder as well as a factory (the latter also needs to be instantiated/parametrized). I wouldn't want our high-level examples to start with something like 

```python
SeededExperimentFactory.from_experiment_builder(experiment_builder).create_default_seeded_experiments()
```

It's a bit scary and might be off-putting for newcomers. They would need to read quite a bit of documentation of when to use what (builder vs. factory), whereas really we could just offer a single public interface that always creates a bunch of seeded experiments. Sometimes the bunch might be of length one, and that's ok :)

opcode81: Why do you want to change the high-level examples in the first place? The examples are fine the way they are.

What we need is not a change in the existing high-level examples but additional functionality for the case where a user wants to run a series of experiments - and for this case, a new abstraction is required.

opcode81: And it makes no sense to create a factory from a builder; a factory can internally create and apply a builder though.

opcode81: > the multiple experiments should be the new default

Why? Initially, a user will usually want to run only a single experiment and see what happens. Once the user knows which effects he wants to look into, he may want to conduct a series of experiments - but he might want to vary many things. Assuming that he wants to vary the seeds only is a random assumption to me.

maxhuettenrauch: In principle, the idea is to be able to evaluate a given configuration on multiple seeds. The examples are, after all, only examples. People who will properly evaluate algorithms on a task will probably build their own structure on their own configuration stack. Most probably, and despite being an anti-pattern, using for example hydra :)

opcode81:
@maxhuettenrauch this branch's history appears to be broken. Many of its commits seem to already exist on master (rebased/squashed versions thereof) because the diff being displayed on GitHub does not show their changes, and there are many conflicts with branch master.

tianshou/highlevel/experiment.py
maxhuettenrauch: What if, instead of adding a function to build a number of experiments, we only provide some functionality that returns the corresponding experiment/sampling configs? Then it's just a matter of either looping over `build` with updated experiment/sampling configs (or newly generated builders), or prividing a launcher with a set of configs.

opcode81:
Closing. Topic will be revisited in a new PR.

PR #1128, opened 2024-04-29
Support Actor preprocessing network reuse for continuous case, fixes in DQN network

opcode81:
- [X] I have added documentation for my changes

This PR fixes a bug in DQN and lifts a limination in reusing the actor's preprocessing network for continuous environments.

* `atari_network.DQN`:
  * Fix input validation
  * Fix output_dim not being set if features_only=True and output_dim_added_layer not None
* `continuous.Critic`: 
     * Add flag `apply_preprocess_net_to_obs_only` to allow the
        preprocessing network to be applied to the observations only (without
        the actions concatenated), which is essential for the case where we want
        to reuse the actor's preprocessing network
     * CriticFactoryReuseActor: Use the flag, fixing the case where we want to reuse an actor's
      preprocessing network for the critic (must be applied before concatenating
      the actions)
* Minor improvements in docs/docstrings

### DIFF
--- a/tianshou/utils/net/continuous.py
+++ b/tianshou/utils/net/continuous.py
@@ -15,6 +15,7 @@ from tianshou.utils.net.common import (
     TLinearLayer,
     get_output_dim,
 )
+from tianshou.utils.pickle import setstate
 
 SIGMA_MIN = -20
 SIGMA_MAX = 2
@@ -109,6 +110,9 @@ class Critic(CriticBase):
         `preprocess_net`. Only used when `preprocess_net` does not have the attribute `output_dim`.
     :param linear_layer: use this module as linear layer.
     :param flatten_input: whether to flatten input data for the last layer.
+    :param apply_preprocess_net_to_obs_only: whether to apply `preprocess_net` to the observations only (before
+        concatenating with the action) - and without the observations being modified in any way beforehand.
+        This allows the actor's preprocessing network to be reused for the critic.
 
     For advanced usage (how to customize the network), please refer to
     :ref:`build_the_network`.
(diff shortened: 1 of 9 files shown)


tianshou/utils/net/continuous.py
```diff
@@ -137,6 +143,14 @@ def __init__(
             flatten_input=flatten_input,
         )
 
+    def __setstate__(self, state: dict) -> None:
```
MischaPanch, line 146: I don't mind doing that (although I doubt anyone actually persistent this type of net since 1.0 release), but let us remember to remove it at release of 2.0. Could you add a TODO-2.0: remove, so we can find it later?

Also, pls add a description of the fix and api extension to `CHANGELOG.md`

MischaPanch (approved):

tianshou/utils/net/continuous.py
opcode81: > I doubt anyone actually persistent this type of net

Why do you doubt that? It's quite likely actually, as it would affect _everyone_ who persisted a policy that contains a continuous.Critic. Persisting the policy is the default behaviour of the high-level API, so it would be unlikely only if you assume that we have no users. ;)

And even if it was unlikely, we should _never_ knowingly break compatibility if we can avoid it.
You make it sound like it's acceptable to do this ("I don't mind") - when in fact it would be unacceptable not to do it.

> add a TODO-2.0: remove

The TODO does not make sense at this stage, because this class has not yet been changed (incompatibly) for 2.0. Once it is changed incompatibly for 2.0 (if ever), the one changing it should add the TODO (or remove it directly) - not me and not now.

MischaPanch: Ok :)

I'll merge this without squashing in a few hours. If you wanna clean up the history and force push, now's the chance ;)

opcode81: > Also, pls add a description of the fix and api extension to CHANGELOG.md

We should add this to the PR template.

MischaPanch: Yes, was on my list but I forgot. Feel free to do that if you want, otherwise I'll push the extended template to master

opcode81: I added it to the template.

PR #1141, opened 2024-05-07
Bugfix/parallel launcher for linux

### LINKED ISSUE #1121: Revisit `Launcher` for starting multiple experiments
I couldn't make it work on ubuntu (loky fails with pickling error, multiprocessing fails with `daemonic process can't have children`), and it generally is of limited use to start experiments in parallel on a single machine, unless there are a lot of CPUs.

The one thing that might work and actually make sense is a ray launcher. We could also borrow a launcher from hydra, though it might be cumbersome.

An LLM with coding abilities can likely implement a ray based launcher almost out of the box

MischaPanch:
Closes: #1121 

For details about changes see additions to changelog

### DIFF
--- a/tianshou/evaluation/launcher.py
+++ b/tianshou/evaluation/launcher.py
@@ -2,7 +2,7 @@
 
 import logging
 from abc import ABC, abstractmethod
-from collections.abc import Sequence
+from collections.abc import Callable, Sequence
 from copy import copy
 from dataclasses import asdict, dataclass
 from enum import Enum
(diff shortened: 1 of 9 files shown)


tianshou/evaluation/launcher.py
```diff
@@ -26,19 +27,89 @@ class JoblibConfig:
 
 
 class ExpLauncher(ABC):
+    def __init__(
+        self,
+        experiment_runner: Callable[
+            [Experiment],
+            InfoStats | None,
+        ] = lambda exp: exp.run().trainer_result,
```
MischaPanch, line 35: Anticipating a question here: we can't return a list of  `ExperimentResult` since it will contain an instance of `World` which is not picklable - joblib complains. 

It would also be unnecessary to return all the worlds, they can be quite heavy.

So, instead we return the trainer_result object, which however can be None (but usually is not)

maxhuettenrauch (approved):

PR #1168, opened 2024-07-06
Added support for minibatch in PPO process_fn

### LINKED ISSUE #1164: No minibatch for computation of logp_old in PPOPolicy
    + [x] exception-raising bug
    + [x] RL algorithm bug
    + [ ] new feature request
    + [x] design request (i.e. "X should be changed to Y.")

I have noticed that in the implementation of the PPOPolicy, the computation of the old log probabilities `logp_old` is performed without using minibatch:
```python
with torch.no_grad():
   batch.logp_old = self(batch).dist.log_prob(batch.act)
```
This makes this algorithm unusable in situations where the batch is too large, with no possibility of controlling it via batch_size.
I simply suggest to add support for minibatch:
```python
logp_old = []
with torch.no_grad():
    for minibatch in batch.split(self._batch, shuffle=False, merge_last=True):
        logp_old.append(self(minibatch).dist.log_prob(minibatch.act))
    batch.logp_old = torch.cat(logp_old, dim=0).flatten()
```

The version of Tianshou that I'm using is 1.0.0.

jvasso:
Closes #1164

In PPOPolicy, the method `process_fn()` now computes `logp_old` in minibatch instead of all at once.

### DIFF
--- a/tianshou/policy/modelfree/ppo.py
+++ b/tianshou/policy/modelfree/ppo.py
@@ -139,7 +139,7 @@ class PPOPolicy(A2CPolicy[TPPOTrainingStats], Generic[TPPOTrainingStats]):  # ty
         batch.act = to_torch_as(batch.act, batch.v_s)
         logp_old = []
         with torch.no_grad():
-            for minibatch in batch.split(self._batch, shuffle=False, merge_last=True):
+            for minibatch in batch.split(self.max_batchsize, shuffle=False, merge_last=True):
                 logp_old.append(self(minibatch).dist.log_prob(minibatch.act))
             batch.logp_old = torch.cat(logp_old, dim=0).flatten()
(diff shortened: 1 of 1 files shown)


Trinkle23897 (approved):

MischaPanch:
This looks innocent, but tests fail. I'll have a look at what goes wrong and will finalize this.

Sorry for the late reaction, my official work has moved me in another direction for a few months, but now I'm back to devoting time to tianshou :)

MischaPanch:
Looks like just a typo, `_batch` -> `batch`

MischaPanch:
Ok, I had a closer look. The current handling is a bit confusing, there is `self.max_batchsize` which is used implicitly through a method defined in super class. So the fix is to split according to this param, pushing the fix now

Issue #1169, opened 2024-07-08
Improve interface of BasePolicy.compute_action


opcode81:
- [X] I have marked all applicable categories:
    + [ ] exception-raising bug
    + [ ] RL algorithm bug
    + [ ] documentation request (i.e. "X is missing from the documentation.")
    + [ ] new feature request
    + [X] design request (i.e. "X should be changed to Y.")
- [ ] I have visited the [source website](https://github.com/thu-ml/tianshou/)
- [ ] I have searched through the [issue tracker](https://github.com/thu-ml/tianshou/issues) for duplicates
- [ ] I have mentioned version numbers, operating system and environment, where applicable:
  ```python
  import tianshou, gymnasium as gym, torch, numpy, sys
  print(tianshou.__version__, gym.__version__, torch.__version__, numpy.__version__, sys.version, sys.platform)
  ```

The interface is not optimal, because not all observations are indeed arrays. Observations can also be array-like only (numpy's `ArrayLike` type). In fact, this is the case for some of gymnasium's standard implementations (e.g. `LazyFrames`).

I am not sure how appropriate the current type annotation is (i.e. `arr_type`, which also includes `Tensor`).

Proposal:
* Change type annotation of `obs` to numpy's `ArrayLike`
* Add first line which ensures array representation: `obs = np.array(obs)`. This will convert anything array-like to an actual array (e.g. using an implementation of `__array__` if necessary).

MischaPanch:
Closed by #1170

PR #1170, opened 2024-07-08
Improve interface of BasePolicy.compute_action

opcode81:
- [X] I have added documentation for my changes and have listed relevant changes in CHANGELOG.md
- [X] If applicable, I have added tests to cover my changes.

Implements #1169

### DIFF
--- a/tianshou/policy/base.py
+++ b/tianshou/policy/base.py
@@ -10,6 +10,7 @@ import numpy as np
 import torch
 from gymnasium.spaces import Box, Discrete, MultiBinary, MultiDiscrete
 from numba import njit
+from numpy.typing import ArrayLike
 from overrides import override
 from torch import nn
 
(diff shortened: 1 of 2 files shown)


tianshou/policy/base.py
```diff
@@ -300,8 +301,8 @@ def compute_action(
         :param state: the hidden state of RNN policy, used for recurrent policy.
         :return: action as int (for discrete env's) or array (for continuous ones).
         """
-        # need to add empty batch dimension
-        obs = obs[None, :]
+        obs = np.array(obs)  # convert array-like to array (e.g. LazyFrames)
+        obs = obs[None, :]  # add batch dimension
```
Trinkle23897, line 305: I think the original intention here I made is to allow any Batch-like obs in this code path, so `batch[None, :]` will add batch-size dimension to all nested tree elements

opcode81: @MischaPanch added the `compute_action` method to have a convenient interface for the inference case (i.e. when actually applaying the agent in practice), so we need a function that can handle the observations the way they are produced by the environment. The operation in line 305, however, will fail for things like `LazyFrames` if we don't ensure that array-like representations are converted to actual arrays first.

MischaPanch: Indeed, it's currently unused. Examples will follow soon in updated tutorials.

Thanks @opcode81 ! Merging this

MischaPanch (approved):

PR #1181, opened 2024-07-29
Feature/improvements batch

MischaPanch:
This PR contains several important extensions and improvements of Batch.

1. A method for setting a new sequence. This is a first step towards using code that's less reliant on setting attributes directly. The method also permits setting a subsequence and filling up with default values. This will help ensuring that all entries in a batch are of the same length, something that we should start enforcing soon.
2. A new method for applying arbitrary transformations to "leaves" in a Batch. This simplifies already existing code and is generally very handy for users
3. The arbitrary transformations allowed implementing `isnull`, `hasnull` and `dropnull`, which helps finding errors early.
4. The arbitrary transformations now also allow extracting a `schema` from a batch. This was used in `Batch.cat_` to perform additional input validation (we now make sure there that the structures are the same when concatenating batches). This input validation is a **breaking change**! Some tests that concatenated incompatible batches were removed. Eventually, we can add a `get_schema` method to the batch that will retrieve metainfo like shapes and datatypes. For now, this is delegated to the user who can use the new `apply_values_transform`
5. New feature: slicing of torch distributions. Previously several batches contained instances of `Distribution` which were not properly sliced, inviting bugs and errors in user code. This is now fixed - albeit not in a pretty way (torch doesn't allow slicing the objects natively)
6. Stricter typing and some further minor extensions and simplifications

The new code was extensively tested and documented.

### DIFF
--- a/tianshou/data/batch.py
+++ b/tianshou/data/batch.py
@@ -6,6 +6,7 @@ from numbers import Number
 from types import EllipsisType
 from typing import (
     Any,
+    Literal,
     Protocol,
     Self,
     TypeVar,
@@ -429,7 +430,7 @@ class BatchProtocol(Protocol):
         self,
         seq: np.ndarray,
         key: str,
-        index: Sequence[int] | None = None,
+        index: IndexType | None = None,
         default_value: float | None = None,
     ) -> None:
         """Set a sequence of values at a given key.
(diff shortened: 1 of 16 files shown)


MischaPanch:
@dantp-ai Pls have a look if you have time, you also worked a lot with Batch in the past

MischaPanch:
When review is done I'd extend the PR description and add a changelog entry. With this and buffer improvements merged, I think we should release version 1.1

Trinkle23897 (approved):
quick skim, lgtm; haven't looked into detail

tianshou/data/batch.py
```diff
@@ -967,3 +1135,151 @@ def split(
...
+            if len(arr) != len(self):
+                raise ValueError(
+                    f"Sequence length {len(arr)} does not match "
+                    f"batch length {len(self)}. For setting a subsequence with missing "
+                    f"entries filled up by default values, consider passing an index.",
+                )
+            self[key] = arr
+
+    def isnull(self) -> Self:
+        return self.apply_values_transform(pd.isnull, inplace=False)
```
Trinkle23897, line 1220: why do you need pandas to check isnull? is there a better way, e.g., using numpy?

MischaPanch: Numpy fails with certain types of NaNs (maybe None in arrays of certain types), tests failed. Pandas is more robust, that's why I used it. I'll add a block comment

MischaPanch:
Thanks for the review @Trinkle23897 !

Some more fixes in tests were needed, now CI should run through. I'd merge then

MischaPanch:
I'll merge now and will update the changelog later

PR #1183, opened 2024-08-01
Enhanced rliable eval

MischaPanch:
Rliable eval: multiple extensions
  
  1. Support for evaluating training runs (previously just test runs were supported)
  2. Improved handling of figures and axes
  3. Allow passing max_env_step
  4. Use min len of all experiments (bugfix, previously it would crash if experiments had different lengths)
  5. Extended the data loading logic such that it can handle data from low-level API experiments

Also some fixes and extensions in loggers

### DIFF
--- a/tianshou/evaluation/rliable_evaluation_hl.py
+++ b/tianshou/evaluation/rliable_evaluation_hl.py
@@ -3,7 +3,8 @@ on different seeds using the rliable library. The API is experimental and subjec
 """
 
 import os
-from dataclasses import asdict, dataclass, fields
+from dataclasses import dataclass, fields
+from typing import Literal
 
 import matplotlib.pyplot as plt
 import numpy as np
@@ -12,7 +13,7 @@ from rliable import library as rly
 from rliable import plot_utils
 
 from tianshou.highlevel.experiment import Experiment
-from tianshou.utils import logging
+from tianshou.utils import TensorboardLogger, logging
 from tianshou.utils.logger.base import DataScope
 
 log = logging.getLogger(__name__)
@@ -61,17 +62,31 @@ class RLiableExperimentResult:
     test_episode_returns_RE: np.ndarray
     """The test episodes for each run of the experiment where each row corresponds to one run."""
 
+    train_episode_returns_RE: np.ndarray
+    """The training episodes for each run of the experiment where each row corresponds to one run."""
+
     env_steps_E: np.ndarray
     """The number of environment steps at which the test episodes were evaluated."""
 
+    env_steps_train_E: np.ndarray
+    """The number of environment steps at which the training episodes were evaluated."""
+
     @classmethod
-    def load_from_disk(cls, exp_dir: str) -> "RLiableExperimentResult":
+    def load_from_disk(
+        cls,
+        exp_dir: str,
+        max_env_step: int | None = None,
+    ) -> "RLiableExperimentResult":
         """Load the experiment result from disk.
 
         :param exp_dir: The directory from where the experiment results are restored.
+        :param max_env_step: The maximum number of environment steps to consider. If None, all data is considered.
+            Note: if the experiments have different numbers of steps, the minimum number is used.
         """
         test_episode_returns = []
+        train_episode_returns = []
         env_step_at_test = None
+        env_step_at_train = None
 
         # TODO: env_step_at_test should not be defined in a loop and overwritten at each iteration
         #  just for retrieving them. We might need a cleaner directory structure.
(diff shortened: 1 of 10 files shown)


tianshou/highlevel/logger.py
```diff
@@ -74,3 +74,36 @@ def create_logger(
                 return TensorboardLogger(writer)
             case _:
                 raise ValueError(f"Unknown logger type '{self.logger_type}'")
+
+
+class WandbLoggerFactory(LoggerFactory):
```
maxhuettenrauch: This is currently not doing much more than the wandblogger from the LoggerFactoryDefault, right? I.e., extended run grouping with wandb groups and jobtypes is left for another PR?

tianshou/utils/logger/wandb.py
```diff
@@ -11,6 +12,8 @@
 with contextlib.suppress(ImportError):
     import wandb
 
+log = logging.getLogger(__name__)
+
 
 class WandbLogger(BaseLogger):
```
maxhuettenrauch, line 18: Will need extra arguments for run grouping

tianshou/evaluation/rliable_evaluation_hl.py
```diff
@@ -61,61 +61,140 @@ class RLiableExperimentResult:
...
+                test_episode_returns.append(test_data.returns_stat.mean)
+                env_step_at_test = test_data.env_step
+
+            if train_data.returns_stat is not None:
+                train_episode_returns.append(train_data.returns_stat.mean)
+                env_step_at_train = train_data.env_step
 
+        test_data_found = True
+        train_data_found = True
         if not test_episode_returns or env_step_at_test is None:
```
maxhuettenrauch, line 159: We probably need another mechanism here as well. In the unfortunate event of all but the last run having recorded data, `env_step_at_test` may be None and the RuntimeError below is raised.

tianshou/evaluation/rliable_evaluation_hl.py
```diff
@@ -216,3 +320,30 @@ def eval_results(
...
+
+    :param log_dir: The directory containing the experiment results.
+    :param show_plots: whether to display plots.
+    :param save_plots: whether to save plots to the `log_dir`.
+    :param scope: The scope of the evaluation, either 'TEST' or 'TRAIN'.
+    :param max_env_step: The maximum number of environment steps to consider. If None, all data is considered.
+            Note: if the experiments have different numbers of steps, the minimum number is used.
+    """
+    rliable_result = RLiableExperimentResult.load_from_disk(log_dir, max_env_step=max_env_step)
+    if scope == "both":
```
maxhuettenrauch, line 350: scope `both` currently doesn't exist in `DataScope`

maxhuettenrauch (commented):
Changes lgtm, I've just added some comments in case we wanted to expand. Otherwise I'll approve

tianshou/highlevel/logger.py
MischaPanch: I haven't looked in detail, this part was copied over from our internal branch. I supposed that you had written it. Feel free to push corrections regarding the Wandb logger 

I had to adjust the logger code to make the restoration method static

tianshou/evaluation/rliable_evaluation_hl.py
MischaPanch: Ah, it was supposed to be a union type, good catch! A shame mypy doesn't see it

tianshou/utils/logger/wandb.py
MischaPanch: See above

tianshou/evaluation/rliable_evaluation_hl.py
MischaPanch: I feel like we should rather improve the logging and stats-saving mechanism to ensure this never happens instead of doing more gymnastics here. Would that be possible? We can do that in a separate issue/PR.

tianshou/evaluation/rliable_evaluation_hl.py
MischaPanch: Fixed

tianshou/highlevel/logger.py
maxhuettenrauch: Should I try to merge the additional WandbLoggerFactory with the existing one in LoggerFactoryDefault, or can there be any benefit in having a "simple" wandblogger and a more advanced one?

MischaPanch: I don't think we need two wandbloggers, so feel free to integrate the extension

maxhuettenrauch: I merged the two and also found some other things. Also I found out that parallel execution and wandb logging currently don't go well together.

tianshou/utils/logger/base.py
```diff
@@ -172,6 +176,9 @@ def prepare_dict_for_logging(
     def write(self, step_type: str, step: int, data: dict[str, VALID_LOG_VALS_TYPE]) -> None:
         """The LazyLogger writes nothing."""
 
+    def finalize(self) -> None:
+        pass
+
```
maxhuettenrauch, line 178: Do we need the extra implementation? Now mypy is complaining about a duplicate. I first tried to have an empty non abstract method in the base class, but then ruff was complaining..

MischaPanch: I removed the ruff rule that caused the complaining and now also removed the abstractmethod

MischaPanch: Ah, well, since it now actually is implemented in all loggers, the non-abstract implementation should be removed from BaseLogger. I'll do that now

MischaPanch:
@maxhuettenrauch Thanks for the logger extension and review! Merging this now

Issue #1090, opened 2024-04-03
Buffer: fix discrepancy in slicing order


MischaPanch:
It doesn't make sense for `buffer[:].obs` to be so wildly different from `buffer.obs[:]`. One is retrieving the full buffer, filled up with zeros, the other just the filled entries.

We should probably never retrieve the full buffer through slicing. There could be a new method like `get_full_entry_sequence` that is documented to retrieve the non-filled entries as well if it's actually needed.

While addressing this issue, buffer and collector related tests should be adjusted and improved.

dantp-ai:
I can look into this.

Currently this is the signature of `ReplayBuffer`'s `__getitem__(index: slice | int | list[int] | np.ndarray)`.

So the idea here is to not support full-buffer (filled and non-filled entries) retrieval, when slice is provided, for `buffer[:].obs` because it can be computationally expensive? 

What about the cases when user gives a `list[int]` or `np.ndarray` with all the indices? Practically it is the same as slice. Should we allow retrieving the full buffer in this case?

MischaPanch:
I'm not sure when we should ever be retrieving the full buffer. Tbh I haven't given too much thought about the best way of resolving this, it just seems very confusing and arbitrary that `buffer.obs[:]` and `buffer[:].obs` would be semantically wildly different entities, right?

dantp-ai:
I have reviewed this issue again. The different way of retrieving with slicing seems arbitrary to me. The user can check the maximum size via the `buffer.max_size` attribute, no need to return the empty values as well. A special method to retrieve the full buffer (if anyone ever needs it!?) as you mentioned is more appropriate.

We should probably update `buffer.obs[:]` (and other slicing methods like `start:stop`) to retrieve only non-empty values like `buffer[:].obs` does.

Here are some examples of how indexing is currently different between the two:

```
In [52]: dummy_buf = ReplayBuffer(size=10)
    ...: for i in range(6):
    ...:     dummy_buf.add(
    ...:         Batch(obs=i, act=i, rew=i, terminated=0, truncated=0, done=0, obs_next=i + 1, info={}),
    ...:     )
    ...: 

In [53]: dummy_buf.obs[2:8]
Out[53]: array([2, 3, 4, 5, 0, 0])

In [54]: dummy_buf[2:8].obs
Out[54]: array([2, 3, 4, 5])

In [55]: dummy_buf.obs[:]
[...]

PR #1186, opened 2024-08-02
Bugfix/batch eq for scalar

### LINKED ISSUE #1182: Make Batch.__eq__ work also for scalar values (0-dimensional arrays)
As first noticed here https://github.com/thu-ml/tianshou/issues/1086#issuecomment-2157976997, DeepDiff which is currently used for batch equality cannot handle 0-dimensional arrays.

My current proposal is to leverage `np.atleast_1d` and the newest Batch API's `Batch.apply_values_transform` to convert any 0-dimensional arrays to 1-d while keeping the others as they are (cf. https://github.com/thu-ml/tianshou/issues/1086#issuecomment-2263079148).

dantp-ai:
Fixes: https://github.com/thu-ml/tianshou/issues/1182

Note: Updated `test_batch.test_slice_distribution()` to use allclose (See: https://github.com/thu-ml/tianshou/pull/1181).

### DIFF
--- a/test/base/test_batch.py
+++ b/test/base/test_batch.py
@@ -545,6 +545,28 @@ class TestBatchEquality:
         assert batch1 != batch2
 
     @staticmethod
+    def test_array_scalars() -> None:
+        batch1 = Batch(a={"b": 1})
+        batch2 = Batch(a={"b": 1})
+        assert batch1 == batch2
+
+        batch3 = Batch(a={"c": 2})
+        assert batch1 != batch3
+
+        batch4 = Batch(b={"b": 1})
+        assert batch1 != batch4
+
+        batch5 = Batch(a={"b": 10})
+        assert batch1 != batch5
+
+        batch6 = Batch(a={"b": [1]})
+        assert batch1 == batch6
+
+        batch7 = Batch(a=1, b=5)
+        batch8 = Batch(a=1, b=5)
+        assert batch7 == batch8
+
+    @staticmethod
     def test_slice_equal() -> None:
         batch1 = Batch(a=[1, 2, 3])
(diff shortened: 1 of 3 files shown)


test/base/test_batch.py
```diff
@@ -837,10 +859,11 @@ def test_slice_distribution() -> None:
         selected_idx = [1, 3]
         sliced_batch = batch[selected_idx]
         sliced_probs = cat_probs[selected_idx]
-        assert (sliced_batch.dist.probs == Categorical(probs=sliced_probs).probs).all()
-        assert (
-            Categorical(probs=sliced_probs).probs == get_sliced_dist(dist, selected_idx).probs
-        ).all()
+        assert torch.allclose(sliced_batch.dist.probs, Categorical(probs=sliced_probs).probs)
```
MischaPanch, line 862: Thx! I guess it was failing randomly, right?

MischaPanch (approved):

MischaPanch:
Thx for the PR! All looks good, merging now

test/base/test_batch.py
dantp-ai: Yes, it was sporadically returning inequality. I will take a look at __getitem__ soon as I have an overdue task there with maintaining [slicing consistency](https://github.com/thu-ml/tianshou/issues/1090). Then I can also check what causes slicing to make these little decimal errors.

MischaPanch: Thanks, appreciate it! I don't think the getitem issue can actually be solved, but implementing eq already solves all practical problems of it. So no need to further address it, unless you see some simple way for doing that. 

The order of slicing however it's still a valid issue

PR #1187, opened 2024-08-05
Lower level access to hl world

MischaPanch:
Gives the user the possibility to extract the `World` from a configured `Experiment`. This required moving the initial rollout (according to `start_timesteps`) from the `AgentFactory` to `Experiment.run`. But since it was only ever used there, this change is essentially backwards compatible

I also enhanced documentation. `__init__` docstrings were previously not rendered and there were some formatting errors in them, hence the changes there.

### DIFF
--- a/tianshou/highlevel/experiment.py
+++ b/tianshou/highlevel/experiment.py
@@ -25,7 +25,7 @@ from contextlib import suppress
 from copy import deepcopy
 from dataclasses import asdict, dataclass
 from pprint import pformat
-from typing import TYPE_CHECKING, Literal, Self, Union, Unpack
+from typing import TYPE_CHECKING, Any, Self, Union, cast
 
 if TYPE_CHECKING:
     from tianshou.evaluation.launcher import ExpLauncher, RegisteredExpLauncher
@@ -75,10 +75,8 @@ from tianshou.highlevel.module.critic import (
 from tianshou.highlevel.module.intermediate import IntermediateModuleFactory
 from tianshou.highlevel.module.special import ImplicitQuantileNetworkFactory
 from tianshou.highlevel.optim import (
-    DEFAULT_OPTIM_FACTORY_PARAMS,
-    DefaultOptimFactoryParams,
     OptimizerFactory,
-    OptimizerFactoryDefault,
+    OptimizerFactoryAdam,
 )
 from tianshou.highlevel.params.policy_params import (
     A2CParams,
@@ -245,7 +243,7 @@ class Experiment(ToStringMixin):
 
     def create_experiment_world(
         self,
-        override_experiment_name: str | Literal["DATETIME_TAG"] | None = None,
+        override_experiment_name: str | None = None,
         logger_run_id: str | None = None,
         raise_error_on_dirname_collision: bool = True,
         reset_collectors: bool = True,
@@ -258,8 +256,11 @@ class Experiment(ToStringMixin):
         is that some configuration or custom logic should happen before the training loop starts, but one
         still wants to use the convenience of high-level interfaces for setting up the experiment.
 
-        :param override_experiment_name: whether to override the experiment name in the resulting `World`.
-            Passing `DATETIME_TAG` will use a name containing the current date and time.
+        :param override_experiment_name: pass to override the experiment name in the resulting `World`.
+            Affects the name of the persistence directory and logger configuration. If None, the experiment's
+            name will be used.
+            The name may contain path separators (i.e. `os.path.sep`, as used by `os.path.join`), in which case
+            a nested directory structure will be created.
         :param logger_run_id: Run identifier to use for logger initialization/resumption.
         :param raise_error_on_dirname_collision: whether to raise an error on collisions when creating the
             persistence directory. Only takes effect if persistence is enabled. Set to `False` e.g., when continuing
@@ -269,13 +270,13 @@ class Experiment(ToStringMixin):
             or for adding custom logic before training starts.
         """
         if override_experiment_name is not None:
-            if override_experiment_name == "DATETIME_TAG":
-                override_experiment_name = datetime_tag()
-            self.name = override_experiment_name
+            exp_name = override_experiment_name
+        else:
+            exp_name = self.name
 
         # initialize persistence directory
         use_persistence = self.config.persistence_enabled
-        persistence_dir = os.path.join(self.config.persistence_base_dir, self.name)
+        persistence_dir = os.path.join(self.config.persistence_base_dir, exp_name)
         if use_persistence:
             os.makedirs(persistence_dir, exist_ok=not raise_error_on_dirname_collision)
 
@@ -284,7 +285,7 @@ class Experiment(ToStringMixin):
             enabled=use_persistence and self.config.log_file_enabled,
         ):
             # log initial information
-            log.info(f"Running experiment (name='{self.name}'):\n{self.pprints()}")
+            log.info(f"Preparing experiment world (name='{exp_name}'):\n{self.pprints()}")
             log.info(f"Working directory: {os.getcwd()}")
 
             self._set_seed()
@@ -320,7 +321,7 @@ class Experiment(ToStringMixin):
             if use_persistence:
                 logger = self.logger_factory.create_logger(
                     log_dir=persistence_dir,
-                    experiment_name=self.name,
+                    experiment_name=exp_name,
                     run_id=logger_run_id,
                     config_dict=full_config,
                 )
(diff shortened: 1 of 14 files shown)


tianshou/highlevel/experiment.py
```diff
@@ -204,43 +242,39 @@ def save(self, directory: str) -> None:
...
         """
-        # backward compatibility
-        _experiment_name = kwargs.pop("experiment_name", None)
-        if _experiment_name is not None:
-            run_name = cast(str, _experiment_name)
-            deprecation(
-                "Parameter run_name should now be used instead of experiment_name. "
-                "Support for experiment_name will be removed in the future.",
-            )
-        assert len(kwargs) == 0, f"Received unexpected arguments: {kwargs}"
```
opcode81, line 236: Why was backward compatibility for parameter `experiment_name` of method `run` dropped?
Surely we want to remain compatible with v1.0.0, don't we?

opcode81: This change basically undoes changes/improvements from commit 35779696ee860684d1e0dfa229cc8f367ef5118f.

See `git difftool 35779696~1 35779696 tianshou/highlevel/experiment.py`.

tianshou/highlevel/experiment.py
```diff
@@ -249,7 +283,7 @@ def run(
             enabled=use_persistence and self.config.log_file_enabled,
         ):
             # log initial information
-            log.info(f"Running experiment (name='{run_name}'):\n{self.pprints()}")
+            log.info(f"Running experiment (name='{self.name}'):\n{self.pprints()}")
```
opcode81: This, too, is a regression.
`run_name` should be maintained (see above and below).
The aforementioned commit's changes must not be eradicated.

tianshou/highlevel/experiment.py
```diff
@@ -372,19 +463,29 @@ def run(
...
-        if sampling_config is None:
-            sampling_config = SamplingConfig()
-        self._config = experiment_config
+        """:param env_factory: controls how environments are to be created.
+        :param experiment_config: the configuration for the experiment. If None, will use the default values
+            of `ExperimentConfig`.
+        :param sampling_config: the sampling configuration to use. If None, will use the default values
+            of `SamplingConfig`.
+        """
+        self._config = experiment_config or ExperimentConfig()
```
opcode81: Not a big fan of this pattern because it relies on implicit Boolean coercion. 
Why did you change my perfectly fine implementation, which was more explicit?

tianshou/highlevel/experiment.py
```diff
@@ -443,18 +544,15 @@ def with_optim_factory(self, optim_factory: OptimizerFactory) -> Self:
...
-        """Configures the use of the default optimizer, Adam, with the given parameters.
+        """Configures the use of the default optimizer, with the given parameters.
 
-        :param betas: coefficients used for computing running averages of gradient and its square
-        :param eps: term added to the denominator to improve numerical stability
-        :param weight_decay: weight decay (L2 penalty)
+        :param kwargs: the parameters to use for the optimizer, see `DefaultOptimFactoryParams`.
         :return: the builder
         """
-        self._optim_factory = OptimizerFactoryAdam(betas=betas, eps=eps, weight_decay=weight_decay)
```
opcode81, line 457: How is this an improvement?
It seems to me that this just creates confusion as to what the default optimiser and its arguments are.

tianshou/highlevel/experiment.py
```diff
@@ -315,23 +355,74 @@ def run(
...
+        persistence_dir = world.persist_directory
+        use_persistence = self.config.persistence_enabled
+
+        with logging.FileLoggerContext(
+            os.path.join(persistence_dir, self.LOG_FILENAME),
+            enabled=use_persistence and self.config.log_file_enabled,
+        ):
+            trainer_result: InfoStats | None = None
+            if self.config.train:
+                # prefilling buffers with random actions
```
opcode81: The comment is inaccurate; the actions are not necessarily random.
Also, break before comment recommended for readability.

tianshou/highlevel/experiment.py
opcode81: How is this an improvement?
It seems to me that this just creates confusion as to what the default optimiser and its arguments are.

tianshou/highlevel/experiment.py
MischaPanch: Some copy-paste changes from our internal repo that wasn't synced with upstream snuck in, gonna undo them

tianshou/highlevel/experiment.py
MischaPanch: I'll change back, it's just an aesthetic preference. 

NB: Personally, I find this reads very nicely, and truthyness is ubiquitous in the python world. But it's not worth an argument :)

tianshou/highlevel/experiment.py
MischaPanch: It's an improvement b/c the default values are no longer duplicated. Previously, it would have been very easy to change the default values for Adam in the `optim` package without realizing that the default values in the experiment builder would also need to be changed.

Note that it's not mere aesthetics - `_get_optim_factory` calls Adam with default params from `optim`, whereas `with_optim_factory_default` would call it with values defined there

tianshou/highlevel/experiment.py
MischaPanch: See below, was a copy-paste error. I didn't put enough attention on the deleted lines in the diff, thx for spotting!

tianshou/highlevel/experiment.py
MischaPanch: I restored backwards compat in the `run` method. 

Here I don't think it's a regression, run will always pass `override_experiment_name` to this method, which will then set it to the `name` attr.

tianshou/highlevel/experiment.py
MischaPanch: Done

tianshou/highlevel/experiment.py
MischaPanch: Done

tianshou/highlevel/experiment.py
MischaPanch: Done

MischaPanch: `poe format` removes the break.. :(

Maybe we can adjust the setting for this somewhere if you think it's important

tianshou/highlevel/experiment.py
opcode81: It is a regression. The idea of the changes that were made in April is that we differentiate between an experiment (which can be run multiple times) and its runs.
An experiment has a fixed name, but the name used for a run can differ from the experiment's name.

What you have done is changed this notion and again introduced the notion of changing an experiment's name.

It was a deliberate decision (by Max and me)
* to rename parameter `override_experiment_name` to `run_name`
* to remove the option to pass "DATETIME_TAG" (because it's not useful; we introduced other mechanisms to handle naming)

Therefore, the parameter `override_experiment_name` of method `create_experiment_world` should also be changed to `run_name` and it should not be used to change the experiment's name but only as the name for the current execution.

MischaPanch: I think you're misinterpeting the diff, this is no longer the `run` method but the `create_experiment_world` method. The param cannot be called `run_name`, and if I'm not mistaken the functionality of  `run` has not changed. If I'm wrong, let's have a short call to clear this up.

MischaPanch: Ah, wait, I think I misread the diff or copied something wrongly from there. Now I see what you mean. 

It can't be called `run_name` here since this is not a run, but I will change the modification of the name attr that I re-introduced

tianshou/highlevel/experiment.py
opcode81: The argument is not about aethetic preferences but the fact that you changed code that didn't have a problem.
When you write new code, you can, of course, do whatever you like best. But why do you change code that is perfectly fine into code that isn't obviously better - and where one could even argue that it's worse?

One could argue that it's worse because
* explicit is better than implicit (as a general rule)
* inexperienced programmers may not understand it
* it handles two logically separate things in a single line (assignment and handling of the None case)

One could argue that it's better because
* it's shorter

tianshou/highlevel/experiment.py
opcode81: You are solving problems that don't actually exist: The parameters of Adam only change once in a blue moon - and when they do, every responsible programmer will check the usages of `OptimizerFactoryAdam` and update them.

But you have created a new problem - a usability problem. Users get no information on what the default optimizer is or what its parameters are or what they mean. 

Usability is far more important. Why are you optimizing duplication of parameters that literally never change over usability?

tianshou/highlevel/experiment.py
MischaPanch: Done

tianshou/highlevel/experiment.py
MischaPanch: Dropped the commit and instead added a few block comments to keep the values in sync

MischaPanch:
Was reviewed in pair programming and in conversation above, merging

PR #1189, opened 2024-08-06
add highlevel to python package.

### LINKED ISSUE #1188: ModuleNotFoundError: No module named 'tianshou.highlevel'
I got ``ModuleNotFoundError`` when install by ``pip  install tianshou``
> from tianshou.highlevel.logger import LoggerFactoryDefault
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'tianshou.highlevel'

Why not add ``highlevel`` module to ``tianshou/__init__.py``?

gitabtion:
fix [issue 1188](https://github.com/thu-ml/tianshou/issues/1188)

### DIFF
--- a/tianshou/__init__.py
+++ b/tianshou/__init__.py
@@ -9,4 +9,5 @@ __all__ = [
     "policy",
     "trainer",
     "exploration",
+    "highlevel",
 ]


tianshou/__init__.py
```diff
@@ -9,4 +9,5 @@
     "policy",
     "trainer",
     "exploration",
+    "highlevel",
```
MischaPanch, line 12: highlevel is not imported here, so this change doesn't affect functionality. It's not clear to me what caused the issue you refer to

opcode81: The issue is typically caused by someone not using the right version of tianshou.
This import is not required.

Issue #1194, opened 2024-08-07
Actor and distribution function should not be specified independently


opcode81:
Currently, the distribution function (e.g. for PPO) and the actor are specified independently, yet there is obviously a very strong connection: The nature of the actor's output determines the distribution function to apply.

This can cause errors that are hard to debug, especially when the input to the distribution function can be interpreted but its semantics are skewed.

Therefore, I propose to create a stronger link, which we can already implement at least in the high-level API: 
The ActorFactory, which knows the nature of the actor's output, shall also create the matching distribution function, eliminating the issue. In turn, the `dist_fn` parameter can be entirely removed from the high-level API (breaking change).

MischaPanch:
Closed by #1195

PR #1195, opened 2024-08-07
High-level API: Establish a strong link between the actor and the distribution function

opcode81:
Establish a strong link between the actor and the distribution function (dist_fn) used in policies by creating the distribution function in the actor factory which knows which function is appropriate.

Consequently, remove the policy parameter 'dist_fn' from the high-level API because it is determined automatically, eliminating the possibility of misspecification by the user. [breaking change: code must not specify the 'dist_fn' parameter, but persisted objects continue to work as expected]

Implements #1194

- [X] I have added documentation for my changes and have listed relevant changes in CHANGELOG.md
- [X] If applicable, I have added tests to cover my changes.

### DIFF
--- a/tianshou/highlevel/module/actor.py
+++ b/tianshou/highlevel/module/actor.py
@@ -19,6 +19,11 @@ from tianshou.highlevel.module.intermediate import (
 )
 from tianshou.highlevel.module.module_opt import ModuleOpt
 from tianshou.highlevel.optim import OptimizerFactory
+from tianshou.highlevel.params.dist_fn import (
+    DistributionFunctionFactoryCategorical,
+    DistributionFunctionFactoryIndependentGaussians,
+)
+from tianshou.policy.modelfree.pg import TDistFnDiscrOrCont
 from tianshou.utils.net import continuous, discrete
 from tianshou.utils.net.common import BaseActor, ModuleType, Net
 from tianshou.utils.string import ToStringMixin
@@ -47,6 +52,14 @@ class ActorFactory(ModuleFactory, ToStringMixin, ABC):
     def create_module(self, envs: Environments, device: TDevice) -> BaseActor | nn.Module:
         pass
 
+    @abstractmethod
+    def create_dist_fn(self, envs: Environments) -> TDistFnDiscrOrCont | None:
+        """
+        :param envs: the environments
+        :return: the distribution function, which converts the actor's output into a distribution, or None
+            if the actor does not output distribution parameters
+        """
+
     def create_module_opt(
         self,
         envs: Environments,
@@ -70,7 +83,7 @@ class ActorFactory(ModuleFactory, ToStringMixin, ABC):
     def _init_linear(actor: torch.nn.Module) -> None:
         """Initializes linear layers of an actor module using default mechanisms.
 
-        :param module: the actor module.
+        :param actor: the actor module.
         """
         init_linear_orthogonal(actor)
         if hasattr(actor, "mu"):
@@ -104,7 +117,7 @@ class ActorFactoryDefault(ActorFactory):
         self.hidden_activation = hidden_activation
         self.discrete_softmax = discrete_softmax
 
-    def create_module(self, envs: Environments, device: TDevice) -> BaseActor:
+    def _create_factory(self, envs: Environments) -> ActorFactory:
         env_type = envs.get_type()
         factory: ActorFactoryContinuousDeterministicNet | ActorFactoryContinuousGaussianNet | ActorFactoryDiscreteNet
         if env_type == EnvType.CONTINUOUS:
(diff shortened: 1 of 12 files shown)


tianshou/highlevel/module/actor.py
```diff
@@ -159,6 +179,9 @@ def create_module(self, envs: Environments, device: TDevice) -> BaseActor:
             device=device,
         ).to(device)
 
+    def create_dist_fn(self, envs: Environments) -> TDistFnDiscrOrCont | None:
```
MischaPanch, line 182: Return type should be `None`

tianshou/highlevel/module/actor.py
```diff
@@ -202,6 +225,9 @@ def create_module(self, envs: Environments, device: TDevice) -> BaseActor:
 
         return actor
 
+    def create_dist_fn(self, envs: Environments) -> TDistFnDiscrOrCont | None:
```
MischaPanch, line 228: Return type can be fixed to `TDistFnCont`

tianshou/highlevel/params/dist_fn.py
```diff
@@ -20,32 +20,37 @@ def create_dist_fn(
...
 
+    # NOTE: Do not move/rename because a reference to the function can appear in persisted policies
     @staticmethod
-    def _dist_fn(p: torch.Tensor) -> torch.distributions.Categorical:
-        return torch.distributions.Categorical(logits=p)
+    def _dist_fn(logits: torch.Tensor) -> torch.distributions.Categorical:
+        return torch.distributions.Categorical(logits=logits)
+
+    # NOTE: Do not move/rename because a reference to the function can appear in persisted policies
+    def _dist_fn_probs(self, probs: torch.Tensor) -> torch.distributions.Categorical:
```
MischaPanch: make static? Or, on the contrary, make `_dist_fn` not static and remove this, in favor of `_dist_fn` taking care of choosing logits or probs.

MischaPanch (approved):
Overall LGTM, only minor comments.

Unrelated comment: At some point soon we should add more exhaustive documentation of the ideas behind HL interfaces - I have a hard time understanding why AgentFuture would be needed from a glance at it (I could search in the IDE, ofc, and would eventually understand it, but we should try to reduce the cognitive load)

tianshou/highlevel/module/actor.py
opcode81: I don't necessarily agree with that.
While Python can use covariant return types, this doesn't mean that they have to be used or that their use is necessarily advantageous.

In fact, I would argue that it is fine for the interface to remain unchanged in subclasses (unless there is a specific reason to change it). My decision was somewhat deliberate and not accidental.
If the more specific return type is used, then, notably, a user cannot inherit from the class, override the method and return anything other than the more specific type (i.e. None) without violating the Liskov substitution principle.
The only reason to use a more specific return type is a case where users implement against specialisations rather than the abstraction and thus might care about having more specific return type information. This isn't the case here; all implementations use the abstraction.

To be clear: I wouldn't be opposed to having None as the return type. I just wanted to point out that either option is admissible - and there are advantages and disadvantages to both choices. It's not that it definitely "should be" one or the other...

tianshou/highlevel/module/actor.py
opcode81: See above.

tianshou/highlevel/params/dist_fn.py
opcode81: Indeed, the method should have been static; I will fix this.

> Or, on the contrary, make _dist_fn not static and remove this, in favor of _dist_fn taking care of choosing logits or probs.

This would be bad idea. In fact, it would constitute a bug. As I remarked in the comments, references to these functions appear in persisted policies. If you change the unbound method to a bound method, all persisted instances will cease to function.

opcode81:
> Unrelated comment: At some point soon we should add more exhaustive documentation of the ideas behind HL interfaces - I have a hard time understanding why AgentFuture would be needed from a glance at it (I could search in the IDE, ofc, and would eventually understand it, but we should try to reduce the cognitive load)

Sure, we can add more information where it helps.

My general documentation principle is to document what things are, not how they are used, because the former is invariant/universal while the latter is just a momentary snapshop (and could potentially change in the future). It's usually fine (for me) to search for usages to find out how something is being used.
Also, my principle is to ignore stuff that doesn't momentarily concern me and assume that it's fine. You shouldn't care what ActorFuture is for if you aren't changing a piece of code that uses it.

To reveal: ActorFuture is necessary to support the concept of critics reusing the actor network. When the user declares that he wants to reuse the actor for the critic, I have to create a critic factory that does this, but the actor does not exist yet. So the critic factory instead receives the future, which will eventually be filled when the actor factory is called. When the critic factory is itself called, it can use the then-filled actor to create the critic.

tianshou/highlevel/module/actor.py
MischaPanch: Ah, I see, ok. In one of the factories you did use a stricter type, but I guess that was also deliberate. Feel free to merge

opcode81: Well, my decision in this case was only *somewhat* deliberate, because a key part is that the unchanged signature had already been auto-generated, and, as explained above, I figured that nobody would care about having a more specific return type in this case, because only the abstraction really matters. So I couldn't be bothered to change the signature.

maxhuettenrauch:
Would it make sense to leave the user the option of providing a custom dist_fn optionally? Then of course they need to know what they are doing. I can't think of a practical example right now, but who knows?

opcode81:
> Would it make sense to leave the user the option of providing a custom dist_fn optionally? Then of course they need to know what they are doing. I can't think of a practical example right now, but who knows?

@maxhuettenrauch it is still possible to modify the distribution function, simply by using your own (subclassed) ActorFactory which can override the creation method to return whatever you want. So only the mechanism changed...

PR #1196, opened 2024-08-08
Feature/extension collector buffer

MischaPanch:
Adds important functionality to buffer and collector. The PR is very large but I didn't want to split it up. It's easiest to review commit by commit, and I think various people should have a look. One can also look file by file. Together we can do this ;)

I'll edit the description when the review is done

@Trinkle23897: pls manly have a look at the the changes in buffer related things, and if you want also in the computation of n_step return. I had to slightly modify one of the tests that was changing the private `_insertion_index` leading to a malformed buffer, which now raises an error. Ofc you are very welcome to look at the rest as well :)

@opcode81 and @maxhuettenrauch : pls have a look at the extensions in Collector. They are untested for now, wanted to get your opinion on the design first. Also, a quick glance at the trainer would be nice

Ah, also @Trinkle23897: I think I found a bug in the PPO implementation, see corresponding commit

@dantp-ai : the changes to the buffer here will make the task of fixing slicing issues easier, especially the new names and additional comments. Would also be happy about your review, if you have time!


tianshou/data/buffer/base.py
```diff
@@ -342,6 +398,10 @@ def get(
         index: int | list[int] | np.ndarray,
         key: str,
         default_value: Any = None,
+        # TODO 1: this is only here because of atari, it should never be needed (can be solved with index)
+        #  and should be removed
+        # TODO 2: does something entirely different from getitem
+        # TODO 3: key should not be required
         stack_num: int | None = None,
```
dantp-ai, line 405: `stack_num` can potentially be used for RNNs. I think I saw this note also in one of the tutorials. But if there is no big support planned for RNNs on the horizon, it can probably be removed.

tianshou/data/buffer/base.py
```diff
@@ -72,38 +82,44 @@ def __init__(
...
+        # TODO: remove double negation and different name
         self._save_obs_next = not ignore_obs_next
         self._save_only_last_obs = save_only_last_obs
         self._sample_avail = sample_avail
         self._meta = cast(RolloutBatchProtocol, Batch())
-        self._ep_rew: float | np.ndarray
-        self.reset()
+
+        # Keep in sync with reset!
+        self.last_index = np.array([0])
```
dantp-ai, line 92: Why is the index `self.last_index` an array and `self._insertion_idx` is not?

MischaPanch: Last index is part of the public interface and it exists as a member of the vectorized buffer, where it's an array with multiple entries. Insertion index is private and does not exist in the vectorized buffer (and doesn't need to exist there, or to have the same type)

tianshou/data/buffer/base.py
```diff
@@ -402,7 +465,9 @@ def __getitem__(self, index: slice | int | list[int] | np.ndarray) -> RolloutBat
         if self._save_obs_next:
             obs_next = self.get(indices, "obs_next", Batch())
         else:
-            obs_next = self.get(self.next(indices), "obs", Batch())
+            obs_next_indices = self.next(indices)
+            obs_next = self.get(obs_next_indices, "obs", Batch())
+        # TODO: don't do this
         batch_dict = {
             "obs": obs,
             "act": self.act[indices],
```
dantp-ai, line 473: Can't we iterate here over the `_reserved_keys` ?

dantp-ai: Also, why do we use different access methods here, like `self.act[indices]` and `self.get(indices, "policy")`? Can't we just use `self._meta[key][indices]` on all of them here?

tianshou/utils/array.py
```diff
@@ -0,0 +1,15 @@
+from collections.abc import Sequence
+
+import numpy as np
+
+
+def bisect_left(arr: Sequence[float], x: float) -> float:
```
dantp-ai: The output, when choosing `x` to be the leftmost element of `arr`, is not right:

```
In [5]: arr = [1., 2., 3., 3., 4.]

In [6]: x = 1.

In [7]: bisect_left(arr, x)
Out[7]: 4.0
```

tianshou/utils/array.py
```diff
@@ -0,0 +1,15 @@
...
+    """Assuming arr is sorted, return the largest element `el` of arr s.t. `el < x`."""
+    el_index = int(np.searchsorted(arr, x, side="left")) - 1
+    return arr[el_index]
+
+
+def bisect_right(arr: Sequence[float], x: float) -> float:
```
dantp-ai: Here, if we choose `x` to be the rightmost element of `arr`, we get an `IndexError: list index out of range`.

tianshou/highlevel/config.py
```diff
@@ -81,11 +84,18 @@ class SamplingConfig(ToStringMixin):
...
+    each training step. If this is set, the number of environment steps collected in each
+    collection step is the sum of the lengths of the episodes collected.
+    """
+
     repeat_per_collect: int | None = 1
     """
     controls, within one gradient update step of an on-policy algorithm, the number of times an
     actual gradient update is applied using the full collected dataset, i.e. if the parameter is
-    5, then the collected data shall be used five times to update the policy within the same
+    `n`, then the collected data shall be used five times to update the policy within the same
```
dantp-ai: Looks like a typo. Shouldn't now be "`n` times" here?

tianshou/highlevel/config.py
```diff
@@ -143,3 +161,21 @@ def test_seed(self) -> int:
...
+
+        if self.num_test_episodes != 0 and self.num_test_episodes % self.num_test_envs != 0:
+            log.warning(
+                f"Number of test episodes ({self.num_test_episodes} "
+                f"is not divisible by the number of test environments ({self.num_test_envs}). "
+                f"This can cause unnecessary memory usage, it is recommended to adjust this.",
+            )
+
+        assert (
+            sum([self.step_per_collect is not None, self.episode_per_collect is not None]) == 1
```
dantp-ai, line 191: This is XOR, so we could also simplify it to `a ^ b`. I guess you wrote it like this because in the future we might add other attributes and then `sum()` is more appropriate in that case.

tianshou/policy/modelfree/sac.py
```diff
@@ -21,6 +21,25 @@
 from tianshou.utils.optim import clone_optimizer
 
 
+def correct_log_prob_gaussian_tanh(
+    log_prob: torch.Tensor,
+    squashed_action: torch.Tensor,
```
dantp-ai: `squashed_action` is actually the value squashed by `tanh`. You indicate it in the docstrings, but should we maybe also rename it here for more clarity too? Maybe, `tanh_action`?

Since `correct_log_prob_gaussian_tanh` is a new method introduced here, we won't have any backward compatibility issues for the API.

tianshou/highlevel/config.py
```diff
@@ -67,7 +70,7 @@ class SamplingConfig(ToStringMixin):
     """the total size of the sample/replay buffer, in which environment steps (transitions) are
     stored"""
 
-    step_per_collect: int = 2048
+    step_per_collect: int | None = 2048
```
opcode81, line 73: Docstring must state that this is mutually exclusive with `episode_per_collect` and that one of them must be set.

tianshou/highlevel/config.py
opcode81: This is wrong.

tianshou/highlevel/config.py
```diff
@@ -116,14 +126,22 @@ class SamplingConfig(ToStringMixin):
...
+    Currently only used in Atari examples and may be removed in the future!
+    """
 
     replay_buffer_save_only_last_obs: bool = False
-    """if True, for the case where the environment outputs stacked frames (e.g. because it
-    is using a `FrameStack` wrapper), save only the most recent frame so as not to duplicate
-    observations in buffer memory. Specifically, if the environment outputs observations `obs` with
-    shape (N, ...), only obs[-1] of shape (...) will be stored.
-    Frame stacking with a fixed number of frames can then be recreated at the buffer level by setting
-    :attr:`replay_buffer_stack_num`.
```
opcode81: The old docstring was clearly better (more precise, stating the boundary condition clearly).

tianshou/highlevel/config.py
```diff
@@ -132,8 +150,8 @@ class SamplingConfig(ToStringMixin):
     to the agent for each time step. Setting this to a value greater than 1 can help agents learn
     temporal aspects (e.g. velocities of moving objects for which only positions are observed).
 
-    If the environment already stacks frames (e.g. using a `FrameStack` wrapper), this should either not
-    be used or should be used in conjunction with :attr:`replay_buffer_save_only_last_obs`.
```
opcode81, line 136: Why was the usage recommendation concerning the related parameter dropped?
Not clearly an improvement over the old docstring.

tianshou/highlevel/params/lr_scheduler.py
```diff
@@ -25,6 +25,7 @@ def create_scheduler(self, optim: torch.optim.Optimizer) -> LRScheduler:
 
     class _LRLambda:
         def __init__(self, sampling_config: SamplingConfig):
+            assert sampling_config.step_per_collect is not None
             self.max_update_num = (
                 np.ceil(sampling_config.step_per_epoch / sampling_config.step_per_collect)
                 * sampling_config.num_epochs
```
opcode81, line 31: How do we handle this properly for the case where the user specifies `episode_per_collect` instead?

tianshou/data/collector.py
```diff
@@ -20,17 +22,40 @@
...
+    edges: Sequence[int],
+    start: int,
+    stop: int,
+) -> tuple[tuple[int, int], tuple[int, int]]:
+    """We assume that stop is smaller than start and that `edges` is a sorted array of integers.
+
+    Then it will return the two tuples containing (start, stop) where we go from start to the next edge,
+    and from the previous edge to stop.
+    :return: (start, upper_edge), (lower_edge, stop)
+    """
```
opcode81: Docstring is not sufficient to understand what this function is for.

tianshou/highlevel/config.py
opcode81: I am not sure `(self.step_per_collect is not None) ^ (self.episode_per_collect is not None)` is significantly more readable or clear.
`sensai.utils` has a function: `count_not_none(self.step_per_collect, self.episode_per_collect) == 1`.

tianshou/highlevel/params/lr_scheduler.py
opcode81: It seems to me that we could only handle it properly if the epoch was also specified in terms of episodes (i.e. if there was `episode_per_epoch`).

tianshou/data/collector.py
```diff
@@ -167,6 +207,27 @@ def __init__(
...
+        return self.buffer.subbuffer_edges
+
+    def _get_start_stop_tuples_for_edge_crossing_interval(
+        self,
+        start: int,
+        stop: int,
+    ) -> tuple[tuple[int, int], tuple[int, int]]:
+        """:return: (start, upper_edge), (lower_edge, stop)"""
+        log.debug(
+            "Received an edge-crossing episode: {start=}, {stop=}, {self._subbuffer_edges=}",
```
opcode81: Not an f-string.

tianshou/utils/array.py
MischaPanch: Should have added tests, thanks for spotting!

tianshou/data/buffer/base.py
MischaPanch: I think you're right, I just didn't want to do too many changes at once. Do it in a follow-up?

Already now I introduced some problem that leads to a test not terminating :cry:

tianshou/data/collector.py
MischaPanch: I'll add examples and explain why we need this

tianshou/highlevel/params/lr_scheduler.py
MischaPanch: Not sure it can be handled truly properly, episode length often changes over the course of a training, so scheduling based may be a bad option. Do that in a follow-up issue? It's unrelated to current changes

MischaPanch: Maybe the schedule-free optimizer works so well that we don't need to think too much about it ^^

tianshou/highlevel/config.py
MischaPanch: I didn't do that consciously, probably a late-night copy paste error that escaped my attention. Will revert

tianshou/highlevel/config.py
MischaPanch: I wanted to recommend to use FrameStack instead of using this option, previously this recommendation was not clear. I will restore the "in conjunction with :attr:`replay_buffer_save_only_last_obs`." part though

MischaPanch:
FYI: my changes to the trainer broke something in the offline case, that's why the tests don't terminate. Gonna fix this asap

tianshou/highlevel/config.py
MischaPanch: Until sensai.util is a standalone package, I'd leave this as is. The bitwise is also not more readable to me, in the end it's a matter of personal preference. Resolving this

MischaPanch:
Fixed the offline thing bug - it happened because for offline learning we disabled the progress bar in the tests, and the trainer relied on the progress bar for crucially important operations. Since progress bars can be disabled globally through an env var, I instead switched to tracking the relevant vars explicitly, even though it entails some code duplication.

The type-checker revealed that `self.steps_per_epoch` cannot be None, and I suspect that many optional params of the trainer are in fact not optional... I guess one just wanted to enforce kwargs in the original code, but since we now are in newer python, we can do that in a better way

The fix is in 9bd17be353

tianshou/data/buffer/base.py
dantp-ai: Yes, it's better to refactor get in a follow-up, maybe even in the slicing PR.

tianshou/data/collector.py
MischaPanch: I had to look at it forever to get it myself. Probably the result of some late night coding in the past.

I will remove this function, it shouldn't exist

tianshou/data/buffer/base.py
MischaPanch: Related to #1090 , resolving

tianshou/highlevel/config.py
MischaPanch: Done

tianshou/highlevel/config.py
MischaPanch: Done

tianshou/highlevel/config.py
MischaPanch: Done

tianshou/highlevel/config.py
MischaPanch: Done, pls resolve if you agree with the new formulation

tianshou/utils/array.py
MischaPanch: In the end I just removed this whole module

tianshou/utils/array.py
MischaPanch: removed

tianshou/data/collector.py
MischaPanch: fixed

tianshou/policy/modelfree/sac.py
MischaPanch: Renamed the var and improved the docstring

maxhuettenrauch:
Regarding the collector updates, did you base them on the changes in the imac branch only? Or did you also consider the approach I made in the imac-with-buffer branch? There, I already made some efforts of moving the buffer related parts to the buffer and also moved some of the logic outside the Collector.

MischaPanch:
> Regarding the collector updates, did you base them on the changes in the imac branch only? Or did you also consider the approach I made in the imac-with-buffer branch? There, I already made some efforts of moving the buffer related parts to the buffer and also moved some of the logic outside the Collector.

@maxhuettenrauch I thought I did base it on the imac-with-buffer branch, but now realized that I was wrong. Will adjust the collector-buffer things accordingly, your design is indeed better than this one (at least after a quick look)!

MischaPanch:
@maxhuettenrauch 

I had a closer look at your implementation in the work on imac. I understood that you needed to customize the CollectStats computation but I think it's not a good idea to couple stats with the buffer. Apart from complicating interactions between various concepts, the collect stats should be info from the collect iteration and not from the whole buffer. I am also not a fan of dynamically defined classes, I consider them an antipattern and they add a lot of complexity (as you have seen yourself when giving appropriate names).

I also considered bundling the callbacks into a single class, but in the end decided not to do that. It's an additional indirection and an additional class, adding complexity to the user and I don't think it adds much value. The hooks are a bit complicated to understand and the documentation of them and access to them from the collector should be as visible as possible. So I think leaving them as explicit kwargs and with explicit getters/setters is better.

In order to permit the customization you required I instead moved the logic of stats computation and updates to the `CollectStats` class itself and made collector generic. So, I did implement your change on the Collector receiving a `CollecStats` constructor. The genericness allows for static analysis with customized stats.

To compensate for the increased complexity in `_collect` I have added very detailed documentation that helps understanding and navigating the code. I also added tests now.

For me, this PR is ready to be merged. Pls have a look at the changes and see whether you agree or have comments.

After this is merged, we should extend the highlevel interfaces to include the possibility to customize stats.

This PR also presents a large step towards customized logging.

The main additional change is in 64cdf149a6. To make mypy happy the generic had to be specified wherever Collector is instantiated, so most files in the project were affected, polluting the diff

@opcode81 @bordeauxred FYI

MischaPanch:
After this PR and the corresponding extension of HL interfaces are through, I'd release the new version of tianshou, as this presents useful additions to the functionality and customizability of the library

tianshou/data/collector.py
```diff
@@ -344,45 +538,136 @@ class Collector(BaseCollector):
...
+    #
     def __init__(
         self,
         policy: BasePolicy,
         env: gym.Env | BaseVectorEnv,
         buffer: ReplayBuffer | None = None,
         exploration_noise: bool = False,
+        on_episode_done_hook: Optional["EpisodeRolloutHookProtocol"] = None,
+        on_step_hook: Optional["StepHookProtocol"] = None,
+        raise_on_nan_in_buffer: bool = True,
```
maxhuettenrauch, line 578: Argument is unused

tianshou/data/collector.py
```diff
@@ -44,20 +87,38 @@ class CollectStatsBase(DataclassPPrintMixin):
...
-    returns_stat: SequenceSummaryStats | None  # can be None if no episode ends during the collect step
+    returns_stat: SequenceSummaryStats | None = None
     """Stats of the collected returns."""
-    lens: np.ndarray
+    lens: np.ndarray = field(default_factory=lambda: np.array([], dtype=int))
     """The collected episode lengths."""
-    lens_stat: SequenceSummaryStats | None  # can be None if no episode ends during the collect step
+    lens_stat: SequenceSummaryStats | None = None
     """Stats of the collected episode lengths."""
+    std_array: np.ndarray | None = None
```
maxhuettenrauch, line 118: Should we really keep this in the basic collect stats? If so, we should give it a more descriptive name

tianshou/data/collector.py
MischaPanch: Nice catch! I moved the check to BaseCollector now

tianshou/data/collector.py
```diff
@@ -642,15 +1066,53 @@ def _collect(
...
 
-        return CollectStats.with_autogenerated_stats(
-            returns=np.array(episode_returns),
-            lens=np.array(episode_lens),
-            n_collected_episodes=num_collected_episodes,
-            n_collected_steps=step_count,
-            collect_time=collect_time,
-            collect_speed=step_count / collect_time,
-        )
+    # TODO: move to buffer
```
maxhuettenrauch, line 1086: Should we still try and remove all this logic from the collector? I had to change something in the buffer to have the same functionality, but, after all, the whole logic only concerns buffer stuff and not collector stuff.

tianshou/data/collector.py
```diff
@@ -543,52 +892,129 @@ def _collect(
...
+                    (
+                        cur_ep_index_array,
+                        cur_ep_batch,
+                    ) = self._get_buffer_index_and_entries_for_episode_from_slice(
+                        cur_ep_index_slice,
+                    )
+                    cur_ep_batch = cast(EpisodeBatchProtocol, cur_ep_batch)
+
+                    # Step 10
+                    episode_hook_additions = self.run_on_episode_done(cur_ep_batch)
```
maxhuettenrauch, line 984: Maybe more of an aesthetics thing, but the on step hook modifies the batch in place, while the episode done hook returns a dictionary and does the buffer manipulation in sight

MischaPanch: Yes, it's because the episode hook operates on a batch that was already added to the buffer. It can't modify the buffer directly. The step hook operates before the transition is added, so it can to anything in place. So they need different signatures

tianshou/data/collector.py
MischaPanch: Ah, I kinda forgot that it was added in this PR and didn't want to change here. Yes, I'll change the name. I'm also not sure the sequence stats work here (the std might be multi-dimensional)

tianshou/data/collector.py
MischaPanch: Yes, do that in a follow-up PR?

MischaPanch: It's related to general buffer slicing logic, which also @dantp-ai is working on afaik. So we could improve the buffer slicing in a separate PR once and for all later

tianshou/data/collector.py
maxhuettenrauch: Yup. Once the buffer related stuff is moved, it may become a little less cluttered

MischaPanch:
Integration to HL Interfaces will follow, some changes to the design might happen then. But this gives a viable state to build on.

Since it's mainly an extension, and was thoroughly tested, I went in for the merge

test/continuous/test_ppo.py
```diff
@@ -122,12 +122,12 @@ def dist(loc_scale: tuple[torch.Tensor, torch.Tensor]) -> Distribution:
         action_space=env.action_space,
     )
     # collector
-    train_collector = Collector(
+    train_collector = Collector[CollectStats](
```
Trinkle23897, line 125: small nit: can we make it as default setting?

tianshou/data/batch.py
```diff
@@ -278,6 +278,13 @@ def get_sliced_dist(dist: TDistribution, index: IndexType) -> TDistribution:
         raise NotImplementedError(f"Unsupported distribution for slicing: {dist}")
 
 
+def get_len_of_dist(dist: Distribution) -> int:
```
Trinkle23897, line 281: nit: why do you put dist util function under buffer?

Trinkle23897 (commented):
very minor point, overall lgtm

tianshou/data/batch.py
MischaPanch: there's a bunch of dist utils that are scattered around collector, batch and buffer at the moment, and we don't have a dedicated module for them yet. 

I wanted to make one and also a dedicated test module in a separate PR soon (before the next release). Until then, it was just a sort of random decision where to put them

test/continuous/test_ppo.py
MischaPanch: Unfortunately, we can't - it's a bug in mypy. It already is the default setting (as the bound in the Generic is set to CollectStats), but mypy will still complain. Until the issue is fixed in numpy, the smaller evil is to specify the type in our codebase.

Users don't have to do this and will get proper autocompletion without this. Also, I think pylance doesn't have this bug, but well, we're using mypy in CI for now

Trinkle23897: should we replace mypy with pyright?

MischaPanch: Maybe, I was also thinking about that for a while. It wouldn't be super smooth though, pyright is stricter and there will probably be new typing errors.

For now, my pain with mypy is not enough to make me want to implement the switch. If you or anyone from the community want to do that, feel free! I essentially don't care as long as we have some type-checker in CI

PR #1202, opened 2024-08-13
Support computing custom scores and terminating/saving based on them in BaseTrainer

anyongjin:
This PR introduces a new concept into tianshou training: a `best_score`. It is computed from the appropriate `Stats` instance and always added to `InfoStats`.

## Breaking Changes:
- `InfoStats` has a new non-optional field `best_score`

## Background

Currently, tianshou uses the maximum average return to find the best model. But sometimes it may not meet user needs, for example, the average return only drops by 5%, but the standard deviation drops by 50%. The latter is generally considered to be more stable and better than the former.

### DIFF
--- a/tianshou/data/stats.py
+++ b/tianshou/data/stats.py
@@ -56,6 +56,8 @@ class InfoStats(DataclassPPrintMixin):
 
     gradient_step: int
     """The total gradient step."""
+    best_score: float
+    """The best score over the test results."""
     best_reward: float
     """The best reward over the test results."""
     best_reward_std: float
--- a/tianshou/data/stats.py
+++ b/tianshou/data/stats.py
@@ -57,7 +57,7 @@ class InfoStats(DataclassPPrintMixin):
     gradient_step: int
     """The total gradient step."""
     best_score: float
-    """The best score over the test results."""
+    """The best score over the test results. The one with the highest score will be considered the best model."""
     best_reward: float
     """The best reward over the test results."""
     best_reward_std: float
--- a/tianshou/trainer/base.py
+++ b/tianshou/trainer/base.py
@@ -73,7 +73,8 @@ class BaseTrainer(ABC):
     :param test_fn: a hook called at the beginning of testing in each
         epoch. It can be used to perform custom additional operations, with the
         signature ``f(num_epoch: int, step_idx: int) -> None``.
-    :param evaluate_test_fn: Calculate the test batch performance score to determine whether it is the best model
+    :param compute_score_fn: Calculate the test batch performance score to
+        determine whether it is the best model, the mean reward will be used as score if not provided.
     :param save_best_fn: a hook called when the undiscounted average mean
         reward in evaluation phase gets better, with the signature
         ``f(policy: BasePolicy) -> None``.
@@ -165,7 +166,7 @@ class BaseTrainer(ABC):
         train_fn: Callable[[int, int], None] | None = None,
         test_fn: Callable[[int, int | None], None] | None = None,
         stop_fn: Callable[[float], bool] | None = None,
-        evaluate_test_fn: Callable[[CollectStats], float] | None = None,
+        compute_score_fn: Callable[[CollectStats], float] | None = None,
         save_best_fn: Callable[[BasePolicy], None] | None = None,
         save_checkpoint_fn: Callable[[int, int, int], str] | None = None,
         resume_from_log: bool = False,
@@ -213,7 +214,9 @@ class BaseTrainer(ABC):
         self.train_fn = train_fn
         self.test_fn = test_fn
         self.stop_fn = stop_fn
-        self.evaluate_test_fn = evaluate_test_fn
+        self.compute_score_fn = compute_score_fn
+        if self.compute_score_fn is None:
+            self.compute_score_fn = lambda stat: stat.returns_stat.mean
         self.save_best_fn = save_best_fn
         self.save_checkpoint_fn = save_checkpoint_fn
 
@@ -277,10 +280,7 @@ class BaseTrainer(ABC):
                 test_result.returns_stat.mean,
                 test_result.returns_stat.std,
             )
-            if self.evaluate_test_fn:
-                self.best_score = self.evaluate_test_fn(test_result)
-            else:
-                self.best_score = self.best_reward
+            self.best_score = self.compute_score_fn(test_result)
         if self.save_best_fn:
             self.save_best_fn(self.policy)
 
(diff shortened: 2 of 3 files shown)


tianshou/trainer/base.py
```diff
@@ -73,6 +73,7 @@ class BaseTrainer(ABC):
     :param test_fn: a hook called at the beginning of testing in each
         epoch. It can be used to perform custom additional operations, with the
         signature ``f(num_epoch: int, step_idx: int) -> None``.
+    :param evaluate_test_fn: Calculate the test batch performance score to determine whether it is the best model
```
MischaPanch: Behavior when None is passed needs to be documented.

tianshou/data/stats.py
```diff
@@ -56,6 +56,8 @@ class InfoStats(DataclassPPrintMixin):
 
     gradient_step: int
     """The total gradient step."""
+    best_score: float
```
MischaPanch, line 59: Rename to `best_custom_score`? Should be optional and the docstring should be extended.

Overall, I'm a bit skeptical about adding it here, since the semantics of it is not well defined and depends on things configured in other places. An alternative to a fixed name would be to extend the `InfoStats` with 

```
metadata: dict = field(default_factory=dict)
```

but this won't work with current Trainer structure.


Generally, looking at the trainer design, the termination of the training should either follow a strategy pattern or take a callable that calculates the score based on CollectStats or InfoStats. Termination should also be based on some stats object, or on the whole trainer. That would be too much of a change, so maybe we can merge this now and make a larger breaking change later.

@opcode81 wdyt?

tianshou/trainer/base.py
```diff
@@ -273,6 +277,10 @@ def reset(self, reset_collectors: bool = True, reset_buffer: bool = False) -> No
                 test_result.returns_stat.mean,
                 test_result.returns_stat.std,
             )
+            if self.evaluate_test_fn:
```
MischaPanch: Instead of doing this, I think we should have 

```
self.evaluate_test_fn = lambda info_stats: info_stats.best_reward
```

by default (if that's the way in which reward is currently being set in the trainer, ofc)

tianshou/trainer/base.py
```diff
@@ -384,17 +393,29 @@ def test_step(self) -> tuple[CollectStats, bool]:
         )
         assert test_stat.returns_stat is not None  # for mypy
         rew, rew_std = test_stat.returns_stat.mean, test_stat.returns_stat.std
-        if self.best_epoch < 0 or self.best_reward < rew:
+        if self.evaluate_test_fn:
```
MischaPanch: see above, no need for the if-else with a default `evaluate_test_fn`

tianshou/trainer/base.py
```diff
@@ -164,6 +165,7 @@ def __init__(
         train_fn: Callable[[int, int], None] | None = None,
         test_fn: Callable[[int, int | None], None] | None = None,
         stop_fn: Callable[[float], bool] | None = None,
+        evaluate_test_fn: Callable[[CollectStats], float] | None = None,
```
MischaPanch: rename to `compute_custom_score_fn`?

tianshou/trainer/base.py
```diff
@@ -384,17 +393,29 @@ def test_step(self) -> tuple[CollectStats, bool]:
...
             self.best_reward = float(rew)
             self.best_reward_std = rew_std
             if self.save_best_fn:
                 self.save_best_fn(self.policy)
-        log_msg = (
-            f"Epoch #{self.epoch}: test_reward: {rew:.6f} ± {rew_std:.6f},"
-            f" best_reward: {self.best_reward:.6f} ± "
-            f"{self.best_reward_std:.6f} in #{self.best_epoch}"
-        )
+        if self.evaluate_test_fn:
```
MischaPanch: also this if-else should disappear with a default scoring function

MischaPanch (changes requested):
Thanks for the PR @anyongjin, it's a good contribution!

Overall, the trainer has to become more flexible, but it would be too much to ask for right now. I think we can merge this after some slight changes and then soon refactor the trainer, taking in consideration the support for custom scoring and custom conditions on terminating the training

anyongjin:
In essence, `Average Reward` and `Test Score` are two different things. The former represents a fixed test result indicator. The latter is a score for the test result. The scoring logic for different tasks and users may be different. For example, some consider the standard deviation and some do not.
Currently, tianshou uses `best_reawrd` for both average reward and test score. It is difficult for users to implement custom scoring logic. So I suggest that `best_reward` be used only for average reward, and `best_score` be added for test score. In this way, `best_reward` and `best_score` are two different things. If it is called `best_custom_score`, people will think that there is a system default score field, so I think it is better not to add 'custom'.

Update:
* Added explanation for InfoStats.best_score.
* Use lambda function when compute_score_fn is None to avoid multiple if-else

tianshou/data/stats.py
MischaPanch: After another look, I guess it's ok to have scores decoupled from rewards as a central concept, so it can be left as is. Resolving

tianshou/trainer/base.py
```diff
@@ -384,17 +398,27 @@ def test_step(self) -> tuple[CollectStats, bool]:
...
             self.best_reward = float(rew)
             self.best_reward_std = rew_std
             if self.save_best_fn:
                 self.save_best_fn(self.policy)
-        log_msg = (
-            f"Epoch #{self.epoch}: test_reward: {rew:.6f} ± {rew_std:.6f},"
-            f" best_reward: {self.best_reward:.6f} ± "
-            f"{self.best_reward_std:.6f} in #{self.best_epoch}"
-        )
+        if score != rew:
```
MischaPanch, line 409: Personally, I'm not a fan of duplicated log message contents, but this is just FYI. I would have done something like 
```
score_info_addition = "" if score == rew else f", score: {score: .6f},"
```
and used it in `log_msg`. Feel free to leave as is or to adjust if you agree with me

MischaPanch (approved):

tianshou/trainer/base.py
anyongjin: I also think that the redundant code around `log_msg` is not so elegant, but I didn't find a better way before. 
Your suggestion is pretty good. I will modify and submit it later. Thank you.

MischaPanch: Perfect, thanks! Merged it

opcode81:
@anyongjin the "terminating" part was not actually implemented. If it is the score we seek to maximize, then the early stopping criterion (`stop_fn`) should also use scores instead of mean returns, don't you agree?
I noticed this, because I am refactoring and improving the library for v2. I will change this in the v2 branch.

anyongjin:
Yes, if `best_score` is used to determine whether it is the optimal model, then `best_score` should also be passed when calling `stop_fn` instead of the previously default average return. I forgot to modify this part earlier, and it should be corrected in v2.

PR #1207, opened 2024-08-14
Support for using random agents, improvements in CollectStats

MischaPanch:
Random action agents can be useful for tests, but also as a baseline.

Also (independent change):

- improved CollectStats for action_std
- Batch: extension for changing shape to atleast_2d, including "casting" torch distributions to atleast_2d

One thing is a bit weird in this implementation: in high level interfaces the optim_factory is obligatory. For random actions it doesn't make sense, of course, but making it optional would require changes and additional type checks in several places. @opcode81 what do you think about this - should we make it optional?

@Trinkle23897 a quick glance should be enough, nothing controversial. @opcode81 is on vacation and won't have time to review


MischaPanch:
@maxhuettenrauch I improved the collect-stats for action dists in 6f8648ab141 and b0ba423184, also adding tests for stats collection.  FYI

MischaPanch:
Since it's mostly an extension (apart from a minor renaming of an existing class), I'm going to merge this. Feel free to comment post-merge if there's anything to be discussed

PR #1243, opened 2025-02-09
pettingzoo example observation_space fix

yashschandra:
Had to make few changes to run tictactoe example (was getting errors otherwise)


liuzhaoze:
Using `reset_before_collect=True` in `collector.collect` may be better (see #1240)

Trinkle23897 (approved):

Issue #956, opened 2023-10-02
Support for Loading Minari Datasets


ivanappliedai:
Currently, the repository provides a convenient method, load_buffer_d4rl, to load D4RL datasets for offline RL algorithms. However, with the ongoing migration from D4RL to Minari, it is becoming increasingly important to support Minari datasets loading directly.

PR #1257, opened 2025-05-13
Minari support 

younik:
Hi, 
I have seen there is an open issue to support Minari https://github.com/thu-ml/tianshou/issues/956, and you are actively working on V2.
I would be happy to help support it!

I provided here a proof-of-concept on how to do it.
Minari divides the dataset into episodes, while Tianshou ReplayBuffer doesn't.
So, here, I just load everything into memory. I am not very familiar with Tianshou, but I guess we can do it better

### DIFF
--- a/examples/offline/utils.py
+++ b/examples/offline/utils.py
@@ -2,9 +2,51 @@ import d4rl
 import gymnasium as gym
 import h5py
 import numpy as np
+import minari
 
-from tianshou.data import ReplayBuffer
+from tianshou.data import ReplayBuffer, Batch
 from tianshou.utils import RunningMeanStd
+
+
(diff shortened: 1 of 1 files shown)


MischaPanch:
Hi @younik . Sorry for the late reply, this slipped through my radar. We will be releasing v2 in version alpha now, I will have a look at this afterwards to see if with this change minari is fully supported. Need to test it before we can merge

younik:
> Hi @younik . Sorry for the late reply, this slipped through my radar. We will be releasing v2 in version alpha now, I will have a look at this afterwards to see if with this change minari is fully supported. Need to test it before we can merge

Hi, no worries.

Sure, let me know if there is anything I can do to help

PR #1284, opened 2025-11-18
Finalize docs

MischaPanch:


MischaPanch:
Merging this since docs are easier to review post-merge. This is an improvement of the current status, even if we might further improve after review

@opcode81 FYI

PR #1287, opened 2026-02-11
fix: restore parameters on TRPO line search failure

Mr-Neutr0n:
## Bug
When the TRPO line search fails to find a step that satisfies the KL constraint, the original policy parameters are not restored. This leaves the policy with an untested update that violates the trust region.

## Fix
Added parameter restoration to the line search fallback path.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Unreleased Changes
 
 Fixes:
+* `algorithm`:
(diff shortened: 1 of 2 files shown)


opcode81:
Please add your change to the changelog, adding a new section "Unreleased Changes" at the top.
Also, it can't hurt to add a comment in the code.

EDIT: I handled this.

PR #1295, opened 2026-04-03
[data collector] Use monotonic clocks for collector timing

### LINKED ISSUE #1291: Collector.collect() crashes with negative collect_time due to time.time() clock skew
## Bug

`Collector.collect()` uses `time.time()` to measure elapsed time ([collector.py#L501-L510](https://github.com/thu-ml/tianshou/blob/master/tianshou/data/collector.py#L501-L510)):

```python
pre_collect_time = time.time()
with torch_train_mode(self.policy, enabled=False):
    collect_stats = self._collect(...)
collect_time = time.time() - pre_collect_time
collect_stats.set_collect_time(collect_time, update_collect_speed=True)
```

`time.time()` returns wall-clock time, which can jump **backward** during NTP clock synchronization. When this happens, `collect_time` becomes negative and `set_collect_time` raises:

```
ValueError: Collect time should be non-negative, but got collect_time=-0.1259613037109375.
```

This kills the training run mid-epoch (in my case, epoch 340 of 800 in a multi-seed experiment).

## Expected behavior

Timing measurement should be robust to system clock adjustments and not crash training.

## Suggested fix

Replace `time.time()` with `time.monotonic()`, which is guaranteed to never go backward (immune to NTP adjustments):

```python
pre_collect_time = time.monotonic()
with torch_train_mode(self.policy, enabled=False):
    collect_stats = self._collect(...)
collect_time = time.monotonic() - pre_collect_time
```

Note: the `AsyncCollector` already guards against this at [line 1385](https://github.com/thu-ml/tianshou/blob/master/tianshou/data/collector.py#L1385) with `max(time.time() - start_time, 1e-9)`, but the main `Collector.collect()` path has no such protection.

## Environment

- tianshou 2.0.0
- Python 3.13
- macOS (Darwin)

Trinkle23897:
## Summary
- Problem: `Collector.collect()` measures elapsed time with `time.time()`, so a wall-clock rollback can produce negative `collect_time` and raise `ValueError` in `CollectStats.set_collect_time`.
- Scope: switch collector elapsed-time measurements to `time.monotonic()` and add a regression test that simulates wall-clock rollback.
- Outcome: collector stats stay valid across system clock adjustments, and the issue report is covered by a dedicated test.

Fixes #1291.

## Technical Details
- Approach: use a monotonic clock for both the synchronous `BaseCollector.collect` wrapper and `AsyncCollector._collect` internal timing so elapsed durations are independent of wall-clock/NTP adjustments.
- Code pointers:
  - `tianshou/data/collector.py`: switch elapsed-time sources to `time.monotonic()`.
  - `test/base/test_collector.py`: add a fake clock regression test where `time.time()` goes backward while `time.monotonic()` increases.
- Notes: no API change; this only affects internal duration measurement and derived collect speed.

### Automated
- `pytest test/base/test_collector.py -k 'monotonic_time or test_collector_with_vector_env or test_async_collector_with_vector_env' -q`: passed
- `pytest test/base/test_collector.py -q`: passed, 17 passed
- `ruff check tianshou/data/collector.py test/base/test_collector.py`: passed
- `ruff format --check tianshou/data/collector.py test/base/test_collector.py`: passed
### Suggested Manual
- `collector.collect(...)` under a host clock adjustment: verify training no longer crashes with a negative `collect_time`.

### DIFF
--- a/test/base/test_collector.py
+++ b/test/base/test_collector.py
@@ -6,6 +6,7 @@ import numpy as np
 import pytest
 import tqdm
 
+import tianshou.data.collector as collector_module
 from test.base.env import MoveToRightEnv, NXEnv
 from tianshou.algorithm.algorithm_base import Policy, episode_mc_return_to_go
 from tianshou.data import (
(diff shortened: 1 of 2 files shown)


opcode81:
@Trinkle23897 please don't forget to update the change log.
Agents don't read the PR template's checklist.

PR #1298, opened 2026-05-04
Fix division-by-zero in NPG advantage normalization

Pran-Ker:
In NPG._preprocess_batch, advantage normalization divides by batch.adv.std() without adding an epsilon, which produces nan values whenever all advantages are equal (std=0) or a single-element batch is processed (std=nan). By contrast, PPO already guards against this with (std + self._eps), and self._eps=1e-8 is available in NPG because it inherits ActorCriticOnPolicyAlgorithm. The nan values silently corrupt the logp_old stored on the batch and cause the actor loss to become nan for the rest of training. TRPO also inherits this preprocess via super(), so both algorithms are affected.

---
*Generated by RL PR Hunter*


opcode81 (changes requested):
Thank you for this improvement.
Please add an entry to the changelog, then we can merge this.

PR #1310, opened 2026-09-09
Allow configuring clip_max on VectorEnvNormObs

09Catho:
RunningMeanStd already accepts a clip_max parameter, but VectorEnvNormObs
always constructed its obs_rms with defaults, so there was no way to
change the observation clip value except by monkey-patching obs_rms
after construction. The issue calls this out directly.

This adds a clip_max parameter to VectorEnvNormObs.__init__, forwarded
straight to RunningMeanStd(clip_max=clip_max). The default stays 10.0,
so existing behavior is unchanged for anyone not passing it.

One note: the issue was originally filed against PGPolicy.ret_rms.
That code has since moved to Reinforce/A2C in
tianshou/algorithm/modelfree, and those classes only use ret_rms.mean
and ret_rms.var for return standardization, never .norm(), so
ret_rms.clip_max isn't wired to anything there. The only place
clip_max actually does something today is VectorEnvNormObs.obs_rms,
via _norm_obs -> obs_rms.norm(). That's what this PR fixes.

Added test_venv_wrapper_norm_obs_clip_max in test/base/test_env.py,
covering the default clip_max, a custom clip_max that actually clips,
and clip_max=None disabling clipping. It fails against the old
constructor (TypeError: unexpected keyword argument 'clip_max') and
passes after the change.

Testing:
- `pytest test/base/test_env.py::test_venv_wrapper_norm_obs_clip_max test/base/test_env.py::test_venv_wrapper_gym` -> 2 passed
- `pytest test/base/test_env.py` (full file) -> 10 passed, 2 skipped (envpool/ray not installed)
- `ruff check tianshou/env/venv_wrappers.py test/base/test_env.py` -> all checks passed
- `ruff format --check tianshou/env/venv_wrappers.py test/base/test_env.py` -> 2 files already formatted
- `mypy tianshou/env/venv_wrappers.py` -> could not run; mypy 1.4.1 (pinned by the repo) crashes parsing numpy's stubs under Python 3.12 with newer numpy (PEP 695 type alias syntax it doesn't support yet), unrelated to this change


09Catho:
Addressed the review feedback in a61f5e14.

- `RunningMeanStd.norm()` now checks `clip_max is not None`, so `clip_max=0.0` clips to zero while `None` still disables clipping.
- Added regression coverage for zero clipping.
- Relaxed the custom and unclipped assertions to check clipping behavior rather than incidental normalization values.

Validation:
- `.venv\Scripts\python.exe -m pytest -q -p no:cov test/base/test_env.py::test_venv_wrapper_norm_obs_clip_max` (1 passed)
- `.venv\Scripts\python.exe -m pytest -q -p no:cov test/base/test_env.py::test_venv_wrapper_gym` (1 passed)
- `.venv\Scripts\ruff.exe check tianshou/utils/statistics.py tianshou/env/venv_wrappers.py test/base/test_env.py` (all checks passed)
- `.venv\Scripts\ruff.exe format --check tianshou/utils/statistics.py tianshou/env/venv_wrappers.py test/base/test_env.py` (3 files already formatted)
- `git diff --check` passed

